Issue #3, opened 2022-03-30
Add Sentry


danielburnley:
We want to use sentry for exception tracking, we will need to set up a sentry account and ensure the service can send exceptions correctly

alice-carr:
closing since #7  is merged

PR #7, opened 2022-04-05
Add Sentry

alice-carr:
I've added a very basic set up for Sentry here, the configuration is largely untouched, but I figure we can add to this once we've started developing.

This requires the use of an API key, which I've put into a .env file for the time being, and pulled in with ruby using the `dotenv` gem - let me know if there's a better or more standard way of doing this in ruby. 
I've imported  `dotenv/load` at the top of  `config/initializers/sentry.rb` but I have a feeling there's a better place to put this. 

To use this, sign up to sentry and add the sentry dns to .env as SENTRY_DNS. I've left in some exceptions to triggers entries to Sentry in `config/initializers/sentry.rb` ll13-19 - uncomment these, run the project, and see entries appear in Sentry.

relates #3

### DIFF
--- a/config/initializers/sentry.rb
+++ b/config/initializers/sentry.rb
@@ -1,0 +1,19 @@
+require 'dotenv/load'
+
+Sentry.init do |config|
+    config.dsn = ENV["SENTRY_DSN"]
+    config.breadcrumbs_logger = [:sentry_logger, :http_logger]
+    config.debug = true
+    config.traces_sample_rate = 1.0
+  end
+
+# Uncomment out the below to test Sentry - this
+# will raise 2 issues in Sentry
+
+# begin
+#   1 / 0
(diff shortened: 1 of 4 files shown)


config/initializers/sentry.rb
```diff
@@ -0,0 +1,19 @@
+require 'dotenv/load'
```
danielburnley, line 1: I don't _think_ we need this require here if i'm reading what `dotenv-rails` does correctly in the README. If you remove it does it still work?

alice-carr: It does work without it - not sure why I put it in there. I'll take it out now

danielburnley (approved):
Looks good to me 🎉

0atman (approved):

PR #8, opened 2022-04-07
Configure frontend build

### LINKED ISSUE #4: Configure Frontend build
This should include anything we're likely to need in both forms-admin and forms-runner, but not anything more specific. We already have a default Rollup build but this may also need a but of configuring.

This should include:
- [ ] babel for transpiling
- [ ] a browserslist configuration (start with a sensible default, we will adjust this later once we have a better idea of our user stats)
- [ ] jest for unit testing
- [ ] cypress for end-to-end testing
- [ ] style and js linting
- [ ] autoprefixer
- [ ] dart-sass instead of ruby-sass
- [ ] install `govuk-frontend`
- [ ] style and js formatting

DavidBiddle:
This is the same frontend build implemented on the forms-admin repo.

Closes #4


danielburnley (approved):
Oop - missed this. Looks good! 🎉

Issue #21, opened 2022-05-03
Notify Integration


alice-carr:
Trello backlink https://trello.com/c/xewAy5s7/123-spike-allow-user-to-create-a-form-with-a-title-that-sends-email-to-govuk-notify

End User can press a button that sends an email using Notify, including the title, text input, and submission time

alice-carr:
To update where I got with this today - the docs for Notify are indeed really good. I set up a basic integration to send an email via notify, but nothing was coming through when I tested - I have a feeling this is mostly because the file is not being run and I'm new to ruby so don't fully know what I'm doing, but I didn't have time to investigate further. 

I've opened a draft PR link above and pushed all my changes up so far.

alice-carr:
I've also just seen that the PR is failing on tests so I have indeed done something wrong with the ruby. Will investigate when I'm back!

PR #22, opened 2022-05-03
Notify Integration

alice-carr:
End User needs to be able to press a button that sends an email using Notify, including the title, text input, and submission time. To do this we need to integrate Notify into the runner. 

This PR is a work in progress

relates #21 

#### Checklist

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -35,3 +35,11 @@ Style/Documentation:
 # https://github.com/rubocop-hq/rubocop/issues/7197
 Style/FrozenStringLiteralComment:
   Enabled: false
+
+# Analog of: 736b3d295f88b9ba6676fc168b823535582388c2
+# "Disable opinionated cops"
+#
+# We should avoid cops that are based on heuristics, since
+# it's not clear what action to take to fix an issue.
+Metrics/MethodLength:
+  Enabled: false
--- a/Gemfile
+++ b/Gemfile
@@ -31,6 +31,7 @@ gem "sentry-rails"
 
 gem 'dotenv-rails', groups: [:development, :test]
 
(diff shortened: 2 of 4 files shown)


alice-carr:
I've created a template to use for this and added this into the PR.
The linting is currently failing because the method has too many lines - I could fix this by removing the check for the api key but that doesn't feel like the best thing to do. 

I would also appreciate if someone could tell me if `forms-runner/app/services/` is the right place for this file, or is there somewhere better for it?

danielburnley (approved):
Seems good to me for now - I can imagine we'll probably revisit some of this a lil when we start the actual work on the runner but as part of this spike it works well 🎉

.rubocop.yml
```diff
@@ -34,4 +34,12 @@ Style/Documentation:
...
 Style/FrozenStringLiteralComment:
+  Enabled: false
+
+# Analog of: 736b3d295f88b9ba6676fc168b823535582388c2
+# "Disable opinionated cops"
+#
+# We should avoid cops that are based on heuristics, since
+# it's not clear what action to take to fix an issue.
+Metrics/MethodLength:
   Enabled: false
```
0atman, line 45: I wonder if we should standardise on a fairly long method length? I agree that rubocop's defaults are very restrictive, but I'd really hate to scroll through many pages of methods, right?

0atman (approved):
Looks good :ok_hand:

.rubocop.yml
0atman: Certainly shouldn't block this pr :ok_hand:

0atman: captured https://github.com/alphagov/forms/issues/27

PR #23, opened 2022-05-12
20 submit a form

thomasiles:
Add a simple way to show and submit a form pulled from the API-runner. To try it, run the API service, add a notify key (see the readme), and hit `/forms/1/` .  You should see an unstyled form, which when you add text and submit., which passes the text on to notify.

There's not much structure/testing atm, but I've added rspec etc to get going.

It's not well tested and
#### Checklist

    - [x] README.md
    - [ ] Elsewhere (please link)


app/controllers/forms/pages_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    def allow_params
+      params.permit(:text, :form_id)
+    end
+
+    def prepare_form
+      @form = Form.find(params[:form_id])
+    end
+
+    def submit
+      # in the controller for now but can be moved to service object, maybe use actionmailer fo easier testing?
```
danielburnley, line 34: I'd agree with this move in the future yea - for now it works out alright to get things done. I think we'll likely end up with some kinda `SubmitAnswers` service which handles the formatting of form -> notify content and then gives that content to a notify mailer

app/models/form.rb
```diff
@@ -0,0 +1,4 @@
+class Form < ActiveResource::Base
+  self.site = ENV.fetch('API_BASE')
```
danielburnley, line 2: I changed this in my branch to be the following so that if we end up needing to handle versioning we can do it per resource - then change the `API_BASE` to no longer include the version

```suggestion
  self.site = "#{ENV.fetch('API_BASE')}/v1"
```

app/views/forms/index.erb
```diff
@@ -0,0 +1 @@
+<p class="govuk-body">TODO: Add a list of forms</p>
```
danielburnley, line 1: I don't think we'll end up with the runner having a list of forms at least - but this can be left as placeholder for now

danielburnley (approved):
Smol suggestion around the versioning but not necessarily important to do now 🎉

app/models/form.rb
thomasiles: Good idea, I've updated it

app/views/forms/index.erb
thomasiles: I agree. I was adding it because the testing it was pretty brutal - I might make it root and add a few hardcoded to links in it as we go along to make things more friendly to people who want to poke it

PR #26, opened 2022-05-18
Rubocop govuk

danielburnley:

Rather than adding each exception one by one, as we're operating within GDS (and also somewhat related to the GOV.UK area) this adds the govuk rubocop config and updates the entire codebase to be inline with it (the gem iteself defines which folders not to lint)

#### Checklist


0atman (approved):
:robot: :police_officer:  :ok_hand:

PR #30, opened 2022-05-19
WIP: Add hosting

thomasiles:
#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)

Add Form repositories

Currently the project needs to have the Form API project running to work.

This is OK but coupling the two in development slows work down.

There are a few ways round this:
 - Run the API alongside the runner in development (maybe both in a
   docker compose environment)
 - Run an API stub alongside the runner
 - Use something like web-mock to mock responses
 - Make changes in the code to allow the project to run without the API

They all have advantages and disadvantages. This commit implements the
last option, adding a repository for Form which can read them locally
from a yaml file in /data when the project is run in SANDBOX. `/data` might not be the best place - these files are almost like test data so they could go in `spec/` they aren't test data though so I've kept them there.

This should make development more independent from the API project and allow testing forms faster.

Add HostingEnivronment

We need to be able to configure the application in different
environments - mainly dev and 'production' ATM.

Environments are modelled in their own class - HostingEnivronment. This
stops config spreading across the app and gives us a simple way to set
policies about what happens in dev/staging and production.

The main draw is to add a sandbox mode to allow running the project
without external dependencies - Notify and the form API.

This commit stops notify being run if SANDBOX=true env var is set before
running the project.

Inspired by
https://github.com/DFE-Digital/apply-for-teacher-training/blob/main/app/lib/hosting_environment.rb

### DIFF
--- a/spec/models/form_spec.rb
+++ b/spec/models/form_spec.rb
@@ -1,15 +1,11 @@
 require "rails_helper"
 
 RSpec.describe Form, type: :model do
-  let(:response_data) { { person: { "id" => 1, "name" => "form name", "submission_email" => "user@example.com" } }.to_json }
-
-  before do
-    ActiveResource::HttpMock.respond_to do |mock|
-      mock.get "/api/v1/forms/1", {}, response_data, 200
-    end
+  it "initializes correctly" do
+    expect(Form.new({id: 1, name: "form name", submission_email: "user@example.com"})).to have_attributes({id: 1, name: "form name", submission_email: "user@example.com"})
   end
 
-  it "returns a simple form" do
-    expect(described_class.find(1)).to have_attributes(id: 1, name: "form name", submission_email: "user@example.com")
+  it "initializes correctly" do
+    expect{Form.new({id: 1, name: "form name", submission_email: "user@example.com", extra_field: 'hetsh'})}.to raise_error
   end
 end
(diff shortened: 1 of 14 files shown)


spec/models/form_spec.rb
```diff
@@ -1,15 +1,11 @@
...
-    ActiveResource::HttpMock.respond_to do |mock|
-      mock.get "/api/v1/forms/1", {}, response_data, 200
-    end
+  it "initializes correctly" do
+    expect(Form.new({id: 1, name: "form name", submission_email: "user@example.com"})).to have_attributes({id: 1, name: "form name", submission_email: "user@example.com"})
   end
 
-  it "returns a simple form" do
-    expect(described_class.find(1)).to have_attributes(id: 1, name: "form name", submission_email: "user@example.com")
+  it "initializes correctly" do
```
danielburnley, line 8: This test seems to test that it raises an error when initialised incorrectly rather than correctly? It also raises a warning in Rspec where you should specify the error it raises

IMO this test is not the most valuable as it's just testing ActiveRecord, but if you want to keep it i'd make sure you check it throws `ActiveModel::UnknownAttributeError` specifically

danielburnley (commented):
Just a smol change - but overall this seems like it might be pretty useful. I'm not sure we'd want to include it for the admin, but for the runner it makes sense to me as a way to easily have a bunch of test data for forms. 

Once the test is changed I'll approve this and we can merge it 🎉

PR #40, opened 2022-06-23
Add X-Robots-Tag header to prevent search indexing

### LINKED ISSUE #39: Prevent search indexing
We need to prevent search indexing so that users don't accidentally access the form pages directly - see [Ensure users start their journey on GOV.UK in the Service Manual](https://www.gov.uk/service-manual/technology/get-a-domain-name#ensure-users-start-their-journey-on-govuk). We can either do this with a meta tag in the view or by adding the X-Robots-Tag HTTP header.

We want this to apply to all pages on the runner. Adding the header in AppliccationController will probably make require less maintenance in the long run, as we are marginally more likely to add new layouts than we are to add new controllers that don't inherit from this one.

DavidBiddle:
Closes #39 

#### Checklist

    - [n/a] README.md

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,4 +2,11 @@
 
 class ApplicationController < ActionController::Base
   default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
+  after_action :add_robots_header
+
+private
+
+  def add_robots_header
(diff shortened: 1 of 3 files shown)


danielburnley (approved):
Looks good to me if that's how you do it these days :D

PR #43, opened 2022-06-29
41 allow form completers to submit their answers

thomasiles:
Add forms to runner

This pull request adds forms to the pages in the runner.

It allows users to enter information so that it can be submitted by notify at the end of the jounrey.

Because it was hard to add forms without considering how validation, saving data in the session and CYA would work. The basics of these features have also been added.

The general approach is to create a question model and a partial to represent it in the form. This gives good control and flexibility over creating new question types and validating them. Partials might not be the best way to handle views - it might be interesting to invert things and give questions full control over the form they are presented in.

All current page types have been added, but the validation for email address, postcode and NINO hasn't been completed as we have another ticket in for those. The CYA screen is also only for demonstration purposes and doesn't implement all the functionality needed as there's a ticket for that too. Submission etc could also be tidied up and fleshed out in the future.

There currently aren't any tests, I'll add those in Friday.
#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -1,5 +1,4 @@
 DATABASE_USERNAME=forms_runner_development
 DATABASE_PASSWORD=forms_runner_development
 DATABASE_HOST=localhost
-NOTIFY_API_KEY=
 API_BASE=http://localhost:9292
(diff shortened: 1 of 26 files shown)


DavidBiddle (approved):
Looks good and works locally - going to merge now so I can pull it into my branch and raise a PR for the check your answers stuff

PR #48, opened 2022-07-11
Improve the validation of question pages

thomasiles:
https://trello.com/c/dsKOQ2fe/305-prevent-form-completers-from-entering-invalid-information

This PR covers the basic validation in the trello card above and leans on the design system pattern pages to implement different types of inputs we use.

Error messages have been moved into the locales file and that, along with the tests is probably the best. overview of whats been implemented.

There are unit tests to cover the validations, but are some properties which would probably benefit from cypress tests.

Everything here is up for discussion -  this is just to get us started with the basics. The design team want to have that conversation too, so once this is merged, I think we can show it to them and then go over what we want to implement and when.

The validation messages themselves are just placeholders.

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)

### DIFF
--- a/app/models/question/date.rb
+++ b/app/models/question/date.rb
@@ -1,10 +1,84 @@
 class Question::Date < Question::QuestionBase
-  include ActiveRecord::AttributeAssignment
+  attribute :date_day
+  attribute :date_month
+  attribute :date_year
+  attribute :date
 
-  attribute :date, :date
-  validates :date, date: true
+  validate :date_valid
+
+  def assign_attributes(new_attributes)
+    translated_attrs = new_attributes.transform_keys { |key| date_field_to_attribute(key) }
+    super(translated_attrs)
+  end
+
+  def date
+    valid_or_invalid_date(date_year, date_month, date_day)
+  end
 
   def show_answer
-    date&.strftime("%d/%m/%Y") || ""
+    if date.is_a?(Date)
+      date.strftime("%d/%m/%Y")
+    else
+      ""
+    end
+  end
+
+private
+
+  def date_valid
+    return errors.add(:date, :blank) if blank?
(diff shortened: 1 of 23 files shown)


app/models/question/date.rb
```diff
@@ -1,10 +1,84 @@
...
+  end
+
+  def blank_fields
+    return [] if date.is_a?(Date)
+
+    date.to_h.select { |_, v| v.blank? }.keys
+  end
+
+  def outside_acceptable_date_range
+    !(150.years.ago.year..2999).cover?(date.year)
```
DavidBiddle, line 61: This range seems plausible for most use cases, but there might be some unusual cases in e.g. the Land Registry that involve dates prior to this. I think we can leave this as is for now, but maybe something to bear in mind as we add more beta partners

DavidBiddle (approved):
Looks good to me! Left one comment about a minor content change and one optional note on the date range

app/models/question/phone_number.rb
```diff
@@ -1,4 +1,28 @@
 class Question::PhoneNumber < Question::QuestionBase
+  # https://design-system.service.gov.uk/patterns/telephone-numbers/
+  # Allow a large range of values - we would probably need more information to
+  # provide better validation - e.g. UK or international number
+  # We check the characters and that the number is in a reasonable range
+  # some code and tests inspired by git@github.com:DFE-Digital/apply-for-teacher-training.git
```
danielburnley, line 6:
```suggestion
  # some code and tests inspired by https://github.com/DFE-Digital/apply-for-teacher-training
```

app/models/question/date.rb
thomasiles: I think you are right - it's a pretty sharp edge for a from creator to put date down as answer type and find that form-fillers can't add information. I'll bring it up with the design team.

PR #58, opened 2022-08-01
Fix asset build by adding URL variables to stylesheet

### LINKED ISSUE #69: Fix asset rendering bug
Currently the govuk-frontend assets (most noticeably the GDS Transport font) aren't being compiled by the asset build. This means that the fonts and assets will load correctly on a local machine, but return 404s on the dev server.

DavidBiddle:
Closes #69 

The same change will be needed in the Admin app once this is merged and I've confirmed that it fixes the issue.

#### Checklist

- [n/a] I've written unit tests for these changes (if code change)
    - [n/a] README.md


thomasiles:
I'm not sure this actually fixes it - I ran up against the same issue last week.
When running the following command:
`RAILS_ENV=production SECRET_KEY_BASE=assets bundle exec rake assets:precompile`
I hit:
`Sprockets::DoubleLinkError: Multiple files with the same output path cannot be linked ("fonts/bold-affa96571d-v2.woff")`

I think its because the assets are already linked by `\config/initializers/assets.rb`:
```
Rails.application.config.assets.paths << Rails.root.join("node_modules/govuk-frontend/govuk/assets")
```
Happy for it to be merged if thats the best way to test it though. I compiled the assets as above and then used the following to run things as they are (ish) :in production.
```
RAILS_SERVE_STATIC_FILES=1 PORT=3002 RAILS_ENV=production SECRET_KEY_BASE=assets API_BASE=http://localhost:9292/ rails s
```

DavidBiddle:
@thomasiles 
Strangely I don't see the `Sprockets::DoubleLinkError`, but when I run `RAILS_SERVE_STATIC_FILES=1 PORT=3002 RAILS_ENV=production SECRET_KEY_BASE=assets API_BASE=http://localhost:9292/ rails s` I still see the issue so I think you're right about this not fixing it.

I think what might be happening is that because we're not using the Rails asset pipeline for our Sass build, Rails isn't seeing the `url()` calls in the Sass, so can't use its custom `url` function to put the SHA digest into the filename. So the font files exist at `/assets/fonts/bold-affa96571d-v2-088c8da9d0319b63b23ac19791f37a33ec0d4b61f884ab56123bfd96d8312153.woff`, but the CSS is calling `/assets/fonts/bold-affa96571d-v2.woff`.

If that's right then I guess our options here are:
1. Keep our build as is but turn off the digest
2. Rework our build to use the asset pipeline properly
3. Rework our build to avoid the asset pipeline completely

Option 1 is probably easiest and quickest but is also quite likely to cause caching issues.
Options 2 and 3 are better, but probably quite painful.

DavidBiddle:
@thomasiles It turned out to be deceptively simple in the end - just had my URLs set wrong!

thomasiles (approved):
Nice one! I didn't think it would take such a small change

aliuk2012 (approved):
Could we tidy up the commit history?  Apart from that all looks good

DavidBiddle:
@aliuk2012 I'm going to do a Squash and Merge so think that should be handled

PR #60, opened 2022-08-02
Add draft watermark

DavidBiddle:

Adds "Draft" water mark to all form pages, CYA and submitted form if the form is a draft. 

Trello card: https://trello.com/c/zoTsIRKe/433-show-that-a-form-is-in-a-draft-state-in-the-runner-watermark

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)


aliuk2012:
From an a11y POV, shouldn't there be a bigger difference or at least content about the fact they are viewing a preview-form.

aliuk2012:
Examples:

image

image
image
image

aliuk2012:
@deberny @christophercameron-ixd  Heres some screenshots of the draft watermark, could you review form design POV?

deberny:
It looks good to me mostly. The thing I'm wondering about is the difference between the first screenshot, where the draft is smaller and the second screenshot. The second screenshot with larger draft seems more prominent and better? Is the 'Draft' size dictated by a page height, then?

DavidBiddle:
@deberny Yeah, it's based on the height of the main page content which is also how Whitehall does it. I guess the reasoning for that is so that if the page is too short the whole word 'draft' will still be present rather than potentially being cut off but might be a good thing to explore alternatives for at some point.

@aliuk2012 I can't approve this because I'm officially the PR author but the bits you've added look good!

aliuk2012 (approved):

PR #65, opened 2022-08-05
Error pages

### LINKED ISSUE #64: Add friendly error pages
This should include:
404 for page not found
500 for unexpected errors
Service unavailable (in case we need to take the service down temporarily)
Notify error (special case of the 500 error with some extra text about the submission failing)

DavidBiddle:
Closes #64 

#### Checklist

    - [x] README.md


config/routes.rb
```diff
@@ -15,4 +15,7 @@
   end
 
   get "/:form_id/:page_id/change" => "page#show", as: :change_form_page, defaults: { changing_existing_answer: true }
```
thomasiles, line 17: Does it make sense to move this under the form resrouces section - eg:
```
  get "/:page_id/change" => "page#show", as: :change_form_page, defaults: { changing_existing_answer: true }
```

and then update the path helper in the check your answers page to:
```
-        actions: [{ href: change_form_page_path(form, page), visually_hidden_text: question_name }],
+        actions: [{ href: form_change_form_page_path(form, page), visually_hidden_text: question_name }],
```

There may be reasons not to?

thomasiles (approved):
Looks good! 

There is one comment about the routes and I'll update the code in the PR I'm doing to make sure the right errors - 404/500 are being thrown depending on the form/page existing

config/routes.rb
DavidBiddle: No, that works very nicely! I've called it `form_change_answer_path` to make it slightly less unwieldy but this is much nicer now 👍

thomasiles (approved):

PR #70, opened 2022-08-09
Add journey validation

thomasiles:
# Add journey validation

Previously the runner only validated individual pages in isolation. Nothing prevented skipping ahead pages of missing parts of the form out when they submit it.

I've changed the testing and layout of the project in order to make sure the user can only visit pages in the order they have been designed.
  

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)


DavidBiddle (approved):
Nice one! Especially like way you implemented the slug bits.

This will cause some conflicts with my anlytics branch but that's fine - feels like it'll be lot easier to resolve on my branch so we should merge this one first.

thomasiles:
> Nice one! Especially like way you implemented the slug bits.
> 
> This will cause some conflicts with my anlytics branch but that's fine - feels like it'll be lot easier to resolve on my branch so we should merge this one first.

Yeah - sorry I think it needed reorganising but I will avoid long living branches and big changes in the future. I am happy to change the analytics branch if that helps.

DavidBiddle:
Nah it's not a problem, and not your fault - was always going to be a conflict I think, and should be a manageable one.

PR #83, opened 2022-08-16
api_key

alice-carr:
I made some silly errors in the last PR so have opened a new one

- passes api key via headers, 
- passes key into paas deploy and 
- updates tests to pass in headers

#### Checklist

- [ ] I've used the pull request template
- [ ] I've linked this PR to the relevant issue (if mission work)
- [ ] I've written unit tests for these changes (if code change)
- [ ] I've updated the documentation in (If any documentation requires updating)
    - [ ] README.md
    - [ ] Elsewhere (please link)

PR #84, opened 2022-08-16
Revert "api_key"

alice-carr:
Reverts alphagov/forms-runner#83 - the deploy was failing with an error saying an env var wasn't supplied. Will revert to debug.


0atman (approved):
how odd

PR #90, opened 2022-08-23
Add privacy policy link

DavidBiddle:
Reads a form's privacy policy URL and displays it in the footer if it's present.

Trello card: https://trello.com/c/aZtUyjei/16-add-link-to-privacy-information-to-forms-runner-in-footer-of-every-form-pageneeds-dev-review

#### Checklist

    - [n/a] README.md

### DIFF
--- a/app/controllers/form_controller.rb
+++ b/app/controllers/form_controller.rb
@@ -1,6 +1,7 @@
 class FormController < ApplicationController
   def show
     @form = Form.find(params.require(:id))
+    set_privacy_policy_url
     if @form.start_page
       redirect_to form_page_path(params.require(:id), @form.start_page)
       EventLogger.log_form_event(Context.new(form: @form, store: session), request, "visit")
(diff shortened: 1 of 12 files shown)


app/controllers/form_controller.rb
```diff
@@ -9,5 +10,12 @@ def show
 
   def submitted
     @form = Form.find(params.require(:form_id))
+    set_privacy_policy_url
+  end
+
+private
+
+  def set_privacy_policy_url
+    @privacy_policy_url = @form.privacy_policy_url
```
aliuk2012, line 19: 🤔  Is it worth maybe setting `@privacy_policy_url` from the main ApplicationController so we shouldn't have to do it in specific controllers. We could also set a default to our own privacy policy for instances like 404/500 pages

thomasiles (approved):
looks good!

app/controllers/form_controller.rb
DavidBiddle: Probably a good idea but I'm running into something tricky locally with the `current_context` stuff, so I'm going to call it a day for now - can have another look when I'm back next week, or happy for someone else to pick this up in the meantime if we want to get it done by end of sprint.

aliuk2012: Ah I see what you mean, there is a lot of code hanging over from initial prototyping meaning we have things like `form_controller.rb` and `forms/form_controller.rb` and a few other controllers that we dont need any more.

I'm happy to push this through and we can tweak this as part of tidying up all the controllers just generally.

PR #92, opened 2022-08-24
Fix sessions

thomasiles:
Add a few security measures to protect users sessions in the runner:
- set `force_ssl=true` in production, this sets secure=true on cookies, so they are only ever transported over ssl, also HSTS
- set TTL on Redis session keys to 20hours. This means sessions will expire after 20hours of no requests to the server from the users browser. Cookies are still session based, so are not kept around after the browser closes.
- cleans up the config of redis and tries to prevent accidental misconfiguration of the session store

There are a few other session based things which we might consider doing - I'm also documenting current behaviour in a sample session and I'll bring a few of them up for review.

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -90,6 +90,7 @@ group :test do
   gem "selenium-webdriver"
   gem "simplecov"
   gem "webdrivers"
+  gem "climate_control"
 end
 
 gem "bundler-audit", "~> 0.9.0"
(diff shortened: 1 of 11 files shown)


aliuk2012 (approved):
Some good tweaks to the config that are needed. 👍

PR #109, opened 2022-09-06
Add cookies link to footer

DavidBiddle:
Adds a cookie page, and links to that page in the footer. Some of the content isn't finalised, so may change, but functionally this should be complete.

Trello card: https://trello.com/c/wLbTy34H/55-draft-and-publish-cookie-information-to-cover-the-product-product-pages-and-forms

#### Checklist

    - [n/a] README.md

### DIFF
--- a/app/views/help/cookies.html.erb
+++ b/app/views/help/cookies.html.erb
@@ -1,0 +1,21 @@
+<% set_page_title("Cookies") %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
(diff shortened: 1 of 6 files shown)


aliuk2012 (approved):
Nice work, we might want to think about how the user goes back to the form they came from otherwise they will be stuck on these pages unless they user browser back button.

aliuk2012 (approved):
Nice work, we might want to think about how the user goes back to the form they came from otherwise they will be stuck on these pages unless they user browser back button.

PR #112, opened 2022-09-08
Fix formatting of content on the accessibility page

aliuk2012:
#### Before
image

#### After 
image

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)


DavidBiddle:
I don't think this should be required - see https://forms-runner-dev.london.cloudapps.digital/help/accessibility-statement

Issue here is that in #104 I added

```
$govuk-new-link-styles: true;
$govuk-global-styles: true;
```

to the stylesheet, so the link and body classes are no longer necessary - I think you have an old version of the stylesheet, try `rails assets:clobber` and then restarting your server and you should hopefully see tyhe styles resolve themselves. Let me know if that's not the case!

aliuk2012:
This PR is not required, thanks @DavidBiddle for confirming 😄

PR #115, opened 2022-09-15
Add preview routes

DavidBiddle:

Adds 'preview' routes to the runner. If a user accesses a form via these routes, no analytics events will be logged. 

Trello card: https://trello.com/b/9cKfKKdR/private-beta-sprint-board

#### Checklist

    - [n/a] README.md


config/routes.rb
```diff
@@ -6,17 +6,28 @@
...
-  get "/form/:form_id/:page_slug" => "forms/page#show", as: :form_page
-  post "/form/:form_id/:page_slug" => "forms/page#save", as: :save_form_page
+    get "/:id" => "form#show", as: "#{prefix}form"
+    get "/:form_id/check_your_answers" => "forms/check_your_answers#show", as: "#{prefix}check_your_answers"
+    post "/:form_id/submit_answers" => "forms/submit_answers#submit_answers", as: "#{prefix}form_submit_answers"
+    get "/:form_id/submitted" => "forms/submitted#submitted", as: "#{prefix}form_submitted"
+    get "/:form_id/:page_slug/change" => "forms/page#show", as: "#{prefix}form_change_answer", defaults: { changing_existing_answer: true }
+    get "/:form_id/:page_slug" => "forms/page#show", as: "#{prefix}form_page"
+    post "/:form_id/:page_slug" => "forms/page#save", as: "#{prefix}save_form_page"
+  end
```
DavidBiddle, line 22: Not 100% sure about this bit - can be done in a less convoluted but more repetitive way by reverting this commit: https://github.com/alphagov/forms-runner/pull/115/commits/efb8d6ac3f009500ace5622e3375ebf625536bc4

config/routes.rb
```diff
@@ -6,17 +6,28 @@
...
+
+  scope "/form" do
+    concerns :previewable_routes
+  end
+
+  scope "/preview-form", defaults: { preview: true } do
```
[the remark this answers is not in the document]
DavidBiddle: I don't think we've really explored it in any depth - I just went with the format suggested on the card. 
We discussed it briefly in the tech standup this morning, and I think the general conclusion was that using the path would make it less likely that the user would remove the query params, but I don't feel particularly strongly either way, so happy to hear what other people think on this

config/routes.rb
thomasiles: I might be totally wrong about this but can we reduce some of the repetition, particularly url_helpers, by using parameters in the routes?

This would allow us to thread a parameter through the requests, so we don't need to switch on the preview param every time we need to generate a url/path and save us a bit of hassle making sure we stay in preview (or not!).

This is the same technique [rails suggests for using locales in urls](https://guides.rubyonrails.org/i18n.html#managing-the-locale-across-requests).

In routes, we could use:

```
    scope "/:mode", locale: /preview-form|form/ do
    get "/:id" => "form#show", as: "form"
    get "/:form_id/check_your_answers" => "forms/check_your_answers#show", as: "check_your_answers"
    post "/:form_id/submit_answers" => "forms/submit_answers#submit_answers", as: "form_submit_answers"
    get "/:form_id/submitted" => "forms/submitted#submitted", as: "form_submitted"
    get "/:form_id/:page_slug/change" => "forms/page#show", as: "form_change_answer", defaults: { changing_existing_answer: true }
    get "/:form_id/:page_slug" => "forms/page#show", as: "form_page"
    post "/:form_id/:page_slug" => "forms/page#save", as: "save_form_page"
  end
```

This would generate urls (/:mode)/:form_id/:page_slug - i.e /preview-form/1/1 or /form/1/1 and setting a parameter called mode to either "form" or "preview-form"


In our form controller, which the other controllers inherit from we could add:
```
    def preview?
      params[:mode] == "preview-form"
    end

    def mode
      params[:mode]
    end

    def default_url_options
      { mode: mode }
    end
```

`default_url_options` sets values for every url_helper - so `form_page_path` in page controller will generate a url for `prevew-form` or `form` depending on the current value of mode. This saves us having to use `params[:preview] ? preview_url : normal_url`.

Adding `preview?` lets us do logging etc in real mode only by adding `unless preview?` to the end of logging statements etc.

I don't know if this is any better but might cut down on making sure the paths are correct?

config/routes.rb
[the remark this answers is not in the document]
thomasiles: I think a dedicated path segment is cleaner and brings it forwards in the URL to show it's important - I'm not sure that there are any cons - maybe params would be better if we had multiple types of previews?

thomasiles:
Random thought: should we keep session data separate in preview/live? I guess filling out half a form in preview `/preview-form/`, then switching to a real version `/form/1`, would jump you to the last page you filled in with all the previous answers complete.

config/routes.rb
DavidBiddle: That sounds like a sensible idea, I'll give it a go!

aliuk2012 (approved):
❤️  Awesome work. 

@thomasiles Good point,  I think we should keep the session data separate so its almost as if forms-runner is actually part of forms-admin and sessions dont get mixed up. Lets do that as a new another ticket.  its a nice iteration on this preview-mode feature

DavidBiddle:
Going to merge this as it feels fairly big already, so agree that we should tackle the session separation in another ticket.

I think the Rails routing system makes it relatively easy to switch from a path segment to a query parameter, so if we have to rework this we can, but for now I think we can try it this way.

PR #116, opened 2022-09-15
Make submission time in emails use local timezone

thomasiles:
https://trello.com/c/XNOhXuOY/128-bug-submission-time-in-the-processing-email-is-off-by-an-hour

Submission emails show the time of submission in UTC zone - which the server uses. This commit changes the timezone to use a zone specified in the config and defaults to London. It also adds tests to the submission code.

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)

### DIFF
--- a/spec/services/notify_service_spec.rb
+++ b/spec/services/notify_service_spec.rb
@@ -1,0 +1,48 @@
+require "rails_helper"
+
+RSpec.describe NotifyService do
+  let(:notify_api_key) { nil }
+  around do |example|
+    ClimateControl.modify NOTIFY_API_KEY: notify_api_key do
+        example.run
+    end
+  end
+
+  context "with api key set" do
+    let(:notify_api_key) { 'test-key' }
+    let(:submission_datetime) { Time.utc(2022, 9, 14, 10, 00, 00) }
+
+    it "sends correct values to notify" do
+      fake_notify_client = instance_double(Notifications::Client)
+      allow(fake_notify_client).to receive(:send_email)
+      allow(Notifications::Client).to receive(:new).and_return(fake_notify_client)
+
+      travel_to submission_datetime do
+        notify_service = NotifyService.new
+        notify_service.send_email('fake-email','title','text')
+        expect(fake_notify_client).to have_received(:send_email).with(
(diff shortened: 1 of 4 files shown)


spec/services/notify_service_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      allow(Notifications::Client).to receive(:new).and_return(fake_notify_client)
+
+      travel_to submission_datetime do
+        notify_service = NotifyService.new
+        notify_service.send_email('fake-email','title','text')
+        expect(fake_notify_client).to have_received(:send_email).with(
+          {:email_address=>'fake-email',
+           :personalisation=>{
+             :submission_date=>"14 September 2022",
+             :submission_time=>"11:00:00",
```
danworth, line 27: Just getting to grips with Ruby and Rails so I could be way off, also time zones shenanigans... Will this test start failing once we leave BST in London? If I've understood the test setup, we're configuring it to run with London timezone (currently UTC +1 ) and setting the `submission_datetime` to 10:00 UTC on line 13. Once we enter GMT in London won't we expect the `subsmission_time` in the email to actually be `10:00:00` rather than `11:00:00`?

thomasiles: I get really confused by timezones! The behaviour I'm aiming to test is this:
- a Londoner submits a form and looks at the clock and sees the timedate as 14th of sept 2022, 11am.
- The email submission has the same time date the Londoner saw

The test mocks the date as 14/9/2022 10:00 UTC - won't that date always be BST in the london timezone, no matter the current date?

DavidBiddle: I think to be extra robust we could test 2 things:
- if a user submits the form at 10:00 UTC in summer then the email should say 11:00
- if a user submits the form at 10:00 UTC in winter then the email should say 10:00

Looks like the existing test already covers the first case and we could probably live without the second but it's a nice to have

aliuk2012: 💯 agree just need an extra test for winter times ;)

DavidBiddle (approved):
Looks good to me 🎉

PR #123, opened 2022-09-23
Refactor application based endpoints

aliuk2012:
It seemed pointless having an empty controller floating around which served as a placeholder for the two routes that were really just part of the overall application.

I know that rails is clever enough for us not to define an action in in the controller if the action name and view name are the same and it doesn't need to set any vars etc. but this might confuse someone who isn't familar with rails and this sort of pattern.
#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)


DavidBiddle (approved):
Looks good and seems like a sensible change 🎉

PR #139, opened 2022-09-26
Add UI and logic for contact details

DavidBiddle:

Trello card: https://trello.com/c/7u2OgU0G/19-add-form-specific-contact-details-to-forms-runner-should-have

#### Checklist

    - [n/a] README.md


app/views/forms/page/show.html.erb
```diff
@@ -15,5 +15,30 @@
...
+        <% end %>
+        <% if @support_details.url.present? && @support_details.url_text.present? %>
+          <h3 class="govuk-heading-s"><%= t("support_details.online") %></h3>
+          <p class="govuk-body">
+            <%= govuk_link_to t("support_details.link_text", link: @support_details.url_text), @support_details.url, target:"_blank", rel: "noopener noreferrer" %>
+          </p>
+        <% end %>
+      <% end %>
+    <% end %>
+
```
aliuk2012, line 42: Could we many abstract this out into its own component which we could easily test different data setups? Also as a component we can easily setup a render condition so the view doens't have to have a big chunky condition in it

aliuk2012 (commented):
Overall it looks good.  Good to see a fly-by fixup of the linting issues. Could you also update this line so it picks up all our code 

https://github.com/alphagov/forms-runner/blob/main/.github/workflows/rubyonrails.yml#L50
Should be a case of just removing `app` from that command.

app/views/forms/page/show.html.erb
DavidBiddle: Done! Good suggestion

DavidBiddle:
@aliuk2012 I did the rubocop change you suggested but then reverted it - it caused a lot of extra linting errors which I didn't see all of locally, so it felt like something that should be tackled in a separate PR.

aliuk2012 (approved):

PR #140, opened 2022-09-27
Fix pages back link bug

aliuk2012:
the back link did not include the form slug in it and the tests that were written were not detecting this and returning false positive. The reason it was returning a false positive was because the get request in the redirect was included a message with a path to the redirect which happened to be the correct path but the page being rendered after the redirect did not include it.

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)


alice-carr (approved):
looks good!

PR #145, opened 2022-09-30
Switch settings config

aliuk2012:
Part of this discussion: https://github.com/alphagov/forms/discussions/43

At the moment, we dont have a single place where we document our environment variables. Some variables might not even need to be specific environment variables but more of a case of just a config setting for the application (i.e like feature flags) 

Moving to yaml config allows us to specify exactly what settings we have and we can config it for each environment. As this gets loaded in as a ruby object it also means that we can toggle feature flags or variables on the fly without restarting the server. It also opens us up to being able to create a super admin ui for other non-dev means in the team to be able to control whether a feature is enabled or not using a UI. See the config gem readme for more info but basically its a case of logging into a console using updating the setting we would like to change followed by running `Settings.reload!` and the server will then use the new Setting rather than the old one.

We have a lot of different environment settings and rather than bloating this PR with all the changes (see https://github.com/alphagov/forms-admin/pull/118 as an example of it bloating 🙈 ). I would like to propose a rollout process where we migrate settings over to the new format. 

As you can also see we can add specs to ensure that settings are not just removed from the yaml file or that yaml file is actually valid and returning the  options we are expecting.

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)


README.md
```diff
@@ -39,15 +39,38 @@ from the [notify service](https://www.notifications.service.gov.uk/) Add it as
 an environment vairable under `NOTIFY_API_KEY=` in `.env.development.local` and
```
DavidBiddle, line 39: Feels like this `NOTIFY_API_KEY` var should be documented in the 'Environment variables' section below

aliuk2012: TBH, I think that would better off moving to setting.yml config file where it can be documented better. I feel like notify should be setup using a feature flag so regardless as to whether theres api keys set or whatever, it should be "can_send_email" feature is enabled and if so then look up all the details required.

aliuk2012: This is the whole point about these piece of work is that there are a few other env vars we have in our app but we haven't documented them in one particular place which this piece of work should help to improve.

DavidBiddle: Yeah, guess I was just flagging that one because it's already in the README. Happy for you to resolve this either way though

thomasiles (approved):
Looks much cleaner, nice one.

aliuk2012:
@DavidBiddle  could you review this one again. I've rebase it from `main` (which already has the config gem) so this PR is now switching forms API base and API key to use settings. Let me know if you want to go through this one.

DavidBiddle (approved):
Looks good to me Al 🎉

PR #146, opened 2022-10-03
Always precompile assets when building image

danworth:
Even to run the local dev setup the `rails assets:precompile` needs to have been run so remove the condition within the Dockerfile.

Make it simpler to run forms locally.

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,8 +31,7 @@ COPY --chown=ruby:ruby . .
 
 # you can't run rails commands like assets:precompile without a secret key set
 # even though the command doesn't use the value itself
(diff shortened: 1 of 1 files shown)


DavidBiddle:
This might not be necessary - you might be able to get the asset build working locally by running `yarn run build` and `yarn run build:css` on the local machine, which will build the assets without copying them into the public folder. In development mode, Rails will still pick these up.

(if you already tried that then ignore this)

thomasiles (approved):
👍

PR #153, opened 2022-10-10
Added long_text option

0atman:
Created a new 5000 character field

API PR https://github.com/alphagov/forms-api/pull/92

### DIFF
--- a/spec/models/question/long_text_spec.rb
+++ b/spec/models/question/long_text_spec.rb
@@ -30,9 +30,7 @@ RSpec.describe Question::LongText, type: :model do
 
   context "when given a string which is too long" do
     it "validates without errors" do
-      question = described_class.new
-      question.validate
-      question.text = "a" * 5001
+      question = described_class.new(text: "a" * 5001)
       question.validate
       expect(question.errors[:text]).to include(I18n.t("activemodel.errors.models.question/long_text.attributes.text.too_long"))
     end
--- a/spec/models/question/long_text_spec.rb
+++ b/spec/models/question/long_text_spec.rb
@@ -22,9 +22,7 @@ RSpec.describe Question::LongText, type: :model do
 
   context "when given a string" do
     it "validates without errors" do
-      question = described_class.new
-      question.validate
-      question.text = "testing"
+      question = described_class.new(text: "testing")
       question.validate
       expect(question).to be_valid
     end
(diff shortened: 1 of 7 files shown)


DavidBiddle:
You can fix the broken Docker build by changing `nodejs=16.16.0-r0` --> `nodejs=16.17.1-r0` in the Dockerfile

DavidBiddle:
There's no view for this field type in `app/views/question` - which one does it render? Let me know if you need an FE to add one

spec/models/question/long_text_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      question = described_class.new
+      expect(question.show_answer).to eq ""
+    end
+  end
+
+  context "when given a string" do
+    it "validates without errors" do
+      question = described_class.new
+      question.validate
+      question.text = "testing"
```
aliuk2012, line 27:
```suggestion
      question = described_class.new(text: "testing")
```

spec/models/question/long_text_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      question.validate
+      expect(question).to be_valid
+    end
+  end
+
+  context "when given a string which is too long" do
+    it "validates without errors" do
+      question = described_class.new
+      question.validate
+      question.text = "a" * 5001
```
aliuk2012, line 37:
```suggestion
      question = described_class.new(text: "a" * 5001)
```

spec/models/question/long_text_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      question.validate
+      expect(question.errors[:text]).to include(I18n.t("activemodel.errors.models.question/long_text.attributes.text.too_long"))
```
aliuk2012, line 39:
```suggestion
			expect(question).not_to be_valid
      expect(question.errors[:text]).to include(I18n.t("activemodel.errors.models.question/long_text.attributes.text.too_long"))
```

aliuk2012:
good point @DavidBiddle, might be worth creating the new question type as a component rather than a partial which we could test better in isolation.

spec/models/question/long_text_spec.rb
alice-carr: Thanks for your suggestions @aliuk2012 - I'll just commit them now

DavidBiddle:
@aliuk2012 will do a separate PR for turning the partial views into components

aliuk2012 (approved):

PR #158, opened 2022-10-14
Add number answer type

DavidBiddle:
Adds a new number input type. This can be any whole or decimal number greater than or equal to 0.

Trello card: https://trello.com/c/tyDQQZ6A/222-add-number-answer-to-a-question-in-the-runner-and-api

#### Checklist

    - [n/a] README.md

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -36,8 +36,8 @@ en:
         question/number:
           attributes:
             number:
-              blank: Enter an answer
-              greater_than_or_equal_to: The answer must be at least 0
+              blank: Enter a number
+              greater_than_or_equal_to: The answer must be greater than or equal to 0
               not_a_number: The answer must be a number
         question/phone_number:
           attributes:
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -60,7 +60,7 @@ en:
     submitted:
       title: Your form has been submitted
   page_titles:
-    error_prefix: "Error: "
+    error_prefix: 'Error: '
   phase_banner:
     after_link: will help us improve it.
     before_link: This is a new service, your
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -33,6 +33,12 @@ en:
             national_insurance_number:
               blank: Enter a National Insurance number
               invalid_national_insurance_number: Enter a National Insurance number that is 2 letters, 6 numbers, then A, B, C or D, like QQ 12 34 56 C
+        question/number:
+          attributes:
+            number:
+              blank: Enter an answer
+              greater_than_or_equal_to: The answer must be at least 0
+              not_a_number: The answer must be a number
         question/phone_number:
           attributes:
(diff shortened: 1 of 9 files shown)


spec/services/notify_service_spec.rb
```diff
@@ -22,7 +22,7 @@
 
         travel_to submission_datetime do
```
DavidBiddle, line 23: (All the changes in this file are automatic Rubocop formatting changes)

spec/requests/base_spec.rb
```diff
@@ -280,10 +280,6 @@
           it "Render the not found page" do
             expect(response.body).to include(I18n.t("not_found.title"))
           end
-
```
DavidBiddle, line 283: Rubocop suggested this (it's a duplicate test - there's an identical one earlier in the file that's not shown in this preview)

config/locales/en.yml
```diff
@@ -33,6 +33,12 @@ en:
             national_insurance_number:
               blank: Enter a National Insurance number
               invalid_national_insurance_number: Enter a National Insurance number that is 2 letters, 6 numbers, then A, B, C or D, like QQ 12 34 56 C
+        question/number:
+          attributes:
+            number:
+              blank: Enter a number
+              greater_than_or_equal_to: The answer must be greater than or equal to 0
```
danworth, line 40: At risk of being pedantic but maybe `greater_than_or_equal_to_zero`?

danworth (approved):
I've been trying this out locally in addition to the tests included and it looks good to me. I haven't looked at what comes out of Notify but if there's no reason to suspect this adds anything to what @aliuk2012 is already working on then I think that's ok.

I noticed that we keep the leading 0s such as `0004` which I guess is the desired outcome? Also it allows `-0` which seems fine and I was just trying things out, led me down a rabbit hole of wondering if negative zero is even a thing, turns out it might be: https://en.wikipedia.org/wiki/Signed_zero#:~:text=The%20notation%20%22%E2%88%920%22%20may,statistical%20mechanics%20and%20other%20disciplines.

config/locales/en.yml
DavidBiddle: (but you're right, that would be better)

danworth: Ah cool, didn't know it was auto generated 👍

spec/models/question/number_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+RSpec.describe Question::Number, type: :model do
+  it_behaves_like "a question model"
+
+  context "when given an empty string or nil" do
+    it "returns invalid with blank message" do
+      question = described_class.new
+      question.validate
+      expect(question.errors[:number]).to include(I18n.t("activemodel.errors.models.question/number.attributes.number.blank"))
+      question.number = ""
+      question.validate
```
aliuk2012, line 13: We can simplify this ;-)

PR #161, opened 2022-10-14
Wrap answers in a three back tick block

aliuk2012:
This seems to escape all markdown characters without having to replace any of the users data. We do need to test this technique with multi-Line question types because it might not work as expected when extra new line breaks are involved'

image

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)


thomasiles (approved):
I've given this aa bit more manual testing and it seems good - merging and escaping multi-line questions separately.

PR #167, opened 2022-10-24
249 form runner log events for when a user completes skips optional question

SamJamCul:
https://trello.com/c/DmZWSCEK/249-form-runner-log-events-for-when-a-user-completes-skips-optional-question

Adding a log for when users skip or complete an optional question. This has been broken into two parts, one that adds a log key for a question being optional or not, and another for when a question is skipped, currently only possible if it's optional. 

The log used to show up like this - `[page_save] {"url":"http://localhost:3001/preview-form/1/test/7","method":"POST","form":"test","question_text":"phone num"}`

And now will show up like this if it's an optional question that isn't skipped - `[optional_save] {"url":"http://localhost:3001/preview-form/1/test/7","method":"POST","form":"test","question_text":"phone num","optional_question":"true","skipped_question":"false"}`

And like this when it is skipped - `[optional_save] {"url":"http://localhost:3001/preview-form/1/test/7","method":"POST","form":"test","question_text":"phone num","optional_question":"true","skipped_question":"true"}`

There are other page event names that have also been conditioned to match whether it's an optional question `"change_answer_optional_save" and `"first_optional_save"`

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)

### DIFF
--- a/app/controllers/forms/page_controller.rb
+++ b/app/controllers/forms/page_controller.rb
@@ -13,7 +13,7 @@ module Forms
 
       if current_context.save_step(@step)
         unless preview?
-          log_page_save(@step, request, changing_existing_answer, page_params)
+          LogEventService.new(current_context, @step, request, changing_existing_answer, page_params).log_page_save
         end
         redirect_to next_page
       else
@@ -52,30 +52,6 @@ module Forms
       else
         form_page_path(@step.form_id, @step.form_slug, @step.next_page_slug)
       end
-    end
-
-    def is_starting_form(step)
-      current_context.form_start_page == step.id
-    end
-
-    def log_page_save(step, request, changing_answer, answers)
-      EventLogger.log_page_event(current_context, step, request, log_event(changing_answer, step), skipped_question?(answers))
-    end
-
-    def log_event(changing_answer, step)
-      page_optional = @step.question.is_optional? ? "optional" : "page"
-
-      if changing_answer
-        "change_answer_#{page_optional}_save"
-      elsif is_starting_form(step)
-        "first_#{page_optional}_save"
-      else
-        "#{page_optional}_save"
-      end
-    end
-
-    def skipped_question?(answers)
-      @step.question.is_optional? ? answers.each_value.map.none?(&:present?) : nil
     end
   end
 end
--- a/app/controllers/forms/page_controller.rb
+++ b/app/controllers/forms/page_controller.rb
@@ -13,7 +13,7 @@ module Forms
 
       if current_context.save_step(@step)
         unless preview?
-          log_page_save(@step, request, changing_existing_answer)
+          log_page_save(@step, request, changing_existing_answer, page_params)
         end
         redirect_to next_page
       else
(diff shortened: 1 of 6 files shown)


app/controllers/forms/page_controller.rb
```diff
@@ -67,7 +67,12 @@ def log_page_save(step, request, changing_existing_answer)
                     "page_save"
                   end
 
-      EventLogger.log_page_event(current_context, step, request, log_event)
+      EventLogger.log_page_event(current_context, step, request, log_event, skipped_question: skipped_question?(answers))
+    end
+
+    def skipped_question?(answers)
+      answers.each_value { |a| return false if a != "" }
```
DavidBiddle, line 74: Think there are a couple of Rails syntax bits you might be able to use here:

```suggestion
      answers.any? { |a| a.present? }
```

DavidBiddle (approved):
This looks good! Have added one comment where I think you can take advantage of some Rails syntax, but if that doesn't work here feel free to ignore.

app/controllers/forms/page_controller.rb
```diff
@@ -58,16 +58,24 @@ def is_starting_form(step)
...
+      elsif is_starting_form(step)
+        "first_#{page_optional}_save"
+      else
+        "#{page_optional}_save"
+      end
+    end
 
-      EventLogger.log_page_event(current_context, step, request, log_event)
+    def skipped_question?(answers)
+      @step.question.is_optional? ? answers.each_value.map.none?(&:present?) : nil
```
SamJamCul, line 78: this is a bit ugly - but it's basically saying if the question isn't optional, just return `nil`

if it is optional, check through the answers, and if all of them are empty, return true - otherwise false.

app/lib/event_logger.rb
```diff
@@ -13,14 +13,16 @@ def self.log_form_event(context, request, event)
...
-  def self.log_page_event(context, page, request, event)
+  def self.log_page_event(context, page, request, event, skipped_question)
     item_to_log = {
       url: request&.url,
       method: request&.method,
       form: context.form_name,
       question_text: page.question.question_text,
     }
 
+    item_to_log.merge!({ skipped_question: skipped_question.to_s }) unless skipped_question.nil?
```
SamJamCul, line 24: Don't know if I quite like how I've got true, false, and nil all doing things here. I'd rather not use nil, but I'm not sure of a particularly elegant way of covering all of the scenarios at once.

app/controllers/forms/page_controller.rb
```diff
@@ -58,16 +58,24 @@ def is_starting_form(step)
...
-                  elsif is_starting_form(step)
-                    "first_page_save"
-                  else
-                    "page_save"
-                  end
+    def log_page_save(step, request, changing_answer, answers)
+      EventLogger.log_page_event(current_context, step, request, log_event(changing_answer, step), skipped_question?(answers))
+    end
+
+    def log_event(changing_answer, step)
```
aliuk2012, line 65: It might be worth breaking this out into a separate `LogEventService` which the controller would use in the action rather than bloating the controller with extra private methods that only relate to event logging.

SamJamCul: Yeah, that makes sense - I'll do some refactoring

aliuk2012 (approved):
That looks, a lot better simplifying the controller. 👍

PR #169, opened 2022-10-26
Escape long answers

thomasiles:
# Fix escaping and display of long text answer-type

Long text is a new answer type which adds a couple of complications:
  - playing back long text answers in the CYA screen
  - submitting text with line breaks via notify

Answers with line breaks need to show paragraphs on the check your answer
screen. This commit breaks text into paragraphs and removes leading/trailing
space. It condenses multiple newlines into single newlines.

Notify interprets submissions as markdown and doesn't offer a way to safely
escape markdown formatting.

This commit fixes both of these issues.

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)

### DIFF
--- a/app/controllers/forms/check_your_answers_controller.rb
+++ b/app/controllers/forms/check_your_answers_controller.rb
@@ -1,6 +1,7 @@
 module Forms
   class CheckYourAnswersController < BaseController
     def show
+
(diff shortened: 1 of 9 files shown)


DavidBiddle (approved):
Looks good! 🎉

PR #173, opened 2022-11-08
Do not redirect to HTTPS for /ping

danworth:
Our Puma server is not configured for TLS despite the `config.force_ssl = true`. Instead if the request has `X-Forwarded-Proto: https` set then Puma will not redirect to `https` and serve the request. Whenever a request comes via a level 7 router such as an ALB or the PaaS Router it sets the header based on the original client protocol.

On AWS the ALB sends a healthcheck to /ping directly and we cannot set the headers. Stop redirecting /ping to https so we can use it as a healthcheck for now.

with @alice-carr

## Testing
We have built and deployed this to AWS dev account and the ECS task is passing the healthchecks.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -45,6 +45,9 @@ Rails.application.configure do
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
   config.force_ssl = true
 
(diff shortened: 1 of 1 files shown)


alice-carr (approved):
👍

PR #194, opened 2022-11-30
Add selection list answer type

DavidBiddle:
Adds a question type for selections, which can be either checkboxes or radios depending on the value of `answer_settings.allow_multiple_answers`.

Trello card: https://trello.com/c/VLYnibtv/225-add-selection-question-type-to-runner

#### Checklist

  - [n/a] README.md

### DIFF
--- a/app/models/question/selection.rb
+++ b/app/models/question/selection.rb
@@ -45,7 +45,7 @@ module Question
     end
 
     def validate_checkbox
-      return errors.add(:selection, :blank) if selection_without_blanks.empty?
+      return errors.add(:selection, is_optional? ? :both_none_and_value_selected : :checkbox_blank) if selection_without_blanks.empty?
       return errors.add(:selection, :both_none_and_value_selected) if selection_without_blanks.count > 1 && :none_of_the_above.to_s.in?(selection_without_blanks)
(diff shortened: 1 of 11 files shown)


app/models/question/selection.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  private
+
+    def selection_without_blanks
+      selection.reject(&:blank?)
+    end
+
+    def validate_checkbox
+      return errors.add(:selection, :blank) if selection_without_blanks.empty?
+      return errors.add(:selection, :both_none_and_value_selected) if selection_without_blanks.count > 1 && "None of the above".in?(selection_without_blanks)
```
thomasiles, line 27: Is there a nice way to avoid the string "None of the above" being in here and in the view? maybe a symbol which gets replaced by the text? If not it doesn't matter.

Also, I don't know if this matters, there is no validation that the option is part of the options listed in `answer_settings`. The user could submit any string here (maybe changing the radios value attribute through dev tools) and get it in the submission email. I don't know how likely that is, or how much of an issue it would be?

thomasiles (commented):
awesome, like the tests

app/models/question/selection.rb
DavidBiddle: Good suggestions, have updated 👍

thomasiles (approved):
great work, really good work getting it done so quickly

PR #199, opened 2022-12-05
Sentry stop 404 for malicious get requests to from pages

thomasiles:
# Reduce sentry noise by removing PageNotFound error from step controller

Sentry reports errors thrown by the pages controller when URLs for form pages contain invalid page URLs or trailing characters. This happens a lot because of bots looking for well known exploitable URLS.

The sentry issue can be found here:
https://sentry.io/organizations/govuk-forms/issues/3551642298/?project=6576261&query=is%3Aunresolved&referrer=issue-stream

This PR includes a failing test and a fix. 

The fix adds a regex constraint on the route which connects requests to the forms/page_controller. By restricting the page_id to being a string of digits or "check_your_answers" we can stop the controller action being called when given invalid urls. This causes the standard 404 route to match, returning a 404.

Only the get page route has been modified. If we find the other page_id requests have similar issues, we might consider wrapping them with constraints, or adding a guard clause to the base_controller.

Trello card: https://trello.com/c/hDbXwqQ6/326-sentry-stop-logging-404-errors-for-bots-trying-to-hack-forms-runner

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/spec/requests/page_spec.rb
+++ b/spec/requests/page_spec.rb
@@ -125,6 +125,31 @@ RSpec.describe "Page Controller", type: :request do
     end
 
     context "with preview mode off" do
+      [
+        "/form/2/1/__",
+        "/form/2/1/debug.cgi",
+        "/form/2/1/hsqldb%0A",
+        "/form/2/1/index_sso.php",
+        "/form/2/1/setup.php",
+        "/form/2/1/test.cgi",
+        "/form/2/1/x"
+      ].each do |path|
+        context "with an invalid URL: #{path}" do
+          before do
+            allow(Sentry).to receive(:capture_exception)
+            get path
+          end
+
+          it "returns a 404" do
+            expect(response.status).to eq(404)
+          end
+
+          it "does not send an expception to sentry" do
+            expect(Sentry).not_to have_received(:capture_exception)
+          end
+        end
+      end
+
       it "Returns a 200" do
         get form_page_path("form", 2, "form-1", 1)
         expect(response.status).to eq(200)
--- a/spec/requests/page_spec.rb
+++ b/spec/requests/page_spec.rb
@@ -126,6 +126,12 @@ RSpec.describe "Page Controller", type: :request do
 
     context "with preview mode off" do
       [
+        "/form/2/1/check_your_answers_trailing",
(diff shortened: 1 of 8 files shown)


spec/requests/page_spec.rb
```diff
@@ -125,6 +125,31 @@
...
+          before do
+            allow(Sentry).to receive(:capture_exception)
+            get path
+          end
+
+          it "returns a 404" do
+            expect(response.status).to eq(404)
+          end
+
+          it "does not send an expception to sentry" do
```
aliuk2012, line 147:
```suggestion
          it "does not send an exception to sentry" do
```

spec/lib/step_factory.rb
```diff
@@ -0,0 +1,18 @@
+require "rails_helper"
+
+RSpec.describe StepFactory do
+  describe StepFactory::PAGE_SLUG_REGEX do
+
+    it "matches valid page_id values" do
+      %w[1 123 0123456789 check_your_answers ].each do |string|
```
aliuk2012, line 7: 🤦  just spotted that we have a few routes that use underscore rather than hyphens. This approach does make me nervous because any rename to check your answer or other pages we might add would mean that we would have to add them to this safe list.

aliuk2012 (commented):
Very smart, TBH I was thinking we would just add a check for attempts to access `:form_id/:form_slug/...` where both params are integers  would be redirected to 404.

spec/lib/step_factory.rb
thomasiles: I think routes with underscores rather than hyphens is my fault across the codebase 🤦 

My first reaction was that 'check_your_answers' wasn't a string defined in the routes but used by the stepping mechanics. I tried changing it there and got a load of errors, so that doesn't help.

I've pushed a new commit which moves the value to a single constant. I've put it in the check_your_asnwers_step. 

Changing it now works for whatever value you put in it. It's hard to write a spec for that though, because the constant is used in the routes file.

This makes the relationship between the values clear but I'm not sure if using a constant like this in the routes file is reasonable or not.

aliuk2012 (approved):

PR #205, opened 2022-12-12
Add hidden field for notify reference

SamJamCul:
This PR is part of a set, the other being in forms-deploy https://github.com/alphagov/forms-deploy/pull/54

Trello card: https://trello.com/c/R9jqd9GT/341-e2e-add-checking-notify-has-received-and-sent-off-the-completed-submitted-form

The idea is to check that submitting a form correctly triggers a Notify email, which is successfully delivered.

The changes on the runner side are to provide the reference to Notify as part of the `notify_service`, and also to display the reference as a hidden field on the check your answers page, so that the e2e tests can obtain it. 

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/views/forms/check_your_answers/show.html.erb
+++ b/app/views/forms/check_your_answers/show.html.erb
@@ -25,6 +25,7 @@
     <% end %>
 
     <%= form_with url: @form_submit_path do |form| %>
+      <%= form.hidden_field :notify_reference, id: 'notification-id', value: @notify_reference %>
       <%= form.govuk_submit 'Submit' %>
     <% end %>
   </div>
--- a/.gitignore
+++ b/.gitignore
@@ -37,8 +37,8 @@
 
 .env
 .env.development.local
-.env.test.local	
-.env.production.local	
+.env.test.local
+.env.production.local
 
 coverage/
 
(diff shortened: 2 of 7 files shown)


app/views/forms/check_your_answers/show.html.erb
```diff
@@ -25,6 +25,7 @@
     <% end %>
 
     <%= form_with url: @form_submit_path do |form| %>
+      <%= form.hidden_field :notify_reference, id: 'notification-id', value: @notify_reference %>
```
aliuk2012, line 28: Any chance we could test this value is getting through to the controller action when the form is posted?

Maybe a view spec would be better just to make sure this value is on the page. The reason I'm suggesting this is that if someone deletes this line, there may not be any tests in the repo that would fail but when you come to deploy the app forms-deploy would only then flag up the issue. It would be better to catch it sooner.

SamJamCul: Sure thing!

aliuk2012 (approved):

app/views/forms/check_your_answers/show.html.erb
SamJamCul: I've added a view test for the hidden notify ref

PR #224, opened 2023-01-06
Fix bug with existing dates

DavidBiddle:
Visiting a question page with an existing date field (i.e. one without an answer settings field) will result in an error, since the date question type expects the input type to be present.

This PR adds a check for this. This should be a temporary solution - the proper solution is to migrate the existing date (and address) fields. There is a ticket for this in the backlog which should be done in the next sprint.

#### Checklist

- [n/a] I've linked this PR to the relevant issue (if mission work)
  - [n/a] README.md

### DIFF
--- a/app/models/question/date.rb
+++ b/app/models/question/date.rb
@@ -25,7 +25,7 @@ module Question
     end
 
     def date_of_birth?
-      answer_settings&.input_type == "date_of_birth"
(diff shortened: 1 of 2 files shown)


aliuk2012 (commented):
Any chance we could add a test for this?

DavidBiddle:
@aliuk2012 Thanks, have added tests now

aliuk2012 (approved):

PR #233, opened 2023-01-17
Refactor how we use Notify and form submission

aliuk2012:

- Installs config gem for configuring settings (we need this for feature specs and want to move more env vars to settings.yaml)
- Updates Readme about using settings.
- Moves the existing notify template id to the settings file
- replaces Env vars NOTIFY_API_KEY to SETTINGS__GOVUK_NOTIFY__API_KEY
- Install and use govuk_notify_rails gem instead of plain ruby notifications ruby
- refactor notify_service but splitting out the functionality into two smaller distinct objects.|
   - form_submission_service - responsible for taking form data and formatting it for notify. This service will call the new form_submission_mailer, (or potentially any other method we might want to add in future send via api/download & print options etc)
    - form_submission_mailer - `actionmailer` object that is responsibly purely for sending the final submission email to form processing team.

- remove old notification ruby client and notify_service.rb

Doing this refactor means we have more control over form submissions in future and its a lot easy to test the functionality in isolation. 

Trello card: https://trello.com/c/hNv5h7Ct/443-add-no-reply-email-address

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


.gitignore
```diff
@@ -44,3 +44,12 @@ coverage/
 
 # Vim
 *.swp
+
+# Ignore Cypress output
+/cypress/screenshots
+/cypress/videos
```
DavidBiddle, line 50: Don't think this Cypress stuff should be in here:
```suggestion
```

README.md
```diff
@@ -70,6 +88,19 @@ yarn dev
 
 For now, to test the API integration, you will also need to run the [API service](https://github.com/alphagov/forms-api).
 
+## Explain how to use GOV.UK Notify
+
+If you want to test the notify function, you will need to get a test API key
+from the [notify service](https://www.notifications.service.gov.uk/) Add it as
+an environment variable under `SETTINGS__GOVUK_NOTIFY__API_KEY=` or creating/edit
```
DavidBiddle, line 95: Tiny spelling thing:
```suggestion
an environment variable under `SETTINGS__GOVUK_NOTIFY__API_KEY=` or create/edit
```

DavidBiddle (approved):
Couple of tiny comments but looking good otherwise! And makes total sense to do it this way

.gitignore
aliuk2012: Good spot, not sure how that got added

PR #250, opened 2023-02-01
Fix undefined method `future?'  for date answer types

aliuk2012:
This method is a rails method that was being called on a Struct rather than a Date object

fixes https://sentry.io/organizations/govuk-forms/issues/3907609448/?project=6576261&query=is%3Aunresolved&referrer=issue-stream

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/models/question/date.rb
+++ b/app/models/question/date.rb
@@ -34,8 +34,8 @@ module Question
       return if blank? && is_optional?
       return errors.add(:date, :blank) if blank?
       return errors.add(:date, :blank_date_fields, fields: blank_fields.to_sentence) if present? && blank_fields.any?
-      return errors.add(:date, :future_date) if date_of_birth? && future_date?
       return errors.add(:date, :invalid_date) if invalid?
+      return errors.add(:date, :future_date) if date_of_birth? && future_date?
     end
 
     def date_field_to_attribute(key)
--- a/spec/models/question/date_spec.rb
+++ b/spec/models/question/date_spec.rb
@@ -54,7 +54,7 @@ RSpec.describe Question::Date, type: :model do
     end
   end
 
-  context "when a date has a valid day, month or year but it's not a real date" do
+  context "when a date has a day, month or year but it's not a real date" do
     before do
       set_date("45", "02", "2021")
     end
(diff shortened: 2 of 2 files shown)


app/models/question/date.rb
```diff
@@ -34,8 +34,8 @@ def date_valid
       return if blank? && is_optional?
       return errors.add(:date, :blank) if blank?
       return errors.add(:date, :blank_date_fields, fields: blank_fields.to_sentence) if present? && blank_fields.any?
-      return errors.add(:date, :future_date) if date_of_birth? && future_date?
       return errors.add(:date, :invalid_date) if invalid?
+      return errors.add(:date, :future_date) if date_of_birth? && future_date?
```
danworth, line 38: Ah ok so the root cause was that the validation was checking to see if it was a future date before first checking if it was a valid date (hence it attempted to call date.future when it wasn't a real date).

spec/models/question/date_spec.rb
```diff
@@ -79,6 +79,26 @@
...
+        day_month_years = [
+          [1, "e", 2022],
+          ["pi", 1, 2022],
+          [1, 1, "zero"],
+        ]
+        question.answer_settings = { input_type: "date_of_birth" }
+
+        day_month_years.each do |v|
+          set_date(*v)
+          question.valid?
```
danworth, line 96: What's the significance that here we call `question.valid?` but in the very similar test above we don't make the same call to `valid?` between lines 77 and 78?

danworth (approved):
Looks good, this all makes much more sense after a morning of doing the Rails tutorial :) Just a question that I'm interested to know the answer to in the new test please 👍

app/models/question/date.rb
aliuk2012: 💯

aliuk2012: `future_date?` could potentially have a guard to only call `.future` if date is a valid date.

spec/models/question/date_spec.rb
aliuk2012: oh.....thats a good catch. 

I was explaining `.valid?` to James. This method calls the validations on an object and returns true/false. At the same time it also add error messages to the object which can used to determine which validations failed.

A shorthand and slightly easier way to write tests that check it is to use rspec `be_valid?` 
```
 expect(question).not_to be_valid
          expect(question.errors[:date]).to include(I18n.t("activemodel.errors.models.question/date.attributes.date.invalid_date"))
```

See b579b3d , No change in the outcome. It just meant we were calling `.valid?` twice and because the object is invalid it doesn't really affect the tests because it was always invalid.

PR #260, opened 2023-02-13
Remove legacy text answer types

DavidBiddle:
Now that we've removed the legacy `single_line` and `long_text` answer types from forms-admin and forms-api, and migrated the existing items in the database to use the `text` type, we can remove any code relating to the legacy types from this repository too.

Trello card: https://trello.com/c/CFCx7yQB/522-remove-autofill-feature-flags

#### Checklist

  - [n/a] README.md


aliuk2012:
Looks really good and love the fact we are removing so much code. Do you think it might be worth adding FactoryBot to forms-runner like we do in forms-admin even if its just to add Page models? It might help some of those tests be a bit more readable without them having to be so verbose.

SamJamCul:
> Looks really good and love the fact we are removing so much code. Do you think it might be worth adding FactoryBot to forms-runner like we do in forms-admin even if its just to add Page models? It might help some of those tests be a bit more readable without them having to be so verbose.

Definitely agree with the factory idea, I've started working on this on the back of this PR...

SamJamCul (approved):

PR #269, opened 2023-02-27
simple catch of preview submit, probably needs more sentry context?

0atman:

Preview submissions send emails, they should not!

Related to https://trello.com/c/8v62dCzt/551-users-can-use-the-back-button-to-submit-a-form-twice-resulting-in-blank-emails-in-processing-inboxes

### DIFF
--- a/app/controllers/forms/submit_answers_controller.rb
+++ b/app/controllers/forms/submit_answers_controller.rb
@@ -4,10 +4,15 @@ module Forms
       unless preview?
         EventLogger.log_form_event(current_context, request, "submission")
       end
-
-      FormSubmissionService.call(form: current_context,
+      if form.email.nil?
+        Sentry.capture_message("preview submitted, caught")
+        render "errors/submission_error", status: :internal_server_error
+      else
+        FormSubmissionService.call(form: current_context,
                                  reference: params[:notify_reference],
                                  preview_mode: preview?).submit_form_to_processing_team
+
+      end
 
       current_context.clear
       redirect_to :form_submitted


DavidBiddle:
> Preview submissions send emails, they should not!

Is this correct? I think we want preview to submit so that form creators can check the content is coming through correctly in the email. Could be wrong but I think the problematic behaviour is that it still tries to submit in preview even if there's no submission email set.

0atman:
> > Preview submissions send emails, they should not!
> 
> Is this correct? I think we want preview to submit so that form creators can check the content is coming through correctly in the email. Could be wrong but I think the problematic behaviour is that it still tries to submit in preview even if there's no submission email set.

My mistake! We're actually going to not send an email when there is no submission_email AND it's in preview

0atman:
image

I've tied myself in a knot so far - committing what I have so far, will comment in-line an explaination

app/lib/context.rb
```diff
@@ -68,4 +68,20 @@ def form
...
+
+  def submit_users_answers(notify_reference, preview_mode, request)
+    unless preview_mode
+      EventLogger.log_form_event(self, request, "submission")
+    end
+
+    FormSubmissionService.call(form: self,
+                               reference: notify_reference,
+                               preview_mode:,
+                               request:).submit_form_to_processing_team
```
0atman, line 80: So to get logging inside the `FormSubmissionService` I'm passing it in here

app/services/form_submission_service.rb
```diff
@@ -5,21 +5,30 @@ def call(**args)
     end
   end
 
-  def initialize(form:, reference:, preview_mode:)
+  def initialize(form:, reference:, preview_mode:, request:)
```
0atman, line 8: Tried to modfiy the interface of the service here, adding the request to be passed in

app/services/form_submission_service.rb
```diff
@@ -5,21 +5,30 @@ def call(**args)
...
     timestamp = submission_timestamp
+    if @form.submission_email.present?
+      FormSubmissionMailer
+        .email_completed_form(title: form_title,
+                              text_input: email_body,
+                              reference: @reference,
+                              timestamp:,
+                              submission_email: @form.submission_email).deliver_now
+    else
+      EventLogger.log_form_event(@form, request, "no email submission")
```
0atman, line 25: so it can be logged here

spec/services/form_submission_service_spec.rb
```diff
@@ -1,8 +1,9 @@
 require "rails_helper"
 
 RSpec.describe FormSubmissionService do
-  let(:service) { described_class.call(form:, reference: "for-my-ref", preview_mode:) }
-  let(:form) { OpenStruct.new(form_name: "Form 1", submission_email: "testing@gov.uk", steps: [step]) }
+  let(:service) { described_class.call(form:, reference: "for-my-ref", preview_mode: request: {}) }
```
0atman, line 4: Adding request here I think is right, but causes ruby to error with "/shrug" and no useful info

0atman:
Sorry I couldn't figure this out @aliuk2012 , I'll be back Tuesday 7th to continue this.
But I think probably you should fix it for now, I'll pick up something else when I get back.

Thank you for your explanation and help on this, I wouldn't have even got so far without it!

aliuk2012:
https://github.com/alphagov/forms-runner/pull/278 replaces this PR

PR #274, opened 2023-03-08
Improve our session management error handling

aliuk2012:

Currently, we do not inform a user why there is an error when their session has expired.  I added a timestamp to the session which is set the first time a user saves their answer. This datestamp is checked on every answer save, check your answers and submit answers controller actions and redirects them to the new session expired page. The condition is that the user needs to complete the form within 20hrs.

Trello card: https://trello.com/c/8v62dCzt/551-users-can-use-the-back-button-to-submit-a-form-twice-resulting-in-blank-emails-in-processing-inboxes

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


aliuk2012:
@thomasiles and I had a chat about this and thing its better to park this PR until everyone is back in. There is a lot of complexity that we need to think about and it doesn't really fix the issue of users using the back button.  

For example: How do sessions work across multiple forms? Should answers for multiple forms be stored as separate sessions or should it all be as one single session.

Currently multiple forms are stored in the same session, the expiry time of 20hrs is reset every time a user refreshes their page (i.e showing some sort of activity).

thomasiles (approved):

jamie-o-wilkinson:
Closing as it has been a while and this has gotten a bit stale. Feel free to reopen if need be.

PR #277, opened 2023-03-09
Fix: Form process teams are receiving forms with no questions/answers

aliuk2012:

We noticed an issue where users would do the following:

- submit their form answers,
- get a confirmation success page,
- use their browsers "back button" to return to "Check your answers" page
- attempte to submit their answers again

This resulted in submission emails being sent to form processing
teams with no questions or answers.

This commit, moves when a session is cleared until the user lands
on the confirmation page. when it clears the answers for the form the
session changes from

```
{ 
  answers: {
    "4": {
      "1": "This is the answer for the first question of form 4"
    }
  }
}
```

```
{ 
  answers: {
    "4": NULL
  }
}
```
This results in the form submission email having no questions or answers.

Trello card:

https://user-images.githubusercontent.com/3441519/224008109-e81ba0b6-92dd-4f42-837a-eea523adc507.mov

#### Checklist

  - [x] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/controllers/forms/submit_answers_controller.rb
+++ b/app/controllers/forms/submit_answers_controller.rb
@@ -1,16 +1,18 @@
 module Forms
   class SubmitAnswersController < BaseController
     def submit_answers
-      unless preview?
-        EventLogger.log_form_event(current_context, request, "submission")
+      if current_context.form_submitted?
+        redirect_to error_repeat_submission_path(current_form.id)
+      else
+        unless preview?
+          EventLogger.log_form_event(current_context, request, "submission")
+        end
+
+        FormSubmissionService.call(form: current_context,
+                                   reference: params[:notify_reference],
+                                   preview_mode: preview?).submit_form_to_processing_team
+        redirect_to :form_submitted
       end
-
-      FormSubmissionService.call(form: current_context,
-                                 reference: params[:notify_reference],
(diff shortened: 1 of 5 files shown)


config/locales/errors.yml
```diff
@@ -8,6 +8,11 @@ en:
       If you typed the web address, check it is correct.
       If you pasted the web address, check you copied the entire address.
     title: Page not found
+  repeat_submission:
+    body: We deleted your answers because they've been submitted successfully already.
```
hannahkc, line 12: Not essential (if you ask me), but would be nice to make this straight apostrophe a curly one. ie: they’ve

SamJamCul (approved):
Looks good to me, just Hannah's comment about the apostrophe!

config/locales/errors.yml
SamJamCul: How do you type that!

DavidBiddle: `Alt + ]` -> `‘`
`Alt + Shift + ]` -> `’`
`Alt + [` -> `“`
`Alt + Shift + {` -> `”` 

```suggestion
    body: We deleted your answers because they’ve been submitted successfully already.
```

hannahkc: (or do what I do and copy it from someplace...! ) Thanks both!

PR #278, opened 2023-03-09
Ensure a form has a submission email and steps before sending submission email

aliuk2012:

We need to validate that when a form filler submits their
answer for a form, before attempting to call GOVUK Notify,
we check that the form actually has a submission email.

This should be very rare because live forms should have
a submission email before they are made live. The only real
use case where a submission email maybe blank is when a
the form creator is previewing a form and they haven't completed
the add submission email task before they previewed the form. Rather than attempting to call GOVUK Notify without
a submission email and for it to raise an unhandled error we
can instead ignore the call to the mailer and process
the request as though the submission email address was provided
(After all, this is just a preview form so it shouldn't really
matter but if the form creator has provided the email then we should continue to send the answers to that email address)

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/services/form_submission_service.rb
+++ b/app/services/form_submission_service.rb
@@ -12,6 +12,8 @@ class FormSubmissionService
   end
 
   def submit_form_to_processing_team
+    raise StandardError, "Form id(#{@form.id}) has no steps i.e questions/answers to include in submission email" if @form.steps.blank?
+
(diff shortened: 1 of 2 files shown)


thomasiles (approved):
👍 looks good. i got a bit confused by the email sending - we only send an email if the submission_email is set and we are not in @preview, right?

PR #294, opened 2023-04-03
Change preview headers

thomasiles:

Trello card: https://trello.com/c/CYcJlAqv/620-add-a-way-to-show-users-whether-theyre-looking-at-a-live-form-a-preview-of-a-live-form-or-a-preview-of-a-draft-form-in-the-runne

This adds the following to the preview-draft/preview-live views:
- altered page titles
- different styling - (colours and header)
- visually hidden text at the start of questions to tell users they are in preview

To do this:
- I changed how we detect which mode were are in, removing the controller methods which use the mode passed in from the url, adding a class with methods
- added components for the headers
- consolidated logic for titles and question_text into application helpers and updated where we called them

Everything has ended up being a bit messy - I should have split this into smaller, easier to merge commits.

If anything doesn't make sense or should be done differently, let me know!

Draft preview:
image

Live preview:
image

Live view:
image

The privacy page also contains form content so gets the preview/draft treatment:
image

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


app/helpers/application_helper.rb
```diff
@@ -10,8 +10,15 @@ def set_page_title(title)
     content_for(:title) { title }
   end
 
-  def title_with_error_prefix(title, error)
-    "#{t('page_titles.error_prefix') if error}#{title}"
+  def form_title(page_name:, mode:, error: false)
+    mode_string = if mode&.preview_draft?
+                    " - #{t('mode.title_text_preview-draft')}"
+                  elsif mode&.preview_live?
+                    " - #{t('mode.title_text_preview-live')}"
```
aliuk2012, line 17: do we need Safe Navigation Operator?

app/components/form_header_component/view.rb
aliuk2012, line 1: Do we have any unit tests/preview for this components?

app/components/preview_component/view.rb
aliuk2012, line 1: Great to see previews for this component but could do with some unit tests.

config/locales/en.yml
```diff
@@ -78,17 +78,24 @@ en:
...
     privacy_policy: Privacy
   form:
+    check_your_answers:
+      title: Check your answers before submitting your form
     submitted:
       title: Your form has been submitted
+  mode:
+    phase_banner_tag_preview-draft: Draft preview
+    phase_banner_tag_preview-live: Live preview
+    phase_banner_text_preview-draft: You're previewing the draft version of this form
```
hannahkc, line 88:
```suggestion
    phase_banner_text_preview-draft: You’re previewing the draft version of this form
```

config/locales/en.yml
```diff
@@ -78,17 +78,24 @@ en:
...
+    phase_banner_text_preview-live: You're previewing the live version of this form
```
hannahkc, line 89:
```suggestion
    phase_banner_text_preview-live: You’re previewing the live version of this form
```

deberny:
I'm gonna try and tweak the size for the 'background-image' so it's less busy and will let you know what works 😄

app/helpers/application_helper.rb
thomasiles: They make things slightly more robust and easier to test - it saves mocking all methods of mode for a live form. I think a better approach would be to use Factorybot but this PR is already quite big. Might be worth it though?

app/components/form_header_component/view.rb
thomasiles: I've added both 👍

thomasiles:
> I'm gonna try and tweak the size for the 'background-image' so it's less busy and will let you know what works 😄

Happy to help if it's easier to do it in a call 👍

spec/components/preview_component/view_spec.rb
```diff
@@ -0,0 +1,25 @@
+require "rails_helper"
+
+RSpec.describe PreviewComponent::View, type: :component do
+  it "shows in preview_draft" do
+    mode = Mode.new('preview-draft')
+    render_inline(described_class.new(mode: mode))
+    expect(page).to have_selector(".govuk-phase-banner")
+    expect(page).to have_content("Draft preview")
+    expect(page).to have_content("You're previewing the draft version of this form")
```
DavidBiddle, line 9: Change to curly quotes is causing the test failure here:

```suggestion
    expect(page).to have_content("Draft preview")
    expect(page).to have_content("You’re previewing the draft version of this form")
```
or call the translation directly:
```suggestion
    expect(page).to have_content(I18n.t("mode.phase_banner_tag_preview-draft"))
    expect(page).to have_content(I18n.t("mode.phase_banner_text_preview-draft"))
```

spec/components/preview_component/view_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  it "shows in preview_live" do
+    mode = Mode.new('preview-live')
+    render_inline(described_class.new(mode: mode))
+    expect(page).to have_selector(".govuk-phase-banner")
+    expect(page).to have_content("Live preview")
+    expect(page).to have_content("You're previewing the live version of this form")
```
DavidBiddle, line 17:
```suggestion
    expect(page).to have_content("Live preview")
    expect(page).to have_content("You’re previewing the live version of this form")
```
or
```suggestion
    expect(page).to have_content(I18n.t("mode.phase_banner_tag_preview-live"))
    expect(page).to have_content(I18n.t("mode.phase_banner_text_preview-live"))
```

app/helpers/application_helper.rb
aliuk2012: We do need to add Factorybot at some point, this is pretty bit PR already so maybe best to leave it for time being. I'd rather mock things out in specs than end up with things than fail gracefully and not trigger alerts for us to investigate weird behaviour. Let me know if I can help with any of the specs and their mocks

spec/components/preview_component/view_spec.rb
aliuk2012: I think translations in tests are better

app/helpers/application_helper.rb
thomasiles: I've removed the & and added the mocks needed in the tests 👍

aliuk2012 (approved):
# 🏅 🎖️ 🏅

PR #305, opened 2023-04-21
Refactor specs

aliuk2012:
There were a couple of issues with our specs in this repo

- spec filenames didn't match the filenames for the code they were testing
- specs main rspec descriptions where using strings rather than the class name that they were testing
- specs folder failed a number of rubocop errors
- rubocop was only checking files in the `app` folder as part of CI. This meant that when locally developing, we had to remember to only scan the app folder otherwise we would end up having to fix failed linting not related to our own code.

Trello card: https://trello.com/c/RKCviWz0/683-change-rubocop-settings-for-the-runner-to-include-specs-dir

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/.github/workflows/rubyonrails.yml
+++ b/.github/workflows/rubyonrails.yml
@@ -47,6 +47,6 @@ jobs:
       - name: Security audit application code
         run: bundler exec brakeman -q -w2
       - name: Lint Ruby files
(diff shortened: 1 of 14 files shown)


thomasiles (approved):
🥇 so nice to not have to run `rubocop app` now

PR #311, opened 2023-05-04
Align Github Actions Across Our App Repos

danworth:
1. Use `snake_case` for file naming
2. Remove the setup of Postgres which is not necessary for Forms-Runner
3. Add a note explaining that the ruby action looks at the .ruby-version to install the correct version.

Improves consistency between our apps which makes maintaining simpler. Documentation for the `ruby/setup-ruby` action: https://github.com/ruby/setup-ruby#single-job

### Notes ###
forms-runner and forms-admin are most similar and look to do things well. A few minor amendments will align these two. The equivalent action on forms-api needs a few more changes which will follow shortly.


lfdebrux:
Can we rename the `rubyonrails.yml` workflow to something like `test.yml`? It took me a few minutes just now to figure out which workflow actually had the tests in 😅

lfdebrux (approved):
:shipit:

PR #320, opened 2023-05-16
Back to edit form link

aliuk2012:

Adds a "Edit this question" link to the phase banner if the user is previewing a draft form.

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


app/components/preview_component/view.rb
```diff
@@ -1,30 +1,32 @@
...
-                         },
-                         text: t("mode.phase_banner_text_#{@mode}"))
+      super
     end
 
     def render?
       @mode.preview?
     end
 
-  private
```
DavidBiddle, line 20: Not sure it makes much difference, but was removing this intentional? it doesn't look like either of the methods underneath it are used outside the component

app/controllers/forms/page_controller.rb
```diff
@@ -6,6 +6,7 @@ def show
       redirect_to form_page_path(@step.form_id, @step.form_slug, current_context.next_page_slug) unless current_context.can_visit?(@step.page_slug)
       back_link(@step.page_slug)
       @is_question = true
+      @question_edit_link = "#{Settings.forms_admin.base_url}/forms/#{@step.form_id}/pages/#{@step.page_slug}/edit"
```
DavidBiddle, line 9: Feels a bit fragile to do this on the runner side - if we change the admin URL the link will break - but I don't think there's a better way to do it. 
Might be worth adding navigation using these links to the end-to-end test journey, so that we'll know if the links get out of sync.

app/controllers/forms/page_controller.rb
```diff
@@ -19,6 +20,7 @@ def save
         redirect_to next_page
       else
         @is_question = true
+        @question_edit_link = "#{Settings.forms_admin.base_url}/forms/#{@step.form_id}/pages/#{@step.page_slug}/edit"
```
DavidBiddle, line 23: Identical to line 9 above - might be worth DRYing this up

DavidBiddle (approved):
Looks good overall. Made a suggestion on something you might be able to DRY up, and a suggestion about end-to-end testing (which will have to be a separate PR anyway).

app/components/preview_component/view.rb
aliuk2012: Good spot 👍

PR #322, opened 2023-05-18
Refactor Context form method uses form object and not just id

aliuk2012:

Context.form was extremely confusing because when you looked at `current_context.form`, one may assume that it would return the form that the user is filling in. Instead it would just return the form id. Not only was this confusing but also it meant that for other form attributes we ended up created attributes in the context to play back the same attributes we had in the form already. There will be more refactors but I wanted to try keep them small and targetted.

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/lib/context.rb
+++ b/app/lib/context.rb
@@ -1,14 +1,14 @@
 class Context
-  attr_accessor :form_slug
+  attr_accessor :form_slug, :form
   attr_reader :form_name, :form_start_page, :privacy_policy_url, :what_happens_next_text, :support_details, :declaration_text
 
   def initialize(form:, store:)
+    @form = form
     @form_context = FormContext.new(store)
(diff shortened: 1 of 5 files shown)


app/lib/context.rb
```diff
@@ -58,15 +58,11 @@ def steps
   end
 
   def clear
-    @form_context.clear(@form_id)
+    @form_context.clear(@form.id)
```
DavidBiddle, line 61: Does the `attr_accessor` you've added mean you can use:
```suggestion
    @form_context.clear(form.id)
```

SamJamCul (approved):
Looks like a healthy change - do we need the first commit of these two though? I think all the changes in there are overridden in the second commit

SamJamCul (approved):

PR #328, opened 2023-05-22
Add factorybot and faker

aliuk2012:

We need factorybot & faker for testing and have been putting it off for a long period of time. This PR is to get factorybot & faker in and replaces a few tests to show the factories/faker is working but we should refactor as we work on new features.

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/spec/factories/models/forms.rb
+++ b/spec/factories/models/forms.rb
@@ -1,0 +1,67 @@
+FactoryBot.define do
+  factory :form, class: "Form" do
+    sequence(:name) { |n| "Form #{n}" }
+    sequence(:form_slug) { |n| "form-#{n}" }
+    has_draft_version { true }
+    has_live_version { false }
+    submission_email { Faker::Internet.email(domain: "example.gov.uk") }
+    privacy_policy_url { Faker::Internet.url(host: "gov.uk") }
+    org { "test-org" }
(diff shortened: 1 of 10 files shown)


spec/factories/models/forms.rb
```diff
@@ -0,0 +1,67 @@
+FactoryBot.define do
```
DavidBiddle, line 1: Not a comment for this PR but wonder if it's a) possible and b) desirable to share these factories between the repos

DavidBiddle: May also just be one for the microservices v monolith debate

DavidBiddle (dismissed):
🎉

spec/factories/models/forms.rb
aliuk2012: It is possible to share these factories in a gem, I do think that factories between forms-admin & forms-runner might be the same but forms-api could be different (ActiveRecord vs ActiveResource). Will take a deeper look when i get a moment.

DavidBiddle (approved):

cadmiumcat (dismissed):

cadmiumcat (approved):

DavidBiddle (approved):

PR #339, opened 2023-06-01
Fix missing routing conditions method

aliuk2012:

Suggest reviewing each commit made. 

- Simplify the params that we pass into Steps model
- Commit to show how to replicate the error we saw in AWS where existing live forms with missing routing_conditions raised an error
- Fix the error by handling the error within the Step model. If pages has routing_conditions method use it otherwise return an empty array which is what newer pages with not routing_conditions will return.

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


DavidBiddle (approved):

jamie-o-wilkinson (approved):
Looks good to me. I like the simplification that passing the page wholesale into the step gives 😄

PR #343, opened 2023-06-09
Add draft/live to form submission email personalisation

lfdebrux:

Trello card: https://trello.com/c/kIjR4WBY

We want to vary the content of the form submission email depending on whether the form was submitted from preview mode (draft) or not. We're using the Notify [optional content][1] feature of their template language.

Unfortunately Notify doesn’t have a way of inverting their conditionals, so with this method we have to send both two booleans values for both cases. We have to be careful that `live` and `draft` never have the same value for the same request, or we will end up with the content for both draft and live submissions in the same email.

[1]: https://www.notifications.service.gov.uk/using-notify/optional-content


app/mailers/form_submission_mailer.rb
```diff
@@ -1,12 +1,17 @@
...
     set_personalisation(
       title:,
       text_input:,
       submission_time: timestamp.strftime("%H:%M:%S"),
       submission_date: timestamp.strftime("%-d %B %Y"),
+      # GOV.UK Notify's templates have conditionals, but only positive
+      # conditionals, so to simulate negative conditionals we add two boolean
+      # flags; but they must always have opposite values!
+      test: preview_mode ? "yes" : "no",
+      not_test: preview_mode ? "no" : "yes",
```
DavidBiddle, line 14: To make sure that they always have opposite values, you could make a method 

```
def make_boolean_readable(boolean)
  boolean ? "yes" : "no"
end
```

and then do 
```suggestion
      test: make_boolean_readable(preview_mode),
      not_test: make_boolean_readable(!preview_mode),
```

aliuk2012 (approved):

PR #346, opened 2023-06-19
Add shell command trace to container run command

lfdebrux:
It is useful when looking at startup logs to be able to know what command was run, to understand the context of the log outputs.

See https://github.com/alphagov/forms-admin/pull/498 for more details.


jamie-o-wilkinson (approved):
Looks good to me 👍

PR #351, opened 2023-06-23
Let specific IP addresses bypass the maintenance page

aliuk2012:
We need some way of bypassing the maintenance page if the users IP is in a specific list of IPs.
These changes allow us to let the user bypass the maintenance page if their IP matches any single IP
address or is included within a specific CIDR notation range of IP

Trello card: trello.com/c/P9CEz21c/824-test-and-update-holding-page

Related: https://github.com/alphagov/forms-admin/pull/499

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


app/controllers/application_controller.rb
```diff
@@ -9,8 +9,16 @@ def accessibility_statement; end
 
   def cookies; end
 
+  # Because determining the clients real IP is hard, simply return the first
+  # value of the x-forwarded_for, checking it's an IP. This will probably be
+  # enough for out basic monitoring in PaaS
```
jamie-o-wilkinson, line 14: Typo? Should out be our?

jamie-o-wilkinson (dismissed):
Looks good to me 👏

app/controllers/application_controller.rb
aliuk2012: fixing now and the source where this was copied from https://github.com/alphagov/forms-admin/pull/510

cadmiumcat (approved):

PR #354, opened 2023-06-26
Use new FORMS_ENV variable in HostingEnvironment

DavidBiddle:
We have a class `HostingEnvironment` which allows the application to know which environment it's running in. This uses the PAAS_ENVIRONMENT variable. However, this value is not available in AWS environments. This means it will always return `unknown-environment` in AWS environments.

We use this class to determine:
- which tag should show in the header (so at present it will show a blue 'Unknown environment' tag in the header)
- whether ViewComponent should render previews in the environment (at present they won't render in the AWS dev environment)
- 
We've now added a FORMS_ENV variable to pass this information into both PaaS and AWS app instances, so this PR updates the appicaltion to use these values.

Trello card: Connected to https://trello.com/c/DFOs5YuI/858-display-which-environment-a-user-is-accessing

#### Checklist

  - [n/a] README.md


jamie-o-wilkinson (approved):
Looks good to me 🎉

config/settings.yml
```diff
@@ -25,3 +25,5 @@ maintenance_mode:
   enabled: false
   # List of IP addresses which will bypass the maintenance mode message as a comma seperated string, optionally using CIDR notation eq. "127.0.0.1, 192.192.192.192/32, 0.0.0.0/24"
   bypass_ips:
+
+forms_env: local
```
aliuk2012, line 29: Could you add a config spec for this new setting?

DavidBiddle: Done!

aliuk2012 (approved):

PR #369, opened 2023-07-14
Convert text question partial to view component

aliuk2012:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/components/question/base.rb
+++ b/app/components/question/base.rb
@@ -1,0 +1,18 @@
+module Question
+  class Base < ViewComponent::Base
+    attr_accessor :form_builder, :question, :extra_question_text_suffix
+
+    def initialize(form_builder:, question:, extra_question_text_suffix:)
+      @form_builder = form_builder
+      @question = question
+      @extra_question_text_suffix = extra_question_text_suffix
+      super
+    end
+
+    def question_text_with_extra_suffix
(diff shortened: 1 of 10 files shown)


app/components/question/base.rb
```diff
@@ -0,0 +1,18 @@
...
+      @form_builder = form_builder
+      @question = question
+      @extra_question_text_suffix = extra_question_text_suffix
+      super
+    end
+
+    def question_text_with_extra_suffix
+      return question.question_text if extra_question_text_suffix.blank?
+
+      [question.question_text, extra_question_text_suffix.html_safe].compact.join(" ").html_safe
```
thomasiles, line 15: can this just be
```ruby
def question_text_with_extra_suffix
  [question.question_text, extra_question_text_suffix.html_safe].compact_blank.join(" ").html_safe
end
```

also do we need the inner .html_safe?

thomasiles (dismissed):
looks really nice!

app/components/question/base.rb
aliuk2012: Ah good spot, also great recommendation of `compact_blank` I had completely forgot about that method. 

I've also refactored how the existing application helper merged the question and mode suffix.

thomasiles (approved):
👍

PR #370, opened 2023-07-14
Convert number question partial to view component

aliuk2012:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/spec/components/question/number_component/view_spec.rb
+++ b/spec/components/question/number_component/view_spec.rb
@@ -1,0 +1,60 @@
+require "rails_helper"
+
+RSpec.describe Question::NumberComponent::View, type: :component do
+  let(:question_page) { build :page, answer_type: "number" }
+  let(:answer_text) { nil }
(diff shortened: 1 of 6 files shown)


spec/components/question/number_component/view_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    end
+
+    context "with unsafe question text" do
+      let(:question_page) { build :page, answer_type: "number", question_text: "What is your name? <script>alert(\"Hi\")</script>" }
+      let(:extra_question_text_suffix) { "<span>Some trusted html</span>" }
+
+      it "returns the escaped title with the optional suffix" do
+        expected_output = "What is your name? &lt;script&gt;alert(\"Hi\")&lt;/script&gt; <span>Some trusted html</span>"
+        expect(page.find("h1 .govuk-label").native.inner_html).to eq(expected_output)
+      end
```
DavidBiddle, line 57: Nice!

DavidBiddle (approved):

PR #371, opened 2023-07-17
Structured logging

danworth:
Changes the log format to be entirely JSON format. This aids searching and reporting in Splunk since it is able to understand the structured log format and automatically extract each key for use. It also formats the log event with syntax highlighting making it simpler to understand.

Before the change:
image

After the change in dev:
image

This also permits simpler splunk queries that make direct reference to any key without needing to first `rex` out content from the log event. E.g. 
```
index="*forms*" source="*forms-runner-dev*" event=form_submission
| stats count by form
```

### Notes on commits ###
- The first commit refactors the specs so that the request specs expect on the higher level logging functions without making direct reference to the output format which is an implementation detail from their concern.
-  The second commit introduces the JSON logging format.
- The third commit adds the exception field to the Lograge event if there is one, useful for tracing exceptions with user activity.

### Other things to Note ###
- Ideally I'd rather not rely on `rescue` from an exception in conditional logic but the `JsonLogFormatter#is_json` seems to be the idiomatic way of doing this from the various Rails solutions I've found.
- I have updated the README.md with details about the logging which should be read and understood as part of this review.

- The Lograge and common logging configuration has been moved into `config/application.rb` since we favour keeping environments common, it also makes it simpler to develop on the log format without needing to make changes (previously Lograge wasn't enabled for non-production environments). It would be possible to toggle between JSON and positional logging formats for local dev if there's preference based upon ENV vars. If that's something people want then it can be added in a later PR.
- **Limitation** the initial start up logs from Puma are not in JSON format since they do not seemingly use the logger configured in the application config. e.g. 
```
Puma starting in single mode...
```
- From some experimentation exceptions are logged thus, seen in development environment. This seems a benefit since the exception message appears in a single event under the `message` key with `level: FATAL` making them easier to find. Note there is a ticket to make this happen: https://trello.com/c/ZGYPzwJP/80-prevent-splunk-from-splitting-logs
```
{ [-]
   level: FATAL
   message:   
ActionView::Template::Error (undefined method `name' for 1:Integer):
    1: <% if @render_args[:component] %>
    2:   <% if ViewComponent::Base.config.render_monkey_patch_enabled || Rails.version.to_f >= 6.1 %>
    3:     <%= render(@render_args[:component], @render_args[:args], &@render_args[:block]) %>
    4:   <% else %>
    5:     <%= render_component(@render_args[:component], &@render_args[:block]) %>
    6:   <% end %>
  
app/components/form_header_component/view.rb:13:in `call'
   time: 2023-07-14T15:33:13.834+00:00
}
```
- This change does not appear to change the Sentry behaviour, the corresponding Sentry event for the exception log above can be viewed here: https://govuk-forms.sentry.io/issues/4315463747/?project=6576261&query=is%3Aunresolved&referrer=issue-stream&stream_index=0
- Lograge is configured to include the exception message if one occurs. This message includes the usual lograge fields and makes it simple to link the exception to the client activity via the request_id
```
{ [-]
   action: save
   controller: Forms::PageController
   duration: 0.59
   exception: [ [-]
     ActionController::InvalidAuthenticityToken
     Can't verify CSRF token authenticity.
   ]
   form_id: 119
   format: html
   host: submit.dev.forms.service.gov.uk
   level: INFO
   method: POST
   path: /preview-draft/119/a-test-form/182
   request_id: 58b528ed-779f-4011-830a-0348ead25137
   status: 422
   time: 2023-07-17T15:07:45.635+00:00
   view: 0
}
```

### How to Test ###
- You can run this locally and see the logs in JSON format.
- **This is currently deployed in dev environment.** You can search for [runner logs in dev in Splunk](https://gds.splunkcloud.com/en-GB/app/gds-543-forms/search?q=search%20index%3D%22*forms*%22%20source%3D%22*forms-runner-dev*%22&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=%40d&latest=now&display.page.search.tab=events&display.general.type=events&sid=1689590083.560392)

### Implications on existing Reporting ###
Before I merge I'll speak with @AnneCarr to make sure any reports are updated accordingly.

#### Checklist
Trello card: https://trello.com/c/VSdLnRsI/130-firebreak-convert-logs-to-json-structured

  - [x] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,14 @@ bin/rake run_code_quality_checks
 
 Use the [update_app_versions.sh script in forms-deploy](https://github.com/alphagov/forms-deploy/blob/main/support/update_app_versions.sh)
 
+## Logging
+
+- HTTP access logs are managed using [Lograge](https://github.com/roidrage/lograge) and configured within [the application config](./config/application.rb)
+- Custom app logging should be made via the [LogEventService](./app/services/log_event_service.rb) and [EventLogger](./app/lib/event_logger.rb). This is
+independent to any Lograge configuration.
+- The output format is JSON using the [JsonLogFormatter](./app/lib/json_log_formatter.rb) to enable simpler searching and visbility especially in Splunk.
+- **DO NOT** use [log_tags](https://guides.rubyonrails.org/configuring.html#config-log-tags) since it breaks the JSON formatting produced by Lograge.
+
 ## Support
 
 Raise a Github issue if you need support.
--- a/app/lib/event_logger.rb
+++ b/app/lib/event_logger.rb
@@ -1,16 +1,16 @@
 class EventLogger
-  def self.log(tag, object)
-    Rails.logger.info "[#{tag}] #{object.to_json}"
+  def self.log(object)
+    Rails.logger.info object.to_json
   end
 
   def self.log_form_event(context, request, event)
-    item_to_log = {
+    log({
       url: request&.url,
       method: request&.method,
       form: context.form.name,
-    }
-
-    log("form_#{event}", item_to_log)
+      request_id: request&.request_id,
+      event: "form_#{event}",
+    })
   end
 
   def self.log_page_event(context, page, request, event, skipped_question)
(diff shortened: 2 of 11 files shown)


danworth:
Going to put it back into DRAFT: without `log_tags` then log events generated outside of Lograge and our custom logging (e.g. unhandled exceptions) do not include the request_id. Although there are still benefits, I would like to try and get request_id on all log entries as well as json format...

DavidBiddle (approved):
Looks good, nice one 🎉

PR #372, opened 2023-07-17
Convert National Insurance Number to component

aliuk2012:

Trello card:  https://trello.com/c/mGqPIFkV/237-switch-the-runner-to-use-components-rather-than-partials

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


spec/components/question/national_insurance_number_component/national_insurance_number_component_preview.rb
```diff
@@ -0,0 +1,24 @@
+class Question::NationalInsuranceNumberComponent::NationalInsuranceNumberComponentPreview < ViewComponent::Preview
+  def national_insurance_number_field
+    question = OpenStruct.new(number: "AB 123456 C",
```
DavidBiddle, line 3: Doesn't look like `number` is actually used in the component

aliuk2012: Good spot 👍 

Fixed up now

DavidBiddle (approved):

PR #373, opened 2023-07-18
Convert phone field to component

aliuk2012:

Asa separate commit, I also refactored this component to use `type="tel"` rather than `type="text"` as recommended by GOV.UK Design System https://design-system.service.gov.uk/patterns/telephone-numbers/

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/components/question/phone_number_component/view.html.erb
+++ b/app/components/question/phone_number_component/view.html.erb
@@ -1,0 +1,6 @@
+<%= form_builder.govuk_text_field :phone_number,
+                                  label: { tag: 'h1', size: 'l', text: question_text_with_extra_suffix },
+                                  hint: { text: question.hint_text },
+                                  width: 'one-half',
+                                  autocomplete: 'tel'
+%>
--- a/app/components/question/phone_number_component/view.html.erb
+++ b/app/components/question/phone_number_component/view.html.erb
@@ -1,6 +1,6 @@
 <%= form_builder.govuk_phone_field :phone_number,
-                                  label: { tag: 'h1', size: 'l', text: question_text_with_extra_suffix },
-                                  hint: { text: question.hint_text },
-                                  width: 'one-half',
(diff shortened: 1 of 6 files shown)


app/components/question/phone_number_component/view.html.erb
```diff
@@ -0,0 +1,6 @@
+<%= form_builder.govuk_phone_field :phone_number,
+                                  label: { tag: 'h1', size: 'l', text: question_text_with_extra_suffix },
+                                  hint: { text: question.hint_text },
+                                  width: 'one-half',
```
DavidBiddle, line 4: Can make this `width: 20` to match the design system pattern:
```suggestion
                                  width: 20,
```

spec/components/question/phone_number_component/view_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    it "renders the question text as a heading" do
+      expect(page.find("h1")).to have_text(question.question_text)
+    end
+
+    it "renders a telephone input field" do
+      expect(page).to have_css("input[type='tel'][name='form[phone_number]']")
+    end
+
+    it "renders at one half the width of the parent container" do
+      expect(page.native.to_html).to include('class="govuk-input govuk-!-width-one-half"')
```
DavidBiddle, line 27: If you go with the `width` change above:
```suggestion
      expect(page.native.to_html).to include('class="govuk-input govuk-input--width-20"')
```

DavidBiddle (approved):
Looks good. Made a suggestion which will bring this in line with the Design System pattern but it's optional

spec/components/question/phone_number_component/view_spec.rb
aliuk2012: Good suggestions, I've added that suggests in a separate commit.

DavidBiddle (approved):

PR #374, opened 2023-07-18
Add automated accessibility tests

DavidBiddle:
### What problem does the pull request solve?
- Adds axe-core-rspec for running automated accessibilty checks in our tests (see gem info below)
- Adds a feature tests to our existing components, which runs against the component previews and checks for accessibility issues. 
- Adds a couple of convenience methods:
  - Inside an `it "..." do` block within a feature test, you can use the methods `expect_page_to_have_no_axe_errors(page)` and `expect_component_to_have_no_axe_errors(page)` to assert that the page or component the browser is currently on doesn't have any errors.
- Adds a `base` layout which all component previews now use by default. The `application.html.erb` layout now inherits from this base layout

Trello card: https://trello.com/c/BARdutlj/916-explore-adding-automated-accessibility-tests

### New gem assessment
This PR adds a new gem, [axe-core-rspec](https://rubygems.org/gems/axe-core-rspec). To review the gem I looked at its stats on [ruby-toolbox.com](https://ruby-toolbox.com/), and also checked security issues using GitHub Security Advisories and the [snyk.io vulnerability database](https://security.snyk.io/).

The gem is maintained by Deque Systems, a major accessibility organisation which also develops the underlying axe-core library that this gem uses.

#### Does it solve a specific problem we have?
Yes, it lets us run automated accessibility tests within our Capybara feature tests. 
#### Is it updated regularly?
Yes, 16 releases so far this year
#### Is it widely used?
Somewhat - 2,973,244 downloads
#### Is it well documented?
The documentation is good, given the limited API of the gem.
#### Does it have any major security issues?
None that I could see.
#### Would writing the code ourselves have a better outcome?
Unlikely - we'd almost certainly still end up using Deque's underlying axe-core library, so we'd still have a dependency on Deque's code and wouldn't gain much. 
#### How much hassle would it be if this gem stopped being maintained?
Some, but not a lot. We would have to rewrite our accessibility tests in another framework (e.g. using jest-axe or a similar library) and we'd likely lose the convenience of having our accessibility tests integrated with our other rspec tests, but that doesn't feel like a major disadavantage.

#### Checklist

  - [n/a] README.md

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -85,6 +85,9 @@ group :test do
   gem "selenium-webdriver"
   gem "simplecov"
   gem "webdrivers"
+
+  # axe-core for running automated accessibility checks
+  gem "axe-core-rspec"
 end
 
 # For security auditing gem vulnerabilities. RUN IN CI
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -74,9 +74,20 @@ GEM
       i18n (>= 1.6, < 2)
       minitest (>= 5.1)
       tzinfo (~> 2.0)
-    addressable (2.8.0)
-      public_suffix (>= 2.0.2, < 5.0)
+    addressable (2.8.4)
+      public_suffix (>= 2.0.2, < 6.0)
     ast (2.4.2)
+    axe-core-api (4.7.0)
+      dumb_delegator
+      virtus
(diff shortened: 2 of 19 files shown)


thomasiles (approved):
looks good to me 👍

PR #379, opened 2023-07-19
Convert organisation name field to component

aliuk2012:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/spec/components/question/organisation_name_component/organisation_name_component_preview.rb
+++ b/spec/components/question/organisation_name_component/organisation_name_component_preview.rb
@@ -1,0 +1,24 @@
+class Question::OrganisationNameComponent::OrganisationNameComponentPreview < ViewComponent::Preview
+  def organisation_name_field
+    question = OpenStruct.new(text: "Organisations R Us",
+                              answer_type: "organisation_name",
+                              question_text: "What is your company name?",
+                              answer_settings: nil)
+    form_builder = GOVUKDesignSystemFormBuilder::FormBuilder.new(:form, question,
+                                                                 ActionView::Base.new(ActionView::LookupContext.new(nil), {}, nil), {})
+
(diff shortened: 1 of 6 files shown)


spec/components/question/organisation_name_component/organisation_name_component_preview.rb
```diff
@@ -0,0 +1,24 @@
+class Question::OrganisationNameComponent::OrganisationNameComponentPreview < ViewComponent::Preview
+  def organisation_name_field
+    question = OpenStruct.new(text: "Organisations R Us ",
```
DavidBiddle, line 3: Is the trailing space deliberate? If not then think it can be removed:

```suggestion
    question = OpenStruct.new(text: "Organisations R Us",
```

spec/components/question/organisation_name_component/organisation_name_component_preview.rb
```diff
@@ -0,0 +1,24 @@
...
+                              question_text: "What is your company name?",
+                              answer_settings: nil)
+    form_builder = GOVUKDesignSystemFormBuilder::FormBuilder.new(:form, question,
+                                                                 ActionView::Base.new(ActionView::LookupContext.new(nil), {}, nil), {})
+
+    render(Question::OrganisationNameComponent::View.new(form_builder:, question:, extra_question_text_suffix: ""))
+  end
+
+  def organisation_name_field_with_hint
+    question = OpenStruct.new(text: "Organisations R Us ",
```
DavidBiddle, line 14: As above:
```suggestion
    question = OpenStruct.new(text: "Organisations R Us",
```

spec/components/question/organisation_name_component/organisation_name_component_preview.rb
aliuk2012: good spot, thanks @DavidBiddle

DavidBiddle (approved):

PR #385, opened 2023-07-22
Fix missing optional copy in labels from components

aliuk2012:
As part of the refactoring of the components we forgot to add "(optional)" to the question label if the question is optional. This is because before  partials were responsible for calling a helper method which worked out whether or not to add "(optional)" suffix and view mode. This is a lot for a helper method. The components are "dumb" and just handle simple presentation. Manipulating data or business logic is best controlled outside a component.

This moves the business logic of whether or not to add "(optional)" from the helper method to the question model instead. It makes testing this code a lot easier for both the helper method and the new model method.

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/models/question/question_base.rb
+++ b/app/models/question/question_base.rb
@@ -40,7 +40,7 @@ module Question
     end
 
     def question_text_with_optional_suffix
-      return question_text unless is_optional?
+      return question_text unless show_optional_suffix
 
       "#{question_text} #{I18n.t('page.optional')}"
     end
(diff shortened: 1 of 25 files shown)


app/models/question/question_base.rb
```diff
@@ -38,5 +38,11 @@ def has_long_answer?
     def show_optional_suffix
       is_optional?
     end
+
+    def question_text_with_optional_suffix
+      return question_text unless show_optional_suffix
+
+      "#{question_text} #{I18n.t('page.optional')}"
+    end
```
aliuk2012, line 46: We will be able to remove this once we have converted the last three answer types to components

thomasiles (approved):
Looks good, a lot less complicated 🥇

PR #388, opened 2023-07-27
Only Bundle Install Production Scope Only

danworth:
We do not need to install the development and test scoped dependencies into our docker image.

See similar PR for forms-admin: https://github.com/alphagov/forms-admin/pull/561

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,11 @@ USER ruby
 
 COPY --chown=ruby:ruby Gemfile* ./
 RUN gem install bundler -v 2.4.10
-RUN bundle install --jobs "$(nproc)"
+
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -12,6 +12,10 @@ USER ruby
 
 COPY --chown=ruby:ruby Gemfile* ./
 RUN gem install bundler -v 2.4.10
+
+RUN bundle config set --local without development:test \
+  && bundle config set --local jobs "$(nproc)"
+
 RUN bundle install --jobs "$(nproc)"
```
lfdebrux, line 19:
```suggestion
RUN bundle install"
```

lfdebrux (approved):

PR #406, opened 2023-08-18
Convert name answer type partial to a view component

aliuk2012:

As part of this refactor it will mean that this answer type can now be easily
ported to other apps because it doesn't rely on specific application specific
code. There is only one dependency for it which is the govuk form builder gem.

Trello card: https://trello.com/c/CUsSXIOy/973-convert-name-answer-type-partial-to-view-component

Try viewing demos found here http://localhost:3001/preview/question/name_component/name_component


thomasiles (approved):
So much nicer!
 🎖️

PR #407, opened 2023-08-18
Add guidance component

DavidBiddle:

Trello card: https://trello.com/c/bfw5T1Ny/975-create-guidance-component-in-forms-runner

Adds a new component for displaying the page heading and rendering the markdown guidance.

As part of that I've installed the [govuk-forms-markdown](https://github.com/alphagov/govuk-forms-markdown) gem and added git to our Docker build.

Note that this doesn't actually add the new component to the view - that'll be done either as part of https://trello.com/c/R5aZ0LWX/950-update-forms-runner-to-display-additional-guidance-to-form-fillers or in a separate PR once that one's merged.

### DIFF
--- a/app/components/guidance_component/view.rb
+++ b/app/components/guidance_component/view.rb
@@ -1,0 +1,20 @@
+require "govuk_forms_markdown"
+
+module GuidanceComponent
+  class View < ViewComponent::Base
+    attr_accessor :question
+
+    def initialize(question)
+      super
+      @question = question
+    end
+
+    def render?
+      question.page_heading.present? && question.guidance_markdown.present?
+    end
+
+    def guidance_html
+      ActionController::Base.helpers.sanitize(GovukFormsMarkdown.render(question.guidance_markdown), scrubber: MarkdownScrubber.new)
+    end
+  end
+end
--- a/app/lib/markdown_scrubber.rb
+++ b/app/lib/markdown_scrubber.rb
@@ -1,0 +1,7 @@
+class MarkdownScrubber < Rails::Html::PermitScrubber
+  def initialize
+    super
+    self.tags = %w[a h2 h3 ol ul li p]
+    self.attributes = %w[href class rel target title]
+  end
+end
--- a/spec/lib/markdown_scrubber_spec.rb
+++ b/spec/lib/markdown_scrubber_spec.rb
@@ -1,0 +1,42 @@
+require "rails_helper"
+
+RSpec.describe MarkdownScrubber do
+  let(:markdown_scrubber) { described_class.new }
+
+  it "returns an array of html tags allowed" do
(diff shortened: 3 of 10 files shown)


app/components/guidance_component/view.rb
```diff
@@ -0,0 +1,20 @@
...
+      super
+      @question = question
+    end
+
+    def render?
+      question.page_heading.present? && question.guidance_markdown.present?
+    end
+
+    def guidance_html
+      GovukFormsMarkdown.render(question.guidance_markdown).html_safe
```
aliuk2012, line 17: 🤔do we need `html_safe` here? if so, maybe we should mark safe here instead so we dont have to do it everytime we use our custom renderer https://github.com/alphagov/govuk-forms-markdown/blob/main/lib/govuk_forms_markdown.rb#L11

DavidBiddle: Ah, the gem doesn't have `html_safe` or `sanitize` baecause it's not a Rails project - so I think I'll leave it here for now. Will convert it to a `sanitize` call though

app/lib/markdown_scrubber.rb
```diff
@@ -0,0 +1,7 @@
+class MarkdownScrubber < Rails::Html::PermitScrubber
+  def initialize
+    super
+    self.tags = %w[a h2 h3 ol ul li p]
+    self.attributes = %w[href class rel target title]
+  end
+end
```
aliuk2012, line 7: Could we add a few tests around this? 
Maybe a few happy paths followed by a few where input is sanitised correctly.

DavidBiddle: Good suggestion, have added 😄

spec/lib/markdown_scrubber_spec.rb
```diff
@@ -0,0 +1,41 @@
+require "rails_helper"
+
+RSpec.describe MarkdownScrubber do
+  describe "#environment_name" do
```
aliuk2012, line 4: Is this describe needed ?

spec/lib/markdown_scrubber_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    let(:markdown_scrubber) { described_class.new }
+
+    it "returns the value of forms_env" do
```
aliuk2012, line 7: Could we maybe improve the description? It doesn't seem to describe what it is we are testing.

app/lib/markdown_scrubber.rb
aliuk2012: Specs look good just a copy minor suggestions but otherwise this PR is just about there 🚢

aliuk2012 (approved):

SamJamCul (approved):

PR #408, opened 2023-08-22
Toggle page headings depending on if question has guidance

aliuk2012:
- refactors how we set answer type view components tag and font size
- passes page_heading/additional_guidance_markdown to the QuestionRegister so we can use it on the page and components.
- adjust all the answer type view components to toggle between using heading and native legend/labels

Trello card: https://trello.com/c/R5aZ0LWX/950-update-forms-runner-to-display-additional-guidance-to-form-fillers

### DIFF
--- a/app/lib/question_register.rb
+++ b/app/lib/question_register.rb
@@ -25,6 +25,13 @@ class QuestionRegister
               raise ArgumentError, "Unexpected answer_type for page #{page.id}: #{page.answer_type}"
             end
     hint_text = page.respond_to?(:hint_text) ? page.hint_text : nil
-    klass.new({}, { question_text: page.question_text, hint_text:, is_optional: page.is_optional, answer_settings: page.answer_settings })
+    page_heading = page.respond_to?(:page_heading) ? page.page_heading : nil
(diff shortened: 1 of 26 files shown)


app/lib/question_register.rb
```diff
@@ -25,6 +25,13 @@ def self.from_page(page)
               raise ArgumentError, "Unexpected answer_type for page #{page.id}: #{page.answer_type}"
             end
     hint_text = page.respond_to?(:hint_text) ? page.hint_text : nil
-    klass.new({}, { question_text: page.question_text, hint_text:, is_optional: page.is_optional, answer_settings: page.answer_settings })
+    page_heading = page.respond_to?(:page_heading) ? page.page_heading : nil
+    guidance_markdown = page.respond_to?(:additional_guidance_markdown) ? page.additional_guidance_markdown : nil
```
SamJamCul, line 29: These feel a bit weird, in an ideal world what would our API be giving us so that we didn't need this?

SamJamCul (approved):

app/lib/question_register.rb
SamJamCul: Ah right, makes sense!

PR #409, opened 2023-08-22
Increase width of single line of text input

lfdebrux:

Trello card: https://trello.com/c/yzKobeRo/977-single-line-text-inputs-should-be-wider

We want the width of the input to be at least two-thirds the width of the page. This commit just removes the `one-half` width qualifier, so the input fluidly fills the space available. As the form group is in a two-thirds grid column, this fulfils the requirement.

<details>
<summary><h3>Screenshots of changes</h3></summary>

#### Before

Screenshot of single line of text input before changes

#### After

Screenshot of single line of text input after changes

</details>

I went through the all question types form by hand and checked that only the single line of text question was affected.


jamie-o-wilkinson (approved):
Nice one! 👍

aliuk2012 (dismissed):

aliuk2012 (approved):

PR #386, opened 2023-07-24
Refactor build files

lfdebrux:
This PR takes the changes made by @chao-xian in alphagov/forms-api#299 and applies them to this repo.

PR #417, opened 2023-09-07
Remove web-console gem

chao-xian:

The web-console gem is never used. It simply provides a Rails console in the browser. It also seems to be throwing an error:

```
Web Console is activated in the test environment. This is
usually a mistake. To ensure it's only activated in development
mode, move it to the development group of your Gemfile:

    gem 'web-console', group: :development

If you still want to run it in the test environment (and know
what you are doing), put this in your Rails application
configuration:

    config.web_console.development_only = false
```

since we put everything into the [default rake task](https://github.com/alphagov/forms-runner/pull/386) but can't work out why so removing for now rather than wasting more time on debugging. 

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -76,17 +76,6 @@ group :development, :test do
   gem "brakeman", "~> 6.0"
 end
 
-group :development do
-  # Use console on exceptions pages [https://github.com/rails/web-console]
-  gem "web-console"
-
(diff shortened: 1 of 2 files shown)


chao-xian:
I'm minded to also remove it from the forms-admin repo just for consistency.

aliuk2012 (approved):

PR #421, opened 2023-09-12
Log form_id with events

lfdebrux:

To make it easier to correlate log messages from `event_logger` with other log messages, add the `form_id` to the log message. In our lograge configuration we use the form ID instead of the form name.

### DIFF
--- a/app/lib/event_logger.rb
+++ b/app/lib/event_logger.rb
@@ -8,6 +8,7 @@ class EventLogger
       url: request&.url,
       method: request&.method,
       form: context.form.name,
+      form_id: context.form.id,
       request_id: request&.request_id,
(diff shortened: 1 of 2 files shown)


app/lib/event_logger.rb
```diff
@@ -8,6 +8,7 @@ def self.log_form_event(context, request, event)
       url: request&.url,
       method: request&.method,
       form: context.form.name,
+      form_id: context.form.id,
```
aliuk2012, line 11: should there be a test for this?

lfdebrux: Yep, fixed now!

DavidBiddle (approved):
Looks good to me 🎉

PR #430, opened 2023-09-28
1059 increment metric in cloudwatch when a form is submitted

SamJamCul:

Trello card: https://trello.com/c/sSoCpSfL/1059-save-as-a-metric-in-cloudwatch-when-a-form-is-submitted

This PR contains a couple of changes:
* Refactoring the LogEventService to incorporate the EventLogger for for submissions. This involved moving the EventLogger call into here. 
* Adding a call to cloudwatch event logging. This is triggered when a form is submitted, and sends an event to Cloudwatch. 

This is a bit of a pain to test locally, but it can be done. You'll need to run the app with AWS credentials, which is doable from the command line with something like `$ gds-cli aws forms-dev-admin -- bundle exec rails s`. This will start the app up with the appropriate credentials to enable Cloudwatch logs. When you submit a form, it will send the form_id as part of a metric to Cloudwatch. You can find this Cloudwatch event by going on to the AWS Cloudwatch interface, and looking for your form ID in the appropriate namespace. I can demo this to anyone interested!

### DIFF
--- a/app/services/log_event_service.rb
+++ b/app/services/log_event_service.rb
@@ -13,7 +13,11 @@ class LogEventService
 
   def self.log_submit(context, request)
     EventLogger.log_form_event(context, request, "submission") # Logging to Splunk
-    CloudWatchService.log_form_submission(form_id: context.form.id) # Logging to Cloudwatch
+    begin
+      CloudWatchService.log_form_submission(form_id: context.form.id) # Logging to Cloudwatch
+    rescue StandardError => e
+      Sentry.capture_exception(e)
+    end
   end
 
 private
--- a/app/services/log_event_service.rb
+++ b/app/services/log_event_service.rb
@@ -12,7 +12,8 @@ class LogEventService
   end
 
   def self.log_submit(context, request)
-    EventLogger.log_form_event(context, request, "submission")
+    EventLogger.log_form_event(context, request, "submission") # Logging to Splunk
+    CloudWatchService.log_form_submission(form_id: context.form.id) # Logging to Cloudwatch
   end
 
 private
--- a/app/services/log_event_service.rb
+++ b/app/services/log_event_service.rb
@@ -11,6 +11,10 @@ class LogEventService
     EventLogger.log_page_event(@current_context, @step, @request, log_event, skipped_question?)
   end
 
+  def self.log_submit(context, request)
+    EventLogger.log_form_event(context, request, "submission")
+  end
+
 private
 
   def log_event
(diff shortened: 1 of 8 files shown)


spec/services/cloud_watch_service_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    let(:form_id) { 3 }
+    let(:forms_env) { "test" }
+
+    before do
+      allow(Settings).to receive(:forms_env).and_return(forms_env)
+    end
+
+    it "calls the cloudwatch client with put_metric_data" do
+      # Stub the CloudWatch client and put_metric_data method
+      cloudwatch_client = Aws::CloudWatch::Client.new(stub_responses: true)
```
chao-xian, line 15: Nice!

app/services/log_event_service.rb
```diff
@@ -13,7 +13,11 @@ def log_page_save
 
   def self.log_submit(context, request)
     EventLogger.log_form_event(context, request, "submission") # Logging to Splunk
-    CloudWatchService.log_form_submission(form_id: context.form.id) # Logging to Cloudwatch
+    begin
+      CloudWatchService.log_form_submission(form_id: context.form.id) # Logging to Cloudwatch
+    rescue StandardError => e
```
chao-xian, line 18: Can we add a set number of retries? Something like the second example in https://blog.appsignal.com/2018/05/16/ensure-retry-and-reraise-exceptions-in-ruby.html#giving-up-reraising-exceptions-using-raise

chao-xian (commented):
Looks good but just wondering if we can retry so we lessen the chance of stats being wrong.

app/services/log_event_service.rb
SamJamCul: I've added a retry loop for the cloudwatch logging - this will retry up to three times, before giving up and logging whatever the error was with Sentry. 

I did discuss this with some others, and because we don't have a queue system for this, it's possible that it could take a while to resolve. As this happens when a form is submitted, the next page won't get shown until this retry loop is finished. I can't find anywhere else in our apps that we've got this, so there might start being a case for introducing a queue that we can put this sort of asynchronous request in.

chao-xian: Ah that's a good point... I'm now unsure whether to include this retry (sorry!). What do you think? Shall we leave it out?

SamJamCul: It's out! Good to have it in the pocket in case we see some weirdness though

chao-xian: Thank you!

chao-xian (approved):

PR #449, opened 2023-10-18
Add confirmation email form

DavidBiddle:

Trello card: https://trello.com/c/ArOlWLI0/1105-build-confirmation-email-form-on-the-check-your-answers-page

This PR:
- Adds new form fields to the Check Your Answers page, asking the user whether they want to receive email confirmation and where it should be sent
- Adds validation to the above form, so it can't be submitted unless a user has told us their preference and supplied an email address if appropriate
- Removes the `submit_answers_controller` and moves the submission logic into the `check_your_answers_controller`- this makes implementing the validation easier, and it also matches the pattern of a single controller with show and submit actions that we use elsewhere in the service.

#### Form placement on check your answers page:
A check your answers page with the email confirmation form positioned in between the main answer summary and the declaration

#### Default form view
A field with two radio buttons and the label 'Do you want to get an email confirming your form has been submitted?'. It has a hint 'We’ll only use this email to send a confirmation that your form’s been successfully submitted. It will not contain a copy of your answers.' and the tradio options are 'Yes' and 'No'.

#### Form validation (no preference specified)
The default form view, but the field is showing an error 'Select ‘Yes’ if you want to get an email confirming your form has been submitted'.

#### Form with email field expanded
The default form view, but with 'Yes' selected and a conditional input field with the label 'What email address do you want us to send your confirmation to?' appearing underneath the yes option.

#### Form validation (no email provided):
The form with email field expanded, but the email field is showing an error 'Enter an email address'.

#### Form validation (invalid email provided):
The form with email field expanded, but the email field is showing an error 'Enter an email address in the correct format, like name@example.com'.


app/controllers/forms/check_your_answers_controller.rb
```diff
@@ -8,11 +8,48 @@ def show
       @rows = check_your_answers_rows
       @form_submit_path = form_submit_answers_path
       @notify_reference ||= SecureRandom.uuid
+      email_confirmation_form = EmailConfirmationForm.new(notify_reference: @notify_reference)
```
aliuk2012, line 11:
```suggestion
      email_confirmation_form = EmailConfirmationForm.new(notify_reference: SecureRandom.uuid)
```

chao-xian:
This is looking good so far

spec/factories/form_objects/email_confirmation_form.rb
```diff
@@ -0,0 +1,7 @@
+FactoryBot.define do
+  factory :email_confirmation_form, class: "EmailConfirmationForm" do
+    send_confirmation { nil }
+    confirmation_email_address { nil }
+    notify_reference { Faker::Internet.uuid }
```
aliuk2012, line 5:
```suggestion
    send_confirmation { nil }
    confirmation_email_address { nil }
    notify_reference {  SecureRandom.uuid }
    
    factory :send_email_confirmation_form do
    	send_confirmation { true }
      confirmation_email_address { Faker::Internet.email }
    end
```

It feels better to have a "different" factory for instances where the user would like to send emails and its feels better than using a trait to declare these values

You should be able to do something like 
`build :send_email_confirmation_form` 

Could you maybe make use of this factory in your specs?

DavidBiddle: Good suggestion - have done this (changed the name slightly)

app/views/forms/check_your_answers/show.html.erb
```diff
@@ -8,24 +8,38 @@
...
+          } })
+        %>
+      <% end %>
+
+      <% if FeatureService.enabled?(:email_confirmations_enabled) %>
+        <%= form.govuk_radio_buttons_fieldset(:send_confirmation, legend: { size: 'm', tag: 'h2' }) do %>
+          <%= form.govuk_radio_button :send_confirmation, 'true' do %>
+            <%= form.govuk_text_field :confirmation_email_address %>
+          <% end %>
+          <%= form.govuk_radio_button :send_confirmation, 'false' %>
```
aliuk2012, line 32: Instead of true/false as strings and having to deal with all the weirdness of boolean values as strings. could the `send_confirmation` values be `send_email`/`skip_confirmation` or potentially just `send`/`skip` ?  

You could also add validations to the form object to make sure the `send_confirmation` is either of these values (users shouldn't see the errors but its helpful if a dev is trying to use FormObject outside the context of the form or if the ui is updated to be list of checkboxes it would fail because you couldn't have both values.

config/locales/en.yml
```diff
@@ -91,6 +98,19 @@ en:
...
+  helpers:
+    hint:
+      email_confirmation_form:
+        send_confirmation: We’ll only use this email to send a confirmation that your form’s been successfully submitted. It will not contain a copy of your answers.
+    label:
+      email_confirmation_form:
+        confirmation_email_address: What email address do you want us to send your confirmation to?
+        send_confirmation_options:
+          'false': 'No'
+          'true': 'Yes'
```
aliuk2012, line 110:
```suggestion
          send: 'Yes'
          skip: 'No'
```

or 

```suggestion
          send_email: 'Yes'
          skip_email: 'No'
```

spec/factories/form_objects/email_confirmation_form.rb
```diff
@@ -0,0 +1,12 @@
...
```
aliuk2012, line 5:
```suggestion
    notify_reference {  SecureRandom.uuid }
```

Dont think we need faker for the uuid

spec/factories/form_objects/email_confirmation_form.rb
aliuk2012: Like you the name you used, its more descriptive 👍

spec/factories/form_objects/email_confirmation_form.rb
DavidBiddle: Done!

app/views/forms/check_your_answers/show.html.erb
DavidBiddle: Done - went with 'send_email'/'skip_confirmation' for explicitness

app/controllers/forms/check_your_answers_controller.rb
```diff
@@ -3,16 +3,35 @@ class CheckYourAnswersController < BaseController
...
+          redirect_to :form_submitted
+        end
+      else
+        setup_check_your_answers
+        email_confirmation_form.notify_reference = SecureRandom.uuid
+        render template: "forms/check_your_answers/show", locals: { email_confirmation_form: }, status: :unprocessable_entity
+      end
+    rescue StandardError => e
+      Sentry.capture_exception(e)
+      render "errors/submission_error", status: :internal_server_error
```
aliuk2012, line 34: This is a bit unwieldy but we agreed that as a first pass its fine and we will refactor it in the next day or too

aliuk2012 (approved):
Thanks for addressing all the comments, looks really good for first pass.

PR #451, opened 2023-10-19
refactor form_submission_service to have a main method to called

aliuk2012:
We have one method which we call to email form submissions but we
want to add other functionality to form_submission_service. eg
once form has been submitted then send confirmation email to form
filler. This PR is the groundwork for these enhancements

Trello card: https://trello.com/c/A7YmTtGq/1104-send-confirmation-email-to-form-filler


chao-xian:
Yes, I agree it's a good change! 😄 TY both

PR #452, opened 2023-10-19
Send confirmation email to form fillers

aliuk2012:

- Adds new email template id for confirmation emails
- Introduces a new mailer to send confirmation emails to form fillers
- Calls the new mailer from form_submission_service. if the feature flag is enabled and form filler provided a confirmation email
- refactors form submission service to accept a form_object instead of single params being passed in.
- fixes the time format used in e-mails to not include seconds and to use 12-hour clock

Trello card: https://trello.com/c/A7YmTtGq/1104-send-confirmation-email-to-form-filler


DavidBiddle (approved):
Looks good! 🎉

PR #456, opened 2023-10-25
Adding Support contact details for phone and online methods

aliuk2012:

Adds phone number and online contact methods if they are setup for the form

Screenshot 2023-10-30 at 14 36 08

Trello card: https://trello.com/c/r4RTqUPa/1143-add-all-support-contact-detail-methods-to-confirmation-email

### DIFF
--- a/app/services/form_submission_service.rb
+++ b/app/services/form_submission_service.rb
@@ -133,6 +133,27 @@ private
   def formatted_support_details
     return nil unless has_support_contact_details?
 
-    @form.support_email.presence
+    [support_phone_details, support_email_details, support_online_details].compact_blank.join("\n\n")
+  end
+
+  def support_phone_details
+    return nil if @form.support_phone.blank?
+
+    notify_body = NotifyTemplateBodyFilter.new
+    formatted_phone_number = notify_body.normalize_whitespace(@form.support_phone)
+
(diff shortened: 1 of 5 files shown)


app/services/form_submission_service.rb
```diff
@@ -133,6 +133,17 @@ def has_support_contact_details?
   def formatted_support_details
     return nil unless has_support_contact_details?
 
-    @form.support_email.presence
+    details = []
+    details << setup_support_phone if @form.support_phone.present?
+    details << "[#{@form.support_email}](mailto:#{@form.support_email.presence})" if @form.support_email.present?
+    details << "[#{@form.support_url_text}](#{@form.support_url})" if [@form.support_url, @form.support_url_text].all?(&:present?)
```
chao-xian, line 139: Can these two lines be abstracted into another method similar tot he setup_support_phone one?

aliuk2012: good point, We could do one better and just call each method as part of the array.

aliuk2012: good point, We could do one better and just call each method as part of the array.

jamie-o-wilkinson (approved):
LGTM. Splitting those formatting bits was worth it IMO 👍

chao-xian (approved):

chao-xian (approved):

PR #476, opened 2023-10-31
Remove dotenv-rails gem

lfdebrux:

We are using config gem instead of dotenv-rails [[1]], so we should remove it.

[1]: https://github.com/alphagov/forms-admin/commit/3ffbf1598871e36332d75fa200d62aaf4e0d60f1

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,8 +13,6 @@ gem "puma", "~> 6.4.0"
 gem "sentry-rails"
 gem "sentry-ruby"
 
-gem "dotenv-rails", groups: %i[development test]
-
 gem "config"
 
 # Use GOV.UK Nofity api to send emails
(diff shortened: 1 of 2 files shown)


aliuk2012:
do we need this though for redis url?

Also do we still need the `.env` file?

lfdebrux:
@aliuk2012 the Redis URL still comes from an environment variable:

https://github.com/alphagov/forms-runner/blob/main/config/initializers/session_store.rb#L6-L7

https://github.com/alphagov/forms-runner/blob/main/app/lib/redis_config.rb

That doesn't necessarily mean we need a `.env` file though, we don't have a template one any more. If you're running forms-runner locally I don't think you'd need to be using Redis in most circumstances?

aliuk2012 (approved):

PR #480, opened 2023-11-02
Convert 'what happens next' text to Notify markup

DavidBiddle:

Trello card: https://trello.com/c/MEZAhBet/1157-add-support-for-links-in-confirmation-emails

Converts HTML entered into the 'what happens next' field by form creators into Notify-compatible markdown when the form is submitted, so that the email a form completer receives will include the formatted what happens next text. Only supports paragraphs, links and lists.

Also includes some general refactoring of the way we use `simple_format` - HTML rendering should now be more predictable. I'd advise that we don't use `simple_format` directly in views any more - we should use one of the methods in `html_markdown_sanitizer.rb` instead (or add a new one) to keep things consistent. See the individual commit messages for more detail.

### Example
What happens next text in forms-admin:
```plain
We aim to respond to all applications within 5 working days, but this may vary at busy times. Our service status page shows our <a href="https://example.com" target="_blank" rel="noopener noreferrer">current expected waiting times</a>.

While your case is being processed, you must not do any of the following in relation to your cat:
<ul class="govuk-list govuk-list--bullet">
<li>feed it onions</li>
<li>post pictures of it on the internet</li>
<li>allow it to use your laptop when you're not at your desk</li>
</ul>
```
Rendered HTML on submitted page:
```html
  <p>We aim to respond to all applications within 5 working days, but this may vary at busy times. Our service status page shows our <a href="https://example.com" target="_blank" rel="noopener noreferrer">current expected waiting times</a>.</p>

<p>While your case is being processed, you must not do any of the following in relation to your cat:
</p><ul class="govuk-list govuk-list--bullet">
<li>feed it onions</li>
<li>post pictures of it on the internet</li>
<li>allow it to use your laptop when you're not at your desk</li>
</ul>
```
Markdown sent to Notify:
```md
We aim to respond to all applications within 5 working days, but this may vary at busy times. Our service status page shows our [current expected waiting times](https://example.com).

While your case is being processed, you must not do any of the following in relation to your cat:

- feed it onions
- post pictures of it on the internet
- allow it to use your laptop when you&#39;re not at your desk
```

#### Email screenshot
A heading 'What happens next', followed by a paragraph 'We aim to respond to all applications within 5 working days, but this may vary at busy times. Our service status page shows our current expected waiting times.' where 'current expected waiting times' is a link. This is followed by a paragraph 'While your case is being processed, you must not do any of the following in relation to your cat:', and a bullet list of three items: 'feed it onions', 'post pictures of it on the internet', and 'allow it to use your laptop when you’re not at your desk'.


app/helpers/application_helper.rb
```diff
@@ -38,10 +38,6 @@ def hidden_text_mode(mode)
     "<span class='govuk-visually-hidden'>&nbsp;#{t("page.#{mode_name}_preview".to_s)}</span>".html_safe
   end
 
-  def format_paragraphs(text)
-    simple_format(html_escape(text), class: "govuk-body", sanitize: true)
-  end
-
```
lfdebrux, line 44: Is there any reason not to include a helper method `format_paragraphs` that just calls `HtmlMarkdownSanitizer.new.format_paragraphs` for us?

Gemfile
```diff
@@ -54,6 +54,10 @@ gem "uk_postcode"
 # For structured logging
 gem "lograge"
 
+gem "aws-sdk-cloudwatch"
+
+gem "reverse_markdown", "~> 2.1"
```
lfdebrux, line 59: Do you have any benchmarking information on the reverse markdown process?

app/models/form.rb
```diff
@@ -32,4 +32,8 @@ def live?(current_datetime = Time.zone.now)
   def live_at_date
     try(:live_at).try(:to_time)
   end
+
+  def what_happens_next_markdown
+    ReverseMarkdown.convert(HtmlMarkdownSanitizer.new.render_scrubbed_html(what_happens_next_text)).strip
```
lfdebrux, line 37: My assumption was that the `what_happens_next_text` would already be Markdown; I'm guessing though this round-tripping is necessary in case the user has put their own HTML in?

I'm curious to know, what does Notify do if they are sent HTML in the template personalisation?

app/helpers/application_helper.rb
DavidBiddle: No reason, so have added 👍

app/models/form.rb
DavidBiddle: It's not actually markdown at the moment - it's a slightly odd plaintext/HTML hybrid field. 

For context - when we implemented it initially we thought it was a plaintext field and the `simple_format` would just split it into paragraphs according to where the line breaks were. In reality `simple_format` is actually quite permissive and allows a subset of HTML (sanitized to prevent XSS). We didn't remove this unintended behaviour because it enabled some of our early private beta users to put links in their what happens next text, which they identified as a need. Our plan is to make this a markdown field sooner rather than later which would make this reverse markdown manoevre unnecessary.

I'm pretty sure Notify will escape the characters, so users would see the text `Our service status page shows our <a href="https://example.com" target="_blank" rel="noopener noreferrer">current expected waiting times</a>.` in their email.

Gemfile
DavidBiddle: I don't but I'll try and generate some!

DavidBiddle: I don't know if there's a better way of doing this, but:
```
what_happens_next_text = "We aim to respond to all applications within 5 working
 days, but this may vary at busy times. Our service status page shows our <a href=\"https://exa
mple.com\" target=\"\_blank\" rel=\"noopener noreferrer\">current expected waiting times</a>.\n
\nWhile your case is being processed, you must not do any of the following in relation to your 
cat:\n<ul class=\"govuk-list govuk-list--bullet\">\n<li>feed it onions</li>\n<li>post pictures 
of it on the internet</li>\n<li>allow it to use your laptop when you're not at your desk</li>\n
</ul>\n"

puts Benchmark.measure{ 1000.times do;ReverseMarkdown.convert(HtmlMarkdownSaniti
zer.new.render_scrubbed_html(what_happens_next_text)).strip; end }
```

returns:
```
user            system      total          real
1.483626   0.015281   1.498907 (  1.508772)
```

So for the example output in the PR description it's taking ~1.5ms for each call to `ReverseMarkdown.convert(HtmlMarkdownSaniti
zer.new.render_scrubbed_html(what_happens_next_text)).strip`

app/models/form.rb
lfdebrux: Ah, I see, I was getting confused with detailed guidance. It's good to know it will eventually be markdown though, maybe we can put a `FIXME` comment in there reminding future readers to check if it's still necessary?

Gemfile
lfdebrux: Okay, so not great but not terrible.

app/models/form.rb
DavidBiddle: Good idea - added that now 👍

aliuk2012 (approved):
L

aliuk2012 (approved):
L

PR #482, opened 2023-11-03
Use blank? instead of nil? in heading size check

DavidBiddle:

Trello card: 

Fixes an issue with the wrong heading size and tag being returned for a question with an empty string as its `page_heading`. We think this arises when a page has detailed guidance which is subsequently removed.

### DIFF
--- a/spec/components/question/base_spec.rb
+++ b/spec/components/question/base_spec.rb
@@ -1,0 +1,72 @@
+require "rails_helper"
+
+RSpec.describe Question::Base, type: :component do
+  let(:extra_question_text_suffix) { nil }
+  let(:form_builder) do
+    GOVUKDesignSystemFormBuilder::FormBuilder.new(:form, question,
+                                                  ActionView::Base.new(ActionView::LookupContext.new(nil), {}, nil), {})
+  end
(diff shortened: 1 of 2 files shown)


spec/components/question/base_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+      end
+    end
+
+    context "when the guidance markdown is empty" do
+      let(:guidance_markdown) { "" }
+
+      it "returns config for a medium default tag" do
+        expect(question_base.question_text_size_and_tag).to eq({ size: "m" })
+      end
+    end
```
lfdebrux, line 52: These tests confused me at first, because they're written backwards almost... I think what we actually care about is that the page heading or guidance markdown is present, rather than either of those being blank or nil, right?

lfdebrux (dismissed):

spec/components/question/base_spec.rb
DavidBiddle: @lfdebrux Good point - I've updated the test/context names to make it a bit clearer what's happening here which should hopefully help

lfdebrux (approved):

PR #497, opened 2023-11-17
1184 prevent forms runner and forms admin cloudwatch requests when in unsuitable environments

SamJamCul:

Trello card: https://trello.com/c/eGLh3RTp/1184-prevent-forms-runner-and-forms-admin-cloudwatch-requests-when-in-unsuitable-environments
Now that https://github.com/alphagov/forms-deploy/pull/435 is in, we can make use of the environment variable that it's set to turn off metrics in dev and user research. 

This PR makes the CloudWatchService check the environment variable `cloudwatch_metrics_enabled` before it continues making CloudWatch requests, and breaks early if they're disabled. 

### DIFF
--- a/app/services/cloud_watch_service.rb
+++ b/app/services/cloud_watch_service.rb
@@ -2,6 +2,8 @@ class CloudWatchService
   REGION = "eu-west-2".freeze
 
   def self.log_form_submission(form_id:)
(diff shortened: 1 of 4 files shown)


DavidBiddle (approved):
Looks good!

DavidBiddle (approved):

PR #500, opened 2023-11-20
Add notification reference for confirmation email

lfdebrux:

Trello card: https://trello.com/c/bnWjTvpU/1197-add-an-e2e-test-to-cover-adding-confirmation-email-being-filled-in-on-forms 

We want to be able to find the confirmation email for a form filler in the GOV.UK Notify API. This is currently for our end to end tests, so we can check that a notification is actually queued.

This commit adds a notification reference for the confirmation email, similar to how we already have a notification reference for the email with the submission. To make it easier for any humans looking at the Notify logs to associate the two emails together, we give them both the same random prefix, and differentiate with a human readable suffix.

### DIFF
--- a/app/controllers/forms/check_your_answers_controller.rb
+++ b/app/controllers/forms/check_your_answers_controller.rb
@@ -5,7 +5,6 @@ module Forms
 
       setup_check_your_answers
       email_confirmation_form = EmailConfirmationForm.new
-      email_confirmation_form.generate_submission_references!
 
       render template: "forms/check_your_answers/show", locals: { email_confirmation_form: }
     end
@@ -30,7 +29,6 @@ module Forms
         end
       else
         setup_check_your_answers
-        email_confirmation_form.generate_submission_references!
 
         render template: "forms/check_your_answers/show", locals: { email_confirmation_form: }, status: :unprocessable_entity
       end
--- a/app/controllers/forms/check_your_answers_controller.rb
+++ b/app/controllers/forms/check_your_answers_controller.rb
@@ -4,7 +4,8 @@ module Forms
       return redirect_to form_page_path(current_context.form.id, current_context.form.form_slug, current_context.next_page_slug) unless current_context.can_visit?(CheckYourAnswersStep::CHECK_YOUR_ANSWERS_PAGE_SLUG)
 
       setup_check_your_answers
-      email_confirmation_form = EmailConfirmationForm.new(**submission_reference_attributes)
+      email_confirmation_form = EmailConfirmationForm.new
+      email_confirmation_form.generate_submission_references!
 
       render template: "forms/check_your_answers/show", locals: { email_confirmation_form: }
     end
@@ -29,7 +30,7 @@ module Forms
         end
       else
         setup_check_your_answers
-        email_confirmation_form.attributes = submission_reference_attributes
+        email_confirmation_form.generate_submission_references!
 
         render template: "forms/check_your_answers/show", locals: { email_confirmation_form: }, status: :unprocessable_entity
       end
@@ -39,14 +40,6 @@ module Forms
     end
 
   private
-
-    def submission_reference_attributes
-      reference = SecureRandom.uuid
-      {
-        confirmation_email_reference: "#{reference}-confirmation-email",
-        notify_reference: "#{reference}-submission-email",
-      }
-    end
 
     def page_to_row(page)
       question_name = helpers.question_text_with_optional_suffix_inc_mode(page, @mode)
--- a/app/controllers/forms/check_your_answers_controller.rb
+++ b/app/controllers/forms/check_your_answers_controller.rb
@@ -42,7 +42,10 @@ module Forms
 
     def submission_reference_attributes
       reference = SecureRandom.uuid
-      { notify_reference: "#{reference}-submission-email" }
+      {
+        confirmation_email_reference: "#{reference}-confirmation-email",
+        notify_reference: "#{reference}-submission-email",
+      }
     end
 
     def page_to_row(page)
@@ -63,7 +66,7 @@ module Forms
     end
 
     def email_confirmation_form_params
-      params.require(:email_confirmation_form).permit(:send_confirmation, :confirmation_email_address, :notify_reference)
+      params.require(:email_confirmation_form).permit(:send_confirmation, :confirmation_email_address, :confirmation_email_reference, :notify_reference)
     end
 
     def setup_check_your_answers
--- a/app/controllers/forms/check_your_answers_controller.rb
+++ b/app/controllers/forms/check_your_answers_controller.rb
@@ -4,7 +4,8 @@ module Forms
       return redirect_to form_page_path(current_context.form.id, current_context.form.form_slug, current_context.next_page_slug) unless current_context.can_visit?(CheckYourAnswersStep::CHECK_YOUR_ANSWERS_PAGE_SLUG)
 
       setup_check_your_answers
-      email_confirmation_form = EmailConfirmationForm.new(notify_reference: SecureRandom.uuid)
+      email_confirmation_form = EmailConfirmationForm.new(**submission_reference_attributes)
+
       render template: "forms/check_your_answers/show", locals: { email_confirmation_form: }
     end
 
@@ -28,7 +29,8 @@ module Forms
         end
       else
         setup_check_your_answers
-        email_confirmation_form.notify_reference = SecureRandom.uuid
+        email_confirmation_form.attributes = submission_reference_attributes
+
         render template: "forms/check_your_answers/show", locals: { email_confirmation_form: }, status: :unprocessable_entity
       end
     rescue StandardError => e
@@ -37,6 +39,11 @@ module Forms
     end
 
   private
+
+    def submission_reference_attributes
+      reference = SecureRandom.uuid
+      { notify_reference: "#{reference}-submission-email" }
+    end
 
     def page_to_row(page)
       question_name = helpers.question_text_with_optional_suffix_inc_mode(page, @mode)
--- a/app/form_objects/email_confirmation_form.rb
+++ b/app/form_objects/email_confirmation_form.rb
@@ -9,6 +9,11 @@ class EmailConfirmationForm
   validates :confirmation_email_address, presence: true, if: :validate_email?
   validates :confirmation_email_address, format: { with: URI::MailTo::EMAIL_REGEXP, message: :invalid_email }, allow_blank: true, if: :validate_email?
 
+  def initialize(...)
+    super(...)
+    generate_submission_references! unless @confirmation_email_reference || @notify_reference
+  end
+
   def validate_email?
     FeatureService.enabled?(:email_confirmations_enabled) && send_confirmation == "send_email"
   end
@@ -16,6 +21,8 @@ class EmailConfirmationForm
   def validate_presence?
     FeatureService.enabled?(:email_confirmations_enabled)
   end
+
+private
 
   def generate_submission_references!
     reference = SecureRandom.uuid
--- a/app/form_objects/email_confirmation_form.rb
+++ b/app/form_objects/email_confirmation_form.rb
@@ -2,7 +2,7 @@ class EmailConfirmationForm
   include ActiveModel::Model
   include ActiveModel::Validations
 
-  attr_accessor :send_confirmation, :confirmation_email_address, :notify_reference
+  attr_accessor :send_confirmation, :confirmation_email_address, :confirmation_email_reference, :notify_reference
 
   validates :send_confirmation, presence: true, if: :validate_presence?
   validates :send_confirmation, inclusion: { in: %w[send_email skip_confirmation] }, if: :validate_presence?
--- a/app/form_objects/email_confirmation_form.rb
+++ b/app/form_objects/email_confirmation_form.rb
@@ -16,4 +16,12 @@ class EmailConfirmationForm
   def validate_presence?
     FeatureService.enabled?(:email_confirmations_enabled)
   end
+
+  def generate_submission_references!
+    reference = SecureRandom.uuid
+    self.attributes = {
+      confirmation_email_reference: "#{reference}-confirmation-email",
+      notify_reference: "#{reference}-submission-email",
+    }
+  end
 end
--- a/app/views/forms/check_your_answers/show.html.erb
+++ b/app/views/forms/check_your_answers/show.html.erb
@@ -31,6 +31,8 @@
           <% end %>
           <%= form.govuk_radio_button :send_confirmation, 'skip_confirmation' %>
         <% end %>
+
+        <%= form.hidden_field :confirmation_email_reference, id: 'confirmation-email-reference' %>
       <% end %>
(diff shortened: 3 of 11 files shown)


jamie-o-wilkinson (dismissed):

app/views/forms/check_your_answers/show.html.erb
```diff
@@ -31,6 +31,8 @@
           <% end %>
           <%= form.govuk_radio_button :send_confirmation, 'skip_confirmation' %>
         <% end %>
+
+        <%= form.hidden_field :confirmation_email_reference, id: 'confirmation-email-reference' %>
```
aliuk2012, line 35: What happens with https://github.com/alphagov/forms-runner/pull/500/files#diff-54584ef745aa35987fb07628b3adc6a69531ecbfb5299bac906480b7d5777bbdR43?

app/controllers/forms/check_your_answers_controller.rb
```diff
@@ -38,6 +40,14 @@ def submit_answers
 
   private
 
+    def submission_reference_attributes
+      reference = SecureRandom.uuid
+      {
+        confirmation_email_reference: "#{reference}-confirmation-email",
+        notify_reference: "#{reference}-submission-email",
+      }
+    end
```
aliuk2012, line 49: I think this should be part of EmailConfirmationForm object not the responsibility of the controller.

app/views/forms/check_your_answers/show.html.erb
lfdebrux: I'm not sure I understand the question, could you elaborate?

aliuk2012: 🤦‍♂️ Sorry, I hadn't picked up that you had created a new attribute in the form object. it makes sense now

spec/requests/forms/check_your_answers_controller_spec.rb
```diff
@@ -86,6 +87,33 @@
     end
   end
 
+  shared_examples "for submission reference" do
+    uuid = /[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}/
+
+    let(:notify_reference) { assigns[:email_confirmation_form].notify_reference }
+    let(:confirmation_email_reference) { assigns[:email_confirmation_form].confirmation_email_reference }
```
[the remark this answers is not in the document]
aliuk2012: We could also add a test in the post request to check if the reference is included in the request

app/controllers/forms/check_your_answers_controller.rb
lfdebrux: I considered that initially, but I decided that it felt weird having that functionality in the form object, it seems like a different concern to handling user input. It's true that we're using the form to present the information in the view and preserve it between requests, but I think that's more of an implementation detail than a natural division of responsibilities.

I considered creating a new service or other kind of model object for this, but that seemed like unnecessary complexity for now. So I left the logic in the controller.

I think if this part of the code does grow, we could then look at refactoring, but for now having the code in the controller makes the most sense to me.

spec/requests/forms/check_your_answers_controller_spec.rb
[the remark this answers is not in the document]
lfdebrux: Using `assigns` seems to be the intended way to write requests specs [[1](https://rspec.info/features/6-0/rspec-rails/request-specs/request-spec/)]; all the examples use `assigns`, and they don't support the Capybara expectations that I'd normally use to test whether the reference is in the output.

That's why I ended up changing the controller code to use an instance variable rather than a local, because otherwise there was no good way of testing this code path.

app/controllers/forms/check_your_answers_controller.rb
```diff
@@ -4,15 +4,16 @@ def show
       return redirect_to form_page_path(current_context.form.id, current_context.form.form_slug, current_context.next_page_slug) unless current_context.can_visit?(CheckYourAnswersStep::CHECK_YOUR_ANSWERS_PAGE_SLUG)
 
       setup_check_your_answers
-      email_confirmation_form = EmailConfirmationForm.new(notify_reference: SecureRandom.uuid)
```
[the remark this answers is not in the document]
lfdebrux: See comment above https://github.com/alphagov/forms-runner/pull/500#discussion_r1399183908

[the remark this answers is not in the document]
aliuk2012: I think a good refactor would be to move the SecureRandom.uuid into the form object. its a small refactor and shows that the form object is in complete control over how the notification ref get set.

app/controllers/forms/check_your_answers_controller.rb
aliuk2012: I still think this is better to be placed inside the form object. The form fields attributes are controlled by the form object and pushed to the page as hidden form fields.  The form object in the traditional MVC structure would be the model. It would be responsible for handling the data and the business logic. Controller retrieves the data it needs from the model in order to respond to the request. The controller gives the retrieved data to the view. 
Also by moving this into the form object you wouldn't need https://github.com/alphagov/forms-runner/pull/500/files#diff-05a8792312bd570f7c19639941639759b413f0c20ef61c4a10856e28dddc2822R32 to to set the attributes because the form object would be responsible for it.

I agree 100% with you that creating a new service or other kind of model object for this would be complete overkill.

This controller is already doing far more than it needs to so I'd like to try not add more things to it and look to try simplify it instead.

lfdebrux: I've moved the code to generate submission references into the form object in commit [e53cfce](https://github.com/alphagov/forms-runner/pull/500/commits/e53cfcef2039390b7c1d66446f4fc613c7b4a30d). Please take a look and see what you think.

app/controllers/forms/check_your_answers_controller.rb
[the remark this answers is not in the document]
lfdebrux: See https://github.com/alphagov/forms-runner/pull/500#discussion_r1400160763.

spec/requests/forms/check_your_answers_controller_spec.rb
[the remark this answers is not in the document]
aliuk2012: The link you posted only mentions `assigns` twice in the examples and both were for expecting the response to redirect_to the instance (using rails magic to pick out the url) but most of the time it still ended up with a `expect(response.body).to include` type of test.

Also way back in 2016 `assigns` was soft depreciated to help rspec support Rails 5. There was a new gem created called `rails-controller-testing` which added back in `assigns` and a few other tests. `assigns` was originally used for controller spec but when request specs come along it was seen as an anti-pattern to use.

From https://rspec.info/blog/2016/07/rspec-3-5-has-been-released/ : Rails: Support for Rails 5
>In [Rails 5](https://github.com/rails/rails/issues/18950) assigns and assert_template are “soft deprecated”. Controller tests themselves are not, and adding :type => :controller to your specs is still 100% supported. Through Rails 3 and 4 it was both prevalent and idiomatic to use assigns in controller specs. As this is a minor release of RSpec our commitment to SemVer means that we are not going to break your existing controller specs. For existing Rails applications that make heavy use of assigns adding the [rails-controller-testing](https://github.com/rails/rails-controller-testing) to your Gemfile will restore assigns and assert_template. RSpec integrates with this gem seamlessly, so your controller specs should just continue to work.

For new Rails apps: we don’t recommend adding the rails-controller-testing gem to your application. The official recommendation of the Rails team and the RSpec core team is to write [request specs](https://www.relishapp.com/rspec/rspec-rails/docs/request-specs/request-spec) instead. Request specs allow you to focus on a single controller action, but unlike controller tests involve the router, the middleware stack, and both rack requests and responses. This adds realism to the test that you are writing, and helps avoid many of the issues that are common in controller specs. In Rails 5, request specs are significantly faster than either request or controller specs were in rails 4, thanks to the work by [Eileen Uchitelle](https://twitter.com/eileencodes?lang=en-gb)[^foot_1] of the Rails Committer Team.

Unfortunately, the blog hasn't been updated in a while and I haven't dug through all the various changelogs since but the only reason we can use `assign` is because we are pulling in `rails-controller-testing` gem.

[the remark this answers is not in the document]
lfdebrux: Okay... so if we're not going to use `assigns`, how can we test that `email_confirmation_form.generate_submission_reference!` is called? If we're making the submission references the responsibility of the form object, surely we don't want to test the behaviour of that in the request spec? If we did we'd just end up writing the same tests in two different places.

app/form_objects/email_confirmation_form.rb
```diff
@@ -16,4 +16,12 @@ def validate_email?
   def validate_presence?
     FeatureService.enabled?(:email_confirmations_enabled)
   end
+
+  def generate_submission_references!
```
aliuk2012, line 20:
```suggestion
	def initialize
	   generate_submission_references
	end
  
  private
  def generate_submission_references
```

I don't think we track the notify reference between page loads in this controller so it should be safe enough to generate a new reference every time.  I would make `generate_submission_references` a private method, move `initialize` method so its just after the validates section above in this file.

The benefit of doing this is just by calling `EmailConfirmationForm.new` should set the notify references and then the controller doesn't need to worry about calling it at a specific time like https://github.com/alphagov/forms-runner/pull/500/files#diff-05a8792312bd570f7c19639941639759b413f0c20ef61c4a10856e28dddc2822R8

Should be easy enough to test if you stub out SecureRandom and then call `EmailConfirmationForm.new` and expect the two notify references to be set to the values you used.

lfdebrux: What about when `submit_answers` happens? Will `EmailConfirmationForm.new(email_confirmation_form_params)` generate a new set of references? We need to use the same references as were generated for the `show` route.

aliuk2012: Good point, We could do something like 7e756e6c8bf300318aa3ea5ad75ff160b35c8191 or use a guard 1bd844a4ab0aa91733449e2cb29549bb82e2a2e6 so only generate those value if they are set when the object is initialised

lfdebrux: Ah, I forgot to mention, in the case where there is an error submitting answers, we then do want to generate a new set of references, even if we already have some. See  https://github.com/alphagov/forms-runner/blob/18c6550b170efca3283023893e907cbafbc09646/app/controllers/forms/check_your_answers_controller.rb#L29-L31

aliuk2012: TBH, I think the only reason this was originally added was because if there was validation error nothing would have generated the values again so the page would have rendered with empty fields not for any other particular reason.

I dont see there being any harm in removing this and playing back the same reference number if there is a validation error.

lfdebrux: > TBH, I think the only reason this was originally added was because if there was validation error nothing would have generated the values again so the page would have rendered with empty fields

That's not the behaviour I'm seeing... in the `main` branch the `email_confirmation_form` will get the value from the form data in the request, and if line 31 is commented out and the form is not valid the rendered form will include the same reference number.

@DavidBiddle it looks like you introduced this line in https://github.com/alphagov/forms-runner/commit/429d381c90b6d5e1cea2e28010c00cf80501079f, do you happen to remember the reason for it? I can't see anything mentioned in the commit message.

DavidBiddle: > @DavidBiddle it looks like you introduced this line in https://github.com/alphagov/forms-runner/commit/429d381c90b6d5e1cea2e28010c00cf80501079f, do you happen to remember the reason for it?

I don't think it was a particularly considered choice on my part - I think I just wasn't 100% sure whether we needed a new UUID or could keep the old one, so opted to generate a fresh one to be on the safe side.

spec/requests/forms/check_your_answers_controller_spec.rb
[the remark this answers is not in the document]
lfdebrux: @aliuk2012 commit [126bd66](https://github.com/alphagov/forms-runner/pull/500/commits/126bd66dfb046a502751a3310921ff178114b3af) makes this change.

app/form_objects/email_confirmation_form.rb
aliuk2012: for simplicity and consistency with how other form fields on the page work, we should keep the same generated uuid between page submits if there are validation errors.

spec/views/forms/check_your_answers/show.html.erb_spec.rb
```diff
@@ -73,6 +78,10 @@
     it "displays the email field" do
       expect(rendered).to have_field(I18n.t("helpers.label.email_confirmation_form.confirmation_email_address"))
     end
+
+    it "contains a hidden notify reference for the confirmation email" do
+      expect(rendered).to have_css("input", id: "confirmation-email-reference", visible: :hidden)
```
aliuk2012, line 83: Instead of only looking for the hidden field we should test it has the value we are expecting to see.
```suggestion
      expect(rendered).to have_field("confirmation-email-reference", type: "hidden", with: email_confirmation_form.confirmation_email_reference)
```

spec/views/forms/check_your_answers/show.html.erb_spec.rb
```diff
@@ -61,6 +62,10 @@
     it "does not display the email field" do
       expect(rendered).not_to have_field(I18n.t("helpers.label.email_confirmation_form.confirmation_email_address"))
     end
+
+    it "does not contain a hidden notify reference for the confirmation email" do
+      expect(rendered).not_to have_css("input", id: "confirmation-email-reference", visible: :hidden)
```
aliuk2012, line 67:
```suggestion
      expect(rendered).not_to have_field("confirmation-email-reference", type: "hidden", with: email_confirmation_form.confirmation_email_reference)
```

spec/requests/forms/check_your_answers_controller_spec.rb
```diff
@@ -339,6 +376,13 @@
         it "renders the check your answers page" do
           expect(response).to render_template("forms/check_your_answers/show")
         end
+
+        it "generates a new submission reference" do
+          expect(assigns[:email_confirmation_form].notify_reference).not_to eq email_confirmation_form[:notify_reference]
+          expect(assigns[:email_confirmation_form].confirmation_email_reference).not_to eq email_confirmation_form[:confirmation_email_reference]
+        end
```
aliuk2012, line 383: should this be checking `response.body`?

app/form_objects/email_confirmation_form.rb
lfdebrux: This has been changed in commit [98245df](https://github.com/alphagov/forms-runner/pull/500/commits/98245df3dfb4b9a9d60e1363e3e7de26d3a12301).

aliuk2012 (approved):
Nice work, sorry it took a little but longer but theres some really good refactoring in here new specs and functionality too 🙌

PR #508, opened 2023-11-24
Add notification banner for form previews

jamie-o-wilkinson:
For both live and draft previews, show a notification banner telling the user not to enter personal information. This is a mitigation for the use of trial accounts to collect information via form previews, even though they can't make forms live.

This is implemented using a separate `PreviewBannerComponent` rather than making changes to the `PreviewComponent` to keep them isolated from one another, but if people would prefer I'm open to merging the components into one.

Trello card: https://trello.com/c/6ZGD5RAE


thomasiles (approved):
nice tests!

PR #517, opened 2023-11-30
Add fill and submit form feature test

SamJamCul:

Trello card: https://trello.com/c/BP0bsQOf/1171-add-beginnings-of-missing-feature-tests-to-the-runner

This adds a basic feature test that covers filling in and submitting a form. 
There's also a change to our Axe check helper, which skips checking radio buttons for the `aria-allowed-attr` rule. This is a rule that probably shouldn't be catching our use of `aria-expanded="false"` as explored in [this thread](https://github.com/alphagov/govuk-frontend/issues/979).

I've opted to turn the email confirmation feature on in this test, as it's currently off by default. I kind of got most of the way through writing the tests before realising I'd turned it on locally, and it felt a bit pointless removing the opt-out-of-email-confirmation step in the test just because we haven't removed the feature flag yet.

The `ARIA_ALLOWED_ATTR` constant was added as sonarcloud thought it smelled funny otherwise.

### DIFF
--- a/spec/features/fill_in_and_submit_form_spec.rb
+++ b/spec/features/fill_in_and_submit_form_spec.rb
@@ -1,0 +1,79 @@
+require "rails_helper"
+
+feature "Fill in and submit a form", type: :feature, feature_email_confirmations_enabled: true do
+  let(:pages) { [(build :page, :with_text_settings, id: 1, form_id: 1, routing_conditions: [])] }
+  let(:form) { build :form, :live, id: 1, name: "Fill in this form", pages:, start_page: 1 }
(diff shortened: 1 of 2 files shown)


chao-xian:
@SamJamCul the commit message itself can have a link to the issue in govuk-frontend too - it's useful background

SamJamCul:
> @SamJamCul the commit message itself can have a link to the issue in govuk-frontend too - it's useful background

Good shout! I was a bit wary of including links in commit messages, but as it's a github issue it seems silly to avoid it!

DavidBiddle (approved):

PR #535, opened 2023-12-11
Update to govuk-frontend v5.0.0

DavidBiddle:

Trello card: https://trello.com/c/VeOkfpZv/1250-update-forms-runner-to-govuk-frontend-v5

Updates the govuk-frontend npm package and the govuk_components and govuk_design_system_formbuilder gems to v5.0.0.

Also includes several changes mandated or suggested by the upgrade guide - see individual commit messages for details. 

### Screenshots
Visually things are almost the same, except tags are now sentence-case rather than uppercase:
The header of a form, showing a pink tag reading 'Local' between the GOV.UK logo and the form name. Below this is a banner containing a blue 'Live preview' tag next to the text 'You're previewing the live version of this form.


thomasiles (approved):
Amazing! Really nicely structured and easy to follow.

looks good to me and works locally and in docker 🏅

app/views/layouts/base.html.erb
```diff
@@ -8,27 +8,19 @@
...
-    <link rel="apple-touch-icon" href="<%= vite_asset_path "#{govuk_assets_path}/images/govuk-apple-touch-icon-180x180.png"%>" sizes="180x180" />
-    <link rel="apple-touch-icon" href="<%= vite_asset_path "#{govuk_assets_path}/images/govuk-apple-touch-icon-167x167.png"%>" sizes="167x167" />
-    <link rel="apple-touch-icon" href="<%= vite_asset_path "#{govuk_assets_path}/images/govuk-apple-touch-icon-152x152.png"%>" sizes="152x152" />
-    <link rel="apple-touch-icon" href="<%= vite_asset_path "#{govuk_assets_path}/images/govuk-apple-touch-icon.png"%>" />
+    <link rel="icon" sizes="48x48" href="<%= vite_asset_path "#{govuk_assets_path}/images/favicon.ico" %>">
+    <link rel="icon" sizes="any" href="<%= vite_asset_path "#{govuk_assets_path}/images/favicon.svg" %>" type="image/svg+xml">
+    <link rel="mask-icon" href="<%= vite_asset_path "#{govuk_assets_path}/images/govuk-icon-mask.svg" %>" color="#0b0c0c">
+    <link rel="apple-touch-icon" href="<%= vite_asset_path "#{govuk_assets_path}/images/govuk-icon-180.png"%>">
+    <link rel="manifest" href="<%= vite_asset_path "#{govuk_assets_path}/manifest.json" %>">
```
lfdebrux, line 15: Does our asset pipeline do fingerprinting? If yes, we might need to generate our own manifest file (see https://github.com/alphagov/govuk-frontend/discussions/4389#discussioncomment-7697959).

DavidBiddle: Good spot - we do fingerprint so you're right, the manifest won't work. For now I've moved the manifest and the icons into the public folder to sidestep the issue, but I'd like to have a look at either copying them automatically as part of the build process, or generating the manifest (possibly with [vite-plugin-pwa](https://github.com/vite-pwa/vite-plugin-pwa)).

lfdebrux: Makes sense, thanks for looking into that.

Might be worth adding a `<!-- TODO -->` comment to this file to say something about the copies of the images?

DavidBiddle: Good suggestion, have updated

lfdebrux (approved):
Nice one!

PR #541, opened 2023-12-12
Fix visible production environment tag

DavidBiddle:

Trello card: https://trello.com/c/VeOkfpZv/1250-update-forms-runner-to-govuk-frontend-v5

In #535 we updated the tags to reflect the new tag casing rules in version 5 of govuk-frontend. Unfortunately, the 'production' text was not updated to 'Production' everywhere, so the logic for hiding the environment tag in production broke.

This PR fixes the problem, and updates the 'production' check to look at the translation directly, so that the environment check will stay up to date with any future changes to this text.

### DIFF
--- a/app/components/form_header_component/view.rb
+++ b/app/components/form_header_component/view.rb
@@ -26,7 +26,7 @@ module FormHeaderComponent
   private
 
     def service_name_with_tag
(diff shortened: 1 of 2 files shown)


DavidBiddle:
@chao-xian yes, we should - we do the same tag in forms-admin and possibly also in forms-product-page, I think it makes sense to do this as part of the govuk-frontend v5 update in those repos.

PR #543, opened 2023-12-13
Remove env var if logic for logger config

chao-xian:
This is now the default and doesn't need to rely on an env var.

See also https://github.com/alphagov/forms-deploy/pull/470


DavidBiddle (approved):

chao-xian:
This is what we should be doing instead: https://github.com/alphagov/forms-runner/pull/552

PR #578, opened 2024-01-08
Change header hompage link to forms-admin homepage when in preview mode

lfdebrux:

Trello card: https://trello.com/c/LtOJrgEO/1173-make-govuk-home-page-link-go-to-my-forms-in-preview 

### DIFF
--- a/app/components/form_header_component/view.rb
+++ b/app/components/form_header_component/view.rb
@@ -1,4 +1,6 @@
 module FormHeaderComponent
+  GOVUK_BASE_URL = "https://www.gov.uk/".freeze
+
   class View < ViewComponent::Base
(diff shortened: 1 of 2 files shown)


aliuk2012:
@lfdebrux  could you maybe add a preview of the component in the different states. I know theres no visual difference but then it means the automated accessibility specs we have will have another state of this component to test to ensure its accessible.

lfdebrux:
@aliuk2012 there's some previews in https://github.com/alphagov/forms-runner/blob/main/spec/components/form_header_component/form_header_component_preview.rb already, do they need to be added somewhere else as well, or is something else missing?

aliuk2012 (approved):
Looks really good and existing previews cover the existing functionality ass this change is only toggling the url that the user would be go to.

PR #585, opened 2024-01-15
Rename check your answers route

lfdebrux:

Our URL slugs use dashes `-` instead of underscores `_`, update the 'check your answers' slug to match.


chao-xian:
Why do we want to change this?

chao-xian:
Is there any effect on the e2e tests?

lfdebrux:
> Is there any effect on the e2e tests?

No effect, tests still pass.

aliuk2012 (approved):

PR #598, opened 2024-01-24
Logging refactor

danworth:
Makes it simpler to define common logging fields such as `form_id`, `session_id_hash` in a single place rather than the current four places which have been prone to divergence:
1. https://github.com/alphagov/forms-runner/blob/main/app/controllers/application_controller.rb#L28
2. https://github.com/alphagov/forms-runner/blob/main/config/application.rb#L69
3. https://github.com/alphagov/forms-runner/blob/main/app/lib/event_logger.rb#L9
4. https://github.com/alphagov/forms-runner/blob/main/app/lib/event_logger.rb#L23

These are now defined on a Controller instance var of `@logging_context` which is passed to services from the controller as necessary. This also reduces the need to pass multiple complex objects such as `context` and `request` down in order to construct these attributes lower down.

Uses the LogRage custom_payload approach which has access to the Controller's scope and can reference the instance var which removes the duplication of configuring fields in the controller and configuration (see points 1 and 2 above).

Extends the testing method introduced by @lfdebrux to check that the correct fields have been added to the log rage payload within the request specs. 

This should also make logging future events and exceptions simpler since it is only necessary to log something similar to `loggging_context.merge({ some_new_key: "something interesting" })`.

The `EventLogger` and its spec are only concerned with adding and checking the fields specific to the event logger, e.g. `event` rather than checking for common fields. This makes logging new fields in future simpler without having to open up existing classes and specs (the presence of the common fields are checked in the request specs, see above).

### Main Concern For Reviewing ###
I'm mostly worried about there being some unsafe referencing within the logging flow which could be encounter in edge cases and cause the app to return an error to the user. I was considering liberal use of `rescue` so that should anything fail in creating the log, we'd log something simpler and then continue to process the request. Open to suggestions on this. When reviewing please consider if you think there are any unsafe references etc.

## What's not included ##
I've not attempted to refactor everything associated with logging because this PR would be too hard to review. But some other things that come to mind for future.

I think we could simplify some for the services and logging classes further but reducing what we pass in, and sticking to either static or instance functions (e.g. https://github.com/alphagov/forms-runner/blob/main/app/lib/event_logger.rb#L20 and https://github.com/alphagov/forms-runner/blob/main/app/services/log_event_service.rb#L32). We also have occurrences of the controller directly calling [EventLogger](https://github.com/alphagov/forms-runner/blob/main/app/controllers/forms/page_controller.rb#L74) and the [LogEventService](https://github.com/alphagov/forms-runner/blob/main/app/controllers/forms/page_controller.rb#L17) which made refactoring more involved (rather than just having a common pattern , `Controller > Service > lower level thing`. 


config/application.rb
```diff
@@ -66,17 +66,8 @@ class Application < Rails::Application
...
-        h[:request_id] = event.payload[:request_id]
-        h[:form_id] = event.payload[:form_id] if event.payload[:form_id]
-        h[:page_id] = event.payload[:page_id] if event.payload[:page_id]
-        h[:page_slug] = event.payload[:page_slug] if event.payload[:page_slug]
-        h[:exception] = event.payload[:exception] if event.payload[:exception]
-        h[:session_id_hash] = event.payload[:session_id_hash] if event.payload[:session_id_hash]
-        h[:trace_id] = event.payload[:trace_id] if event.payload[:trace_id]
-      end
+    config.lograge.custom_payload do |controller|
+      controller.instance_variable_get(:@logging_context)
```
lfdebrux, line 70: Any reason we can't add a getter for `@logging_context` to `ApplicationController`?

app/controllers/forms/base_controller.rb
```diff
@@ -26,6 +31,10 @@ def current_context
       @current_context ||= Context.new(form: current_form, store: session)
     end
 
+    def logging_context
+      @logging_context ||= set_logging_context
+    end
+
```
lfdebrux, line 37: I'd suggest moving this to `ApplicationController` and changing it to initialise the variable with an empty dict:

```
def logging_context
  @logging_context ||= {}
end
```

then we can use the accessor method everywhere, including in `set_logging_context`, instead of the instance var.

app/lib/event_logger.rb
```diff
@@ -3,36 +3,15 @@ def self.log(object)
...
-      method: request&.method,
-      request_id: request&.request_id,
-      session_id_hash:,
-      form: context.form.name,
-      form_id: context.form.id,
-      event: "form_#{event}",
-    })
+  def self.log_form_event(logging_context, event)
+    log(logging_context.merge({ event: "form_#{event}" }))
   end
```
lfdebrux, line 8: Love it! 🙌

app/controllers/application_controller.rb
```diff
@@ -25,15 +26,16 @@ def check_maintenance_mode_is_enabled
...
-    payload[:trace_id] = request.env["HTTP_X_AMZN_TRACE_ID"].presence
+  def set_logging_context
+    @logging_context = {}.tap do |h|
+      h[:host] = request.host
+      h[:request_id] = request.request_id
+      h[:form_id] = params[:form_id] if params[:form_id].present?
+      h[:page_id] = params[:page_slug] if params[:page_slug].present? && params[:page_slug].match(Page::PAGE_ID_REGEX)
+      h[:page_slug] = params[:page_slug] if params[:page_slug].present?
+      h[:session_id_hash] = session_id_hash
+      h[:trace_id] = request.env["HTTP_X_AMZN_TRACE_ID"].presence
```
lfdebrux, line 37: I kind of mangled this line when I added it, it probably should be

```suggestion
      h[:trace_id] = request.env["HTTP_X_AMZN_TRACE_ID"] if request.env["HTTP_X_AMZN_TRACE_ID"].present?
```

config/application.rb
danworth: Only that I couldn't get it to work originally but I've tried harder and I've got it working (the current healtcheck controller doesn't inherit from Application::Base which caused a slight wrinkle, interesting `controller&.logging_context` wouldn't safely reference it but `controller.try(:logging_context)` does ). 👍

app/controllers/forms/base_controller.rb
danworth: Yeah good suggestion and accepted 👍

app/controllers/application_controller.rb
danworth: yeah I think as it was it would have just set the `trace_id:` to `nil` but I've changed it as you suggest to omit the field if its not available instead.

app/lib/event_logger.rb
danworth: yeah I think it's much simpler now.

app/services/log_event_service.rb
```diff
@@ -7,18 +7,18 @@ def initialize(current_context, step, request, changing_answer, answers)
...
   end
 
-  def self.log_form_start(context, request)
-    EventLogger.log_form_event(context, request, "visit") # Logging to Splunk
+  def self.log_form_start(logging_context)
+    EventLogger.log_form_event(logging_context, "visit") # Logging to Splunk
   end
 
-  def self.log_submit(context, request, requested_email_confirmation: false, preview: false)
+  def self.log_submit(logging_context, context, requested_email_confirmation: false, preview: false)
```
DavidBiddle, line 14: Do we need to supply both `logging_context` and `context`? I think `context` is only used in the  `CloudWatchService.log_form_submission(form_id: context.form.id)` call on line 25, which I reckon could be replaced with `CloudWatchService.log_form_submission(form_id: logging_context.form_id)`

DavidBiddle (approved):
Code looks good and looks like the event logging still works on my machine! 🎉

PR #606, opened 2024-01-30
Refactor submission error exception handling

lfdebrux:

While looking at how to improve the amount of information we log related to form submission emails, I noticed a few things that could be improved in our current codebase.

This PR refactors the `submit_answers` action to use a specific exception class `SubmissionError`, rather than the generic `StandardError`.

See the commit messages for more details.

### DIFF
--- a/app/services/form_submission_service.rb
+++ b/app/services/form_submission_service.rb
@@ -1,4 +1,6 @@
 class FormSubmissionService
+  class SubmissionError < StandardError; end
+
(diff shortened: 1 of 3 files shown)


lfdebrux:
I've gone off the idea, `rescue StandardError` _is_ a code smell, but this change just pushes the smell lower down the stack. And anyway, in this case `rescue StandardError` is the correct behaviour.

PR #607, opened 2024-01-30
Refactor submission error tests

lfdebrux:

While looking at how to improve the amount of information we log related to form submission emails, I noticed a few things that could be improved in our current codebase.

This PR fixes up a few tests to be more robust.

I've raised this a standalone PR, so that it can be reviewed and merged quickly without getting into changes to implementation.


DavidBiddle (approved):
Code looks good, and good spot on the second issue especially!

PR #609, opened 2024-01-31
Refactor tests of controller logging context

lfdebrux:

We're using the same bit of code to subscribe to `process_action` notifications and capture the payload in a couple of different tests now, this commit creates a shared helper to reduce duplication.

### DIFF
--- a/spec/support/action_controller_notifications_helpers.rb
+++ b/spec/support/action_controller_notifications_helpers.rb
@@ -1,6 +1,12 @@
+# rubocop:disable Style/ClassVars
+
 module ActionControllerNotificationsHelpers
   @@payloads = []
   @@process_action_subscriber = nil
+
+  def logging_context
+    payload[:custom_payload]
+  end
 
   def payloads
     @@payloads ||= []
@@ -24,6 +30,8 @@ module ActionControllerNotificationsHelpers
     process_action_subscriber
   end
 end
+
+# rubocop:enable Style/ClassVars
 
 RSpec.configure do |config|
   config.include ActionControllerNotificationsHelpers, type: :request
--- a/spec/support/action_controller_notifications_helpers.rb
+++ b/spec/support/action_controller_notifications_helpers.rb
@@ -1,0 +1,38 @@
+module ActionControllerNotificationsHelpers
+  @@payloads = []
+  @@process_action_subscriber = nil
+
+  def payloads
+    @@payloads ||= []
+  end
+
+  def payload
+    payloads.last
+  end
+
+  def process_action_subscriber
+    @@process_action_subscriber ||= ActiveSupport::Notifications.subscribe("process_action.action_controller") do |_, _, _, _, payload|
+      payloads << payload
+    end
+  end
+
+  def reset_action_controller_notifications
+    payloads.clear
+  end
+
+  def subscribe_to_action_controller_notifications
+    process_action_subscriber
+  end
+end
+
+RSpec.configure do |config|
+  config.include ActionControllerNotificationsHelpers, type: :request
+
+  config.before(:example, type: :request) do
+    subscribe_to_action_controller_notifications
+  end
+
+  config.after(:example, type: :request) do
+    reset_action_controller_notifications
+  end
+end
--- a/spec/requests/application_controller_spec.rb
+++ b/spec/requests/application_controller_spec.rb
@@ -18,21 +18,8 @@ RSpec.describe ApplicationController, type: :request do
   end
 
   context "when setting logging context" do
-    let(:payloads) { [] }
-    let(:payload) { payloads.last }
-
-    let!(:subscriber) do
-      ActiveSupport::Notifications.subscribe("process_action.action_controller") do |_, _, _, _, payload|
-        payloads << payload
-      end
-    end
-
     before do
       get root_path, headers: { "HTTP_X_AMZN_TRACE_ID": "Root=1-63441c4a-abcdef012345678912345678" }
-    end
-
-    after do
-      ActiveSupport::Notifications.unsubscribe(subscriber)
     end
 
     it "adds the trace ID to the instrumentation payload" do
--- a/spec/requests/application_controller_spec.rb
+++ b/spec/requests/application_controller_spec.rb
@@ -22,20 +22,20 @@ RSpec.describe ApplicationController, type: :request do
       get root_path, headers: { "HTTP_X_AMZN_TRACE_ID": "Root=1-63441c4a-abcdef012345678912345678" }
     end
 
-    it "adds the trace ID to the instrumentation payload" do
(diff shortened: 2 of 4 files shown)


spec/support/action_controller_notifications_helpers.rb
```diff
@@ -0,0 +1,46 @@
+# rubocop:disable Style/ClassVars
```
danworth, line 1: I was curious about this so I set the class vars back to class instance vars and all the tests still pass locally. What error were you seeing?

lfdebrux: That's weird, I got these errors:

```
`payloads` is not available on an example group (e.g. a `describe` or `context` block). It is only available from within individual examples (e.g. `it` blocks) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc).

`process_action_subscriber` is not available on an example group (e.g. a `describe` or `context` block). It is only available from within individual examples (e.g. `it` blocks) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc).
```

What changes did you make?

danworth: I changed `@@` to `@` and removed the rubocop disable.

lfdebrux: @danworth ah, that's a plain instance variable, rather than a _class_ instance variable [[1](https://www.ruby-lang.org/en/documentation/faq/8/)]. I'm not sure why Ruby has two different kinds of class variable, but it does 😅.

I wanted the `subscriber` to be global, so that we don't need to subscribe/unsubscribe for each example (apparently there is a performance penalty when doing so [[2](https://api.rubyonrails.org/classes/ActiveSupport/Notifications.html#module-ActiveSupport::Notifications-label-Temporary+Subscriptions)]. Having the `subscriber` as some kind of class variable seemed the obvious way to go to do that.

But if I make the recommended changes to make it into a class instance variable:

```
diff --git a/spec/support/action_controller_notifications_helpers.rb b/spec/support/action_controller_notifications_helpers.rb
index df63c77..7c0ad4d 100644
--- a/spec/support/action_controller_notifications_helpers.rb
+++ b/spec/support/action_controller_notifications_helpers.rb
@@ -1,15 +1,17 @@
-# rubocop:disable Style/ClassVars
-
 module ActionControllerNotificationsHelpers
-  @@payloads = []
-  @@process_action_subscriber = nil
+  @payloads = []
+  @process_action_subscriber = nil
+
+  class << self
+    attr_accessor :payloads, :process_action_subscriber
+  end
 
   def logging_context
     payload[:custom_payload]
   end
 
   def payloads
-    @@payloads ||= []
+    self.class.payloads ||= []
   end
 
   def payload
@@ -17,7 +19,8 @@ module ActionControllerNotificationsHelpers
   end
 
   def process_action_subscriber
-    @@process_action_subscriber ||= ActiveSupport::Notifications.subscribe("process_action.action_controller") do |_, _, _, _, payload|
+    puts "oh no" unless self.class.process_action_subscriber
+    self.class.process_action_subscriber ||= ActiveSupport::Notifications.subscribe("process_action.action_controller") do |_, _, _, _, payload|
       payloads << payload
     end
   end
@@ -31,8 +34,6 @@ module ActionControllerNotificationsHelpers
   end
 end
 
-# rubocop:enable Style/ClassVars
-
 RSpec.configure do |config|
   config.include ActionControllerNotificationsHelpers, type: :request

```

then I get the error listed above.

danworth: So my curiosity was to explore `@@` class variables vs `@` class instance variables because that's the rubocop warning and advice in the Ruby docs (generally it says you don't want to use `@@` class vars because they change if any class in the inheritance tree changes it and you probably want to use `@` instead) https://www.ruby-lang.org/en/documentation/faq/8/#:~:text=What%20is%20the%20difference%20between,belong%20to%20one%20specific%20class.

I changed what you had to make `@@` class variables to `@` class instance variables and it worked in that the tests passed. I don't yet understand Ruby exceptionally well but I mentioned it because the code you had worked with making them class instance variables and therefore didn't need the rubocop disable. Is there a reason not to make them class instance variables (per git diff beneath). 

I'm out tomorrow so don't let this conversation hold up approval and merging though I am interested in it.

<details>
<summary>Git diff</summary>

```

diff --git a/spec/support/action_controller_notifications_helpers.rb b/spec/support/action_controller_notifications_helpers.rb
index df63c77..f63ecff 100644
--- a/spec/support/action_controller_notifications_helpers.rb
+++ b/spec/support/action_controller_notifications_helpers.rb
@@ -1,15 +1,13 @@
-# rubocop:disable Style/ClassVars
-
 module ActionControllerNotificationsHelpers
-  @@payloads = []
-  @@process_action_subscriber = nil
+  @payloads = []
+  @process_action_subscriber = nil
 
   def logging_context
     payload[:custom_payload]
   end
 
   def payloads
-    @@payloads ||= []
+    @payloads ||= []
   end
 
   def payload
@@ -17,7 +15,7 @@ module ActionControllerNotificationsHelpers
   end
 
   def process_action_subscriber
-    @@process_action_subscriber ||= ActiveSupport::Notifications.subscribe("process_action.action_controller") do |_, _, _, _, payload|
+    @process_action_subscriber ||= ActiveSupport::Notifications.subscribe("process_action.action_controller") do |_, _, _, _, payload|
       payloads << payload
     end
   end
@@ -31,8 +29,6 @@ module ActionControllerNotificationsHelpers
   end
 end
 
-# rubocop:enable Style/ClassVars
-
 RSpec.configure do |config|
   config.include ActionControllerNotificationsHelpers, type: :request
 
```

</details>

lfdebrux: @danworth My understanding is that the code you shared has both class instance variables and instance variables, but the two don't connect. I tested this with a little code snippet:

```
# instance_variables.rb
class A
  @a = "a"

  def self.print_a
    puts @a.inspect
  end

  def print_a
    puts @a.inspect
  end
end

A.print_a      # => nil
A.new.print_a  # => nil
```

Running this with `ruby instance_variables.rb` prints `"a"` and then `nil`, showing that the scope of the declaration on line 2 is limited to the class object and not instances of the class.

danworth: Let's catch up now you're back (I also have a related change I need to make to the logging of the question number). The code I have modified with the diff above has only class instance vars within `ActionControllerNotificationsHelpers` and the tests run and pass.

danworth: After chatting I now understand the problem, with the changes in my diff it would be subscribing each invocation since the `@process_action_subscriber` is nil each time according to Laurence's investigation.

danworth (approved):
Looks good 🏅

PR #610, opened 2024-01-31
Refactor Dockerfile

lfdebrux:

This PR makes a few tweaks to the Dockerfile to make it easier to use with Docker Compose and the `update_app_versions` script in forms-deploy.

The changes here should be the same as in alphagov/forms-admin#905.


DavidBiddle (approved):
Code looks good to me and builds on my local machine 🎉

PR #622, opened 2024-02-12
Log Question_Number

danworth:
The `question_number` field was omitted during the recent refactor to logging[[1]]. This adds it back in. Note that `question_number` is the sequential position of the question with reference to the form (e.g. the first or second question in the form) and the `page_id` is the unique identifier of the page in our records.

[1]: https://github.com/alphagov/forms-runner/pull/598

Puts question_number back into the logs to help with analytics.

I have run this locally and seen that `question_number` appears as expected.


thomasiles (approved):
neat 🏅

PR #625, opened 2024-02-13
Add exception logging to submit your answers action

lfdebrux:

We want to log exceptions that happen in the `submit_answers` action even if we rescue them.

Unfortunately I couldn't find an easy way to hook into the standard Rails exception logging mechanisms at the same time as using `rescue` or `rescue_from`; so this commit adds a new method to our application controller that has the same behaviour.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -42,6 +42,33 @@ class ApplicationController < ActionController::Base
     end
   end
 
+  def log_rescued_exception(exception)
+    # add exception to payload with same format used by ActiveSupport::Notifications::Instrumenter
+    logging_context[:rescued_exception] = [exception.class.name, exception.message]
+
+    # log exception with the same format used by ActionDispatch::DebugExceptions
+    backtrace_cleaner = request.get_header("action_dispatch.backtrace_cleaner")
+    wrapper = ActionDispatch::ExceptionWrapper.new(backtrace_cleaner, exception)
+    trace = wrapper.exception_trace
+    message = []
+    message << "Rescued exception:"
+    message << "  "
+    message << "#{wrapper.exception_class_name} (#{wrapper.message}):"
+    if wrapper.has_cause?
+      message << "\nCauses:"
+      wrapper.wrapped_causes.each do |wrapped_cause|
+        message << "#{wrapped_cause.exception_class_name} (#{wrapped_cause.message})"
+      end
+    end
+    message.concat(wrapper.annotated_source_code)
+    message << "  "
+    message.concat(trace)
+    logger.warn message.join("\n")
+
+    # send exception to Sentry
+    Sentry.capture_exception(exception)
+  end
+
 private
 
   def add_robots_header
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -65,6 +65,9 @@ class ApplicationController < ActionController::Base
     message.concat(trace)
     logger.warn message.join("\n")
 
+    # add additional info to request log
+    logging_context[:rescued_exception_trace] = message
+
     # send exception to Sentry
     Sentry.capture_exception(exception)
   end
--- a/app/controllers/forms/check_your_answers_controller.rb
+++ b/app/controllers/forms/check_your_answers_controller.rb
@@ -31,7 +31,8 @@ module Forms
         render template: "forms/check_your_answers/show", locals: { email_confirmation_form: }, status: :unprocessable_entity
       end
     rescue StandardError => e
-      Sentry.capture_exception(e)
+      log_rescued_exception(e)
+
       render "errors/submission_error", status: :internal_server_error
     end
 
--- a/spec/controllers/application_controller_spec.rb
+++ b/spec/controllers/application_controller_spec.rb
@@ -36,6 +36,25 @@ RSpec.describe ApplicationController do
 
       expect(controller.logging_context).to include(rescued_exception: ["RuntimeError", "oh no\nwhat happened"])
     end
+
+    it "adds exception trace to logging context" do
+      get :raise_error
+
+      expect(controller.logging_context).to include(:rescued_exception_trace)
+      expect(controller.logging_context[:rescued_exception_trace]).to include(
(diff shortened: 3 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -42,6 +42,33 @@ def set_logging_context
...
+    if wrapper.has_cause?
+      message << "\nCauses:"
+      wrapper.wrapped_causes.each do |wrapped_cause|
+        message << "#{wrapped_cause.exception_class_name} (#{wrapped_cause.message})"
+      end
+    end
+    message.concat(wrapper.annotated_source_code)
+    message << "  "
+    message.concat(trace)
+    logger.warn message.join("\n")
```
danworth, line 66: Nice change. Why have you chosen to log this independently (if I've understood this correctly) to the logging_context/LogRage line which has had the `rescued_exception` added, is it to mimic what Rails would have done had the exception not been caught? I think the cleaned back trace would be useful context for the logging_context since the exception's class name and message may not be enough to identify where the exception was thrown within the method?

lfdebrux: Yeah, it's to mimic what Rails would have done. I definitely felt like the exception traceback would be useful, but wasn't sure about putting it in the logging context, because if felt weird treating rescued exceptions differently to other exceptions.

I guess ideally we would have structured logging of all exceptions, apparently that was on the roadmap for Rails 7 (https://discuss.rubyonrails.org/t/support-to-easily-override-actiondispatch-debugexceptions-log-format/78225/2) but obviously it didn't make it in. Fingers crossed maybe it will be in Rails 8?

danworth: Yeah I agree the traceback would likely be very useful. If we keep it separate I presume the only way to find it and link to the request would be searching for exception class name/message since they'd be common which could be a bit of pain. I would favour making things as simple for us as possible which to me would be adding the traceback to the logging_context `:rescued_exception_trace` for example, I don't see a harm also logging it independently for consistency (especially because these are exceptions raised in exceptional circumstances and not like we'd be flooding our logs).

lfdebrux: Yeah, that's a fair point about linking the error message with the request. I'd like it if we make it so that every log message has the request ID attached; if that was the case would you still want the trace with the logging context?

danworth: Yes, for the reason of making things as simple as we can for ourselves when things go wrong (the trace would provide useful context beyond the class name and message alone I suspect and it's simpler than having to join things together). We'd only be adding it in exceptional circumstances when exceptions occur so its not like we'll be bloating every LogRage log line. If you want to log it independently as well then fine by me. As always we can see how it plays out and revisit if we think its not working for us. Perhaps the structure logging in rails 8 will solve much of this 🤞 

Do you have any inclination to the potential to accidentally log PII, the backtrace seems to just be log lines right, rather than throwing in any request details etc...

lfdebrux: Cool, okay, I'll add the trace to the context. Is there any other information you'd want in the logging context?

TBH I don't often find the plain trace useful, Sentry gives us a nicely formatted trace that's much easier to use, but I think if we add the annotated source code and causes that is in the Rails error message I would personally find that useful.

> Do you have any inclination to the potential to accidentally log PII, the backtrace seems to just be log lines right, rather than throwing in any request details etc...

Yeah, the backtrace is just line numbers in the application source code, it shouldn't include any user data. I thought we were okay were logging request details/PII to Splunk though?

danworth: >I would personally find that useful.

Yeah use that as your guide, these are after all messages to ourselves in future when something goes wrong. Add the annotated code and causes rather than trace, if you think it'll be useful then please do it (by `trace` I meant the whole message that was going to be logged independently, stick it in `:rescued_exception_<something>`, sorry if I wasn't clear.)

>Is there any other information you'd want in the logging context?

Let's see how it goes, logging this extra information is an improvement itself and we can iterate it. The ideal is that when something goes wrong its simple for us to understand what has gone wrong and the potential impact of this without a lot of diving in the code base for line numbers and trying to infer what did or didn't happen especially during an incident etc. I think we could do more to help with understanding the consequences, and I've seen patterns like logging `Failed to send submission email: <Exception Class>, <Exception Message>` or `Failed to send submission confirmation email: <Class>, <Message>`. Since we have a single `rescue` for the entire `submit_answers` function which is doing a few different things (like sending confirmation email, submission email, updating CloudWatch metrics etc) we can't make such specific statements as it currently stands (I think). Armed with the trace/line number, the exception class and message we can most likely work it out but something to consider in future.

Overall I'd like us to continue to invest time up front to make it simple and obvious to ourselves when things go wrong.

lfdebrux: Cool, I've added another commit for this, see https://github.com/alphagov/forms-runner/pull/625/commits/2f9fa877765fe48b9da9651ee3a06d428467b464.

danworth (approved):
Looks good to me and I think sets us up in the right direction to make things simpler to understand when things go wrong. I ran this locally and forced an exception, the log looks helpful to me:
```
{
  "level": "INFO",
  "time": "2024-02-13T14:29:14.374+00:00",
  "method": "POST",
  "path": "/preview-live/2/laurence-s-birthday-form/check-your-answers",
  "format": "html",
  "controller": "Forms::CheckYourAnswersController",
  "action": "submit_answers",
  "status": 500,
  "allocations": 10986,
  "duration": 95.32,
  "view": 40.42,
  "host": "localhost",
  "request_id": "ace05f3c-7dc9-47d7-91a9-ee3198ec67ff",
  "form_id": "2",
  "session_id_hash": "44939adefc6f3cbfff887b8db664ff593611f7a13b0b2d60daca5e166d7c3569",
  "form_name": "Laurence's Birthday Form",
  "rescued_exception": [
    "RuntimeError",
    "Testing this out"
  ],
  "rescued_exception_trace": [
    "Rescued exception:",
    "  ",
    "RuntimeError (Testing this out):",
    "  ",
    "app/controllers/forms/check_your_answers_controller.rb:13:in `submit_answers'"
  ]
}
```

PR #627, opened 2024-02-14
Add notification references to logging context for form submission

lfdebrux:

We want to be able to correlate logs in GOV.UK Notify with logs in our system; to do this we can use the notification references.

This commit adds the notification references for submission emails and confirmation emails to the log messages for the `submit_answers` action.

### DIFF
--- a/spec/requests/forms/check_your_answers_controller_spec.rb
+++ b/spec/requests/forms/check_your_answers_controller_spec.rb
@@ -87,26 +87,26 @@ RSpec.describe Forms::CheckYourAnswersController, type: :request do
     end
   end
 
-  shared_examples "for submission reference" do
-    prepend_before do
-      allow(EmailConfirmationForm).to receive(:new).and_wrap_original do |original_method, *args|
-        double = original_method.call(*args)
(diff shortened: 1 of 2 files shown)


spec/requests/forms/check_your_answers_controller_spec.rb
```diff
@@ -370,6 +385,8 @@
       it "redirects to confirmation page" do
         expect(response).to redirect_to(form_submitted_path)
       end
+
+      include_examples "for submission references"
```
danworth, line 389: I think it could cause some confusion `when the user has not requested a confirmation email` to still add the confirmation email reference to the logging_context (which I presume is being expected here in the common example).

```confirmation_email_reference: email_confirmation_form[:confirmation_email_reference],```

DavidBiddle (dismissed):

spec/requests/forms/check_your_answers_controller_spec.rb
lfdebrux: Good idea, done!

danworth (approved):
Looks good 🏅

PR #634, opened 2024-02-16
Rename submission email reference attribute

lfdebrux:

Rename `notify_reference` to `submission_email_reference`.

This better describes the attribute, and disambiguates it from the confirmation email reference.

Because this change affects the end to end tests, we use the 'expand and contract' pattern to keep the old name around where necessary. There is also an associated change in alphagov/forms-e2e-tests#41 to rename the reference ID. That PR can be merged after this PR has been deployed.

### DIFF
--- a/app/controllers/forms/check_your_answers_controller.rb
+++ b/app/controllers/forms/check_your_answers_controller.rb
@@ -56,7 +56,7 @@ module Forms
     end
 
     def email_confirmation_form_params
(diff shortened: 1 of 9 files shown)


DavidBiddle (approved):
Nice change, looks much clearer! Expand and contract approach makes sense to me.

Tested locally - emails still work and I can see the two hidden fields 🎉

PR #636, opened 2024-02-19
Add attributes to the confirmation_email input

thomasiles:
# From the [trello ticket](https://trello.com/c/XpmRK8zg/1382-email-confirmation-form-field-is-a-plain-text-field):

## WCAG context

This fails WCAG 1.3.5 Identify Input Purpose, because this field should have an autocomplete=”email” attribute. The other aspects - type=”email” and spellcheck=”false” - aren’t WCAG failures, but they’ll help users enter their email addresses more easily, and it makes sense to implement them at the same time.

The field should have the following attributes:

- `type=”email”`
- `autocomplete=”email”`
- `spellcheck=”false”`

This commit adds a test to check the attributes are present and changes the code to include them.

### DIFF
--- a/app/views/forms/check_your_answers/show.html.erb
+++ b/app/views/forms/check_your_answers/show.html.erb
@@ -26,7 +26,7 @@
 
       <%= form.govuk_radio_buttons_fieldset(:send_confirmation, legend: { size: 'm', tag: 'h2' }) do %>
         <%= form.govuk_radio_button :send_confirmation, 'send_email' do %>
-          <%= form.govuk_text_field :confirmation_email_address %>
+          <%= form.govuk_text_field :confirmation_email_address, type: 'email', autocomplete: 'email', spellcheck: false  %>
         <% end %>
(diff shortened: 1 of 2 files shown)


app/views/forms/check_your_answers/show.html.erb
```diff
@@ -26,7 +26,7 @@
 
       <%= form.govuk_radio_buttons_fieldset(:send_confirmation, legend: { size: 'm', tag: 'h2' }) do %>
         <%= form.govuk_radio_button :send_confirmation, 'send_email' do %>
-          <%= form.govuk_text_field :confirmation_email_address %>
+          <%= form.govuk_text_field :confirmation_email_address, type: 'email', autocomplete: 'email', spellcheck: false  %>
```
DavidBiddle, line 29: You can use `govuk_email_field` to avoid having to set `type` to `email` manually
```suggestion
          <%= form.govuk_email_field :confirmation_email_address, autocomplete: 'email', spellcheck: false  %>
```

DavidBiddle (approved):
Looks good, thanks for bringing this one in!

PR #646, opened 2024-02-22
Add reference number to confirmation screen

DavidBiddle:

Trello card: https://trello.com/c/44ZcfuuB/1360-generate-a-reference-for-form-submissions

This PR generates a reference number and adds it to the confirmation page.

To do this we:
- generate an 8 digit alphanumeric reference number and store it in a special bit of the session which doesn't get cleared on submit
- read the stored reference number on the confirmation page
- move the `email_sent` flag into the session, and rename it to make it clearer
- add the submission reference to the logging context
- clear the submission details from the session when the user restarts the form and saves the first question.
- feature flag the logging and display of the reference, so that we can turn it on after the email work is complete.

There's additional context on each commit message.

#### To be done in another PR
- pass the reference number into Notify and display it in the submission and confirmation emails ([trello #1361](https://trello.com/c/dm8jKunt/1361-generate-a-reference-for-form-submissions-email-work))

#### Other things we might want to do later
These are some loosely related things we might want to do at some point, but which I don't think are part of this reference work. 
- at some point we might want to edit FormContext's 'form_submitted?' method to check for the presence of the submission reference rather than an empty session. I think this would make sense because there are other possible ways to empty the answers store, but you can only get a submission reference by submitting the form.
- it might make sense to redirect from the submitted page back to an earlier step in the form if the user doesn't have a confirmation reference. We do this for every other page, and I think this work makes it possible to do it for the confirmation screen too.

### Screenshots
A confirmation page showing a green GOV.UK panel component with the heading 'Your form has been submitted' and the text 'Your reference number BQJMNQNW'. Undernathe the panel component is a paragraph reading 'We have sent you a confirmation email', a small heading 'What happens next', and a paragraph saying 'We aim to respond to all applications within 5 working days, but this may vary at busy times. Our service status page shows our current expected waiting times (opens in new tab)'

### DIFF
--- a/app/controllers/forms/check_your_answers_controller.rb
+++ b/app/controllers/forms/check_your_answers_controller.rb
@@ -17,11 +17,13 @@ module Forms
         if current_context.form_submitted?
           redirect_to error_repeat_submission_path(current_form.id)
         else
-          FormSubmissionService.call(logging_context:,
-                                     current_context:,
-                                     request:,
-                                     email_confirmation_form:,
-                                     preview_mode: mode.preview?).submit
+          submission_reference = FormSubmissionService.call(logging_context:,
+                                                            current_context:,
+                                                            request:,
+                                                            email_confirmation_form:,
+                                                            preview_mode: mode.preview?).submit
+
+          current_context.save_submission_reference(submission_reference)
 
           redirect_to :form_submitted, email_sent: requested_email_confirmation
         end
--- a/app/controllers/forms/check_your_answers_controller.rb
+++ b/app/controllers/forms/check_your_answers_controller.rb
@@ -23,9 +23,9 @@ module Forms
                                                             email_confirmation_form:,
                                                             preview_mode: mode.preview?).submit
 
-          current_context.save_submission_reference(submission_reference)
+          current_context.save_submission_details(submission_reference, requested_email_confirmation)
 
-          redirect_to :form_submitted, email_sent: requested_email_confirmation
+          redirect_to :form_submitted
         end
       else
         setup_check_your_answers
--- a/app/views/forms/submitted/submitted.html.erb
+++ b/app/views/forms/submitted/submitted.html.erb
@@ -3,7 +3,7 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <%= govuk_panel(title_text: t('form.submitted.title'), classes: ["govuk-!-margin-bottom-7"], ) do %>
-      <% if @current_context.get_submission_reference.present? %>
+      <% if FeatureService.enabled?(:reference_numbers_enabled) && @current_context.get_submission_reference.present? %>
         <%= t('form.submitted.your_reference') %><br>
         <strong><%= @current_context.get_submission_reference %></strong>
       <% end %>
--- a/app/views/forms/submitted/submitted.html.erb
+++ b/app/views/forms/submitted/submitted.html.erb
@@ -9,7 +9,7 @@
       <% end %>
     <% end %>
 
-    <% if email_sent %>
+    <% if @current_context.email_sent? %>
       <p><%= t('form.submitted.email_sent') %></p>
     <% end %>
 
--- a/app/views/forms/submitted/submitted.html.erb
+++ b/app/views/forms/submitted/submitted.html.erb
@@ -9,7 +9,7 @@
       <% end %>
     <% end %>
 
-    <% if @current_context.email_sent? %>
+    <% if @current_context.requested_email_confirmation? %>
       <p><%= t('form.submitted.email_sent') %></p>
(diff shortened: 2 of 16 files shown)


app/controllers/forms/check_your_answers_controller.rb
```diff
@@ -17,13 +17,19 @@ def submit_answers
         if current_context.form_submitted?
           redirect_to error_repeat_submission_path(current_form.id)
         else
+          submission_id = SecureRandom.alphanumeric(8).upcase
```
SamJamCul, line 20: With Tom's work on [generating external IDs](https://github.com/alphagov/forms-admin/blob/af1c4c72bf96abd112e87bc7d9e0894967105b87/app/lib/external_id_provider.rb#L3), it made sense to use SecureRandom.base58. This was because it strips out characters that could be confused with each other, and I think in this case it would also make sense to avoid the I/l and 0/O (https://api.rubyonrails.org/classes/SecureRandom.html) in case they get mixed up. I think we could replace this with something like 

```suggestion
          submission_id = SecureRandom.base58(8).upcase
```

DavidBiddle: Ooh, great shout - didn't know that existed but should be really handy in case form fillers have to quote their reference over the phone or want to write it down.

DavidBiddle: Done!

app/controllers/forms/check_your_answers_controller.rb
```diff
@@ -81,5 +85,9 @@ def set_logging_context
         end
       end
     end
+
+    def generate_submission_reference
```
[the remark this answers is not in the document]
DavidBiddle: Yeah, think this would work. Thinking of adding it as a class method on the CheckYourAnswersStep class, do you reckon that makes sense?

spec/requests/forms/check_your_answers_controller_spec.rb
```diff
@@ -270,6 +276,10 @@
         expect(mail.govuk_notify_reference).to eq "for-my-ref"
       end
 
+      it "calls SecureRandom to generate an 8 digit reference code" do
+        expect(SecureRandom).to have_received(:base58).with(8).once
```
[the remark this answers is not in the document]
DavidBiddle: Yes. Have added now!

app/views/forms/submitted/submitted.html.erb
```diff
@@ -2,14 +2,17 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <%= govuk_panel(title_text: t('form.submitted.title'), classes: ["govuk-!-margin-bottom-7"]) %>
+    <%= govuk_panel(title_text: t('form.submitted.title'), classes: ["govuk-!-margin-bottom-7"], ) do %>
```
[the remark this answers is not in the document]
DavidBiddle: Fair point, let's see if I can chop these up

spec/lib/context_spec.rb
```diff
@@ -95,5 +95,20 @@
       expect(context2.find_or_create("1").show_answer).to eq("")
     end
   end
+
+  describe "submission references" do
```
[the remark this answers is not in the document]
DavidBiddle: Done

app/models/check_your_answers_step.rb
```diff
@@ -21,4 +21,8 @@ def state
   def end_page?
     true
   end
+
+  def self.generate_submission_reference
```
chao-xian, line 25: What do you think of putting this into the FormSubmissionService? We could also move the logging into that class, and when we call submit it returns the submission_reference... Thoughts?

chao-xian (commented):
Looking really good but just one suggestion which I think would make more sense in the code. Let me know what you think. Thanks David for breaking up the commits! It made it really easy to read and review.

app/models/check_your_answers_step.rb
DavidBiddle: Yeah, reckon that would work - will have a go!

thomasiles:
This looks really neat, well thought through and documented.

Not sure it matters and might have been considered already:

Will the changes to the session cause users with an existing session to get error pages and lose their current answers?

A reasonable choice, but might be worth a note.

app/models/check_your_answers_step.rb
DavidBiddle: @chao-xian That should be done now, thanks for the suggestion!

chao-xian (approved):
TYSM for your work on this David. A follow up PR to add the reference to the emails, and merge this now to unblock the Payment Links work is good by me.

thomasiles (approved):
amazing 🏅

PR #653, opened 2024-02-28
Log notification IDs for form submission

lfdebrux:

Log notification IDs after call to GOV.UK Notify is made, to make it easier to lookup the notification in their database.

To make it possible to have this code run under test, I had to introduce a new test mailer, as with the default Rails test mailer `govuk_notify_response` attribute isn't populated.


DavidBiddle (approved):
Code looks good and can see these fields being logged on my local machine 🎉

PR #659, opened 2024-03-04
Add reference numbers to email

DavidBiddle:

Trello card: https://trello.com/c/dm8jKunt/1361-generate-a-reference-for-form-submissions-email-work

Adds two new variables to the Notify personalisation for the submission email (sent to processors) and confirmation email (sent to form fillers, if requested):

- include_submission_reference: this is a boolean controlled by the `reference_numbers_enabled` feature flag.
- submission_reference: the reference number itself.

This means that once this is merged and deployed, we can add the variable to the Notify templates in the format `((INCLUDE_SUBMISSION_REFERENCE??Your reference number is ))((SUBMISSION_REFERENCE))`. When the feature flag is off, this will return nothing, so the email will be the same as it was previously. But when the flag is on, it will return `Your reference is R34R3NC3`.

### DIFF
--- a/app/mailers/form_submission_mailer.rb
+++ b/app/mailers/form_submission_mailer.rb
@@ -1,19 +1,19 @@
 class FormSubmissionMailer < GovukNotifyRails::Mailer
-  def email_completed_form(title:, text_input:, reference:, preview_mode:, timestamp:, submission_email:, submission_reference:)
+  def email_completed_form(text_input:, reference:, submission_email:, mailer_options:)
     set_template(Settings.govuk_notify.form_submission_email_template_id)
 
     set_personalisation(
-      title:,
+      title: mailer_options.title,
       text_input:,
-      submission_time: timestamp.strftime("%l:%M%P").strip,
-      submission_date: timestamp.strftime("%-d %B %Y"),
+      submission_time: mailer_options.timestamp.strftime("%l:%M%P").strip,
+      submission_date: mailer_options.timestamp.strftime("%-d %B %Y"),
       # GOV.UK Notify's templates have conditionals, but only positive
       # conditionals, so to simulate negative conditionals we add two boolean
       # flags; but they must always have opposite values!
-      test: make_notify_boolean(preview_mode),
-      not_test: make_notify_boolean(!preview_mode),
+      test: make_notify_boolean(mailer_options.preview_mode),
+      not_test: make_notify_boolean(!mailer_options.preview_mode),
       include_submission_reference: make_notify_boolean(FeatureService.enabled?(:reference_numbers_enabled)),
-      submission_reference: FeatureService.enabled?(:reference_numbers_enabled) ? submission_reference : "",
+      submission_reference: FeatureService.enabled?(:reference_numbers_enabled) ? mailer_options.submission_reference : "",
     )
 
     set_reference(reference)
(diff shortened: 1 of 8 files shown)


app/mailers/form_submission_mailer.rb
```diff
@@ -1,5 +1,5 @@
 class FormSubmissionMailer < GovukNotifyRails::Mailer
-  def email_completed_form(title:, text_input:, reference:, preview_mode:, timestamp:, submission_email:)
+  def email_completed_form(title:, text_input:, reference:, preview_mode:, timestamp:, submission_email:, submission_reference:)
```
chao-xian, line 2: There's an opportunity here to refactor https://refactoring.guru/introduce-parameter-object

chao-xian (commented):
Just a quick look but not done a proper review yet

spec/requests/forms/check_your_answers_controller_spec.rb
```diff
@@ -85,6 +87,10 @@
       allow(context_spy).to receive(:form_submitted?).and_return(repeat_form_submission)
       context_spy
     end
+
+    # rubocop:disable RSpec/AnyInstance
```
chao-xian, line 91: I'm a bit concerned about using this. There are pros and cons (see https://medium.com/beambenefits/rspec-tips-allow-any-instance-of-c11e38b098be) but I'm wondering if we need to test it in the controller. What are your thoughts?

chao-xian (commented):
Just a couple thoughts. Looking good! Thanks David

spec/requests/forms/check_your_answers_controller_spec.rb
chao-xian: OK cool, give it a go

app/mailers/form_submission_mailer.rb
DavidBiddle: I've done something along these lines - created a `FormSubmissionService::FormSubmissionMailerOptions` class which covers the parameters shared by this function and the one in `FormSubmissionConfirmationMailer`

spec/requests/forms/check_your_answers_controller_spec.rb
DavidBiddle: @chao-xian Done, thanks for the suggestion!

app/services/form_submission_service.rb
```diff
@@ -5,6 +5,8 @@ def call(**args)
     end
   end
 
+  FormSubmissionMailerOptions = Data.define(:title, :preview_mode, :timestamp, :submission_reference)
```
chao-xian, line 8: I know it's a nitpick but: Can we shorten this to `MailerOptions` or just `Options` please? Given it's already namespaced in the module, it's otherwise quite long 🙏

app/services/form_submission_service.rb
```diff
@@ -42,14 +42,12 @@ def submit_form_to_processing_team
     end
 
     unless @form.submission_email.blank? && @preview_mode
+
       mail = FormSubmissionMailer
-      .email_completed_form(title: form_title,
-                            text_input: email_body,
-                            preview_mode: @preview_mode,
+      .email_completed_form(text_input: email_body,
```
chao-xian, line 47: Not 100% sure and will need to look at it again tomorrow but feels like this could be something like:

```ruby
.email_completed_form(email, options)
```

app/services/form_submission_service.rb
DavidBiddle: Good shout, have gone with `MailerOptions`

DavidBiddle:
Have pushed an extra commit to resolve an ambiguity that was bugging me.

app/services/form_submission_service.rb
DavidBiddle: @chao-xian what would `email` contain here?

DavidBiddle: I'm inclined to do this in a subsequent PR but yes, something like that would make sense

chao-xian (approved):
Thanks David!

PR #663, opened 2024-03-11
1358 enable taking payments in forms via runner

SamJamCul:

https://trello.com/c/Gfqxj8yV/1358-enable-taking-payments-in-forms-via-runner

Adds payment stuff when there is a `payment_url` for a given form. This adds content and modifies the `form_submitted` page. 

It also adds content to form submission and confirmation emails. This all relies on the reference number feature flag being turned on, as it's necessary for payments. When there isn't a `payment_url` for a given form, no content relating to payments should appear. 

In order for this to fully function, the gov.uk/notify templates will need to be updated so that they accept the additional fields. We have templates that contain the appropriate content, but they'll need to replace the main ones in order for this to fully go live!

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -32,4 +32,10 @@ class Form < ActiveResource::Base
   def live_at_date
     try(:live_at).try(:to_time)
   end
+
+  def payment_url_with_reference(reference)
+    return nil if payment_url.blank?
+
+    "#{payment_url}?reference=#{reference}"
+  end
 end
--- a/app/views/forms/submitted/submitted.html.erb
+++ b/app/views/forms/submitted/submitted.html.erb
@@ -2,12 +2,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <%= govuk_panel(title_text: t('form.submitted.title'), classes: ["govuk-!-margin-bottom-7"], ) do %>
-      <% if FeatureService.enabled?(:reference_numbers_enabled) && @current_context.get_submission_reference.present? %>
-        <%= t('form.submitted.your_reference') %><br>
-        <strong><%= @current_context.get_submission_reference %></strong>
-      <% end %>
-    <% end %>
+    <%= render partial: "forms/submitted/submitted_panel", locals: {current_context: @current_context, } %>
 
     <% if @current_context.requested_email_confirmation? %>
       <p><%= t('form.submitted.email_sent') %></p>
@@ -16,6 +11,10 @@
     <%if @current_context.form.what_happens_next_markdown.present? %>
       <h2 class="govuk-heading-m"><%= t('form.submitted.what_happens_next') %></h2>
       <%= HtmlMarkdownSanitizer.new.render_scrubbed_markdown(@current_context.form.what_happens_next_markdown) %>
+    <% end %>
+
+    <% if @current_context.form.payment_url.present? && FeatureService.enabled?(:reference_numbers_enabled) && @current_context.get_submission_reference.present? %>
+      <%= govuk_button_to(t('form.submitted.continue_to_pay'), @current_context.form.payment_url_with_reference(@current_context.get_submission_reference))%>
     <% end %>
   </div>
 </div>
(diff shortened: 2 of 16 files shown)


app/views/forms/submitted/submitted.html.erb
```diff
@@ -2,11 +2,18 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <%= govuk_panel(title_text: t('form.submitted.title'), classes: ["govuk-!-margin-bottom-7"], ) do %>
-      <% if FeatureService.enabled?(:reference_numbers_enabled) && @current_context.get_submission_reference.present? %>
+    <% if @current_context.form.payment_url? && FeatureService.enabled?(:reference_numbers_enabled) && @current_context.get_submission_reference.present? %>
```
DavidBiddle, line 5: I don't think we define `form.payment_url?` anywhere - should probably be `form.payment_url.present?`

```suggestion
    <% if @current_context.form.payment_url.present? && FeatureService.enabled?(:reference_numbers_enabled) && @current_context.get_submission_reference.present? %>
```

chao-xian (commented):
Cool! Couple of suggestions though.

app/views/forms/submitted/submitted.html.erb
DavidBiddle: For me `form.payment_url? returns` true even if there isn't a payment link set

app/views/forms/submitted/submitted.html.erb
```diff
@@ -2,11 +2,18 @@
...
         <strong><%= @current_context.get_submission_reference %></strong>
       <% end %>
+    <% else %>
+      <%= govuk_panel(title_text: t('form.submitted.title'), classes: ["govuk-!-margin-bottom-7"], ) do %>
+        <% if FeatureService.enabled?(:reference_numbers_enabled) && @current_context.get_submission_reference.present? %>
+          <%= t('form.submitted.your_reference') %><br>
+          <strong><%= @current_context.get_submission_reference %></strong>
+        <% end %>
+      <% end %>
     <% end %>
```
DavidBiddle, line 17: I think we can make this less repetitive like so:
```suggestion
    <%= govuk_panel(
        title_text: @current_context.form.payment_url.present? ? t('form.submitted.need_to_pay') : t('form.submitted.title'),
        classes: class_names(
          "govuk-!-margin-bottom-7",
          { "app-panel--payment-required": @current_context.form.payment_url.present? }
        )
    ) do %>
      <% if FeatureService.enabled?(:reference_numbers_enabled) && @current_context.get_submission_reference.present? %>
        <%= t('form.submitted.your_reference') %><br>
        <strong><%= @current_context.get_submission_reference %></strong>
      <% end %>
    <% end %>
```

app/views/forms/submitted/submitted.html.erb
SamJamCul: Yes you're right, it should be checking `.present?` as otherwise it always comes back true!

app/models/form.rb
```diff
@@ -32,4 +32,8 @@ def live?(current_datetime = Time.zone.now)
   def live_at_date
     try(:live_at).try(:to_time)
   end
+
+  def payment_url_with_reference(reference)
+    payment_url.present? ? "#{payment_url}?reference=#{reference}" : nil
```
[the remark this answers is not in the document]
SamJamCul: Changed it

app/views/forms/submitted/submitted.html.erb
SamJamCul: I've added this as a separate partial

app/views/forms/submitted/submitted.html.erb
```diff
@@ -2,11 +2,18 @@
...
         <%= t('form.submitted.your_reference') %><br>
         <strong><%= @current_context.get_submission_reference %></strong>
       <% end %>
+    <% else %>
+      <%= govuk_panel(title_text: t('form.submitted.title'), classes: ["govuk-!-margin-bottom-7"], ) do %>
+        <% if FeatureService.enabled?(:reference_numbers_enabled) && @current_context.get_submission_reference.present? %>
+          <%= t('form.submitted.your_reference') %><br>
+          <strong><%= @current_context.get_submission_reference %></strong>
+        <% end %>
+      <% end %>
```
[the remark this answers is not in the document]
SamJamCul: I've used the refactor suggested below as the content for a separate partial

DavidBiddle (approved):
Nice one Sam - code looks good and things behave the way I expect locally!

PR #665, opened 2024-03-12
Generate user friendly reference numbers

SamJamCul:
https://trello.com/c/4lWx25f7/1426-update-form-reference-number-generation-logic

We want to try and match [GOV.UK Pay's reference number generation](https://github.com/alphagov/pay-products/blob/master/src/main/java/uk/gov/pay/products/util/RandomIdGenerator.java#L34), as these will be user facing numbers and we want to avoid any user unfriendly results. 

This should generate a 10 character reference, with the 3rd and 7th character always being a number. The characters `Z, I, 1, O and 0` are excluded as they're too confusable with each other (Z and 2 except keeping 2 is alright). 

Suggestions on how to make this more readable are welcome!

### DIFF
--- a/app/services/reference_number_service.rb
+++ b/app/services/reference_number_service.rb
@@ -1,0 +1,17 @@
+class ReferenceNumberService
+  ALPHABETICAL_CHARACTER_SET = %w[A B C D E F G H J K L M N P R S T U V W X Y].freeze
+  NUMERIC_CHARACTER_SET = %w[2 3 4 5 6 7 8 9].freeze
+  ALPHANUMERIC_CHARACTER_SET = ALPHABETICAL_CHARACTER_SET + NUMERIC_CHARACTER_SET
+  REFERENCE_LENGTH = 8
+
+  def self.generate
+    (0...REFERENCE_LENGTH)
+      .map { |i|
+        if i % 3 == 2
+          NUMERIC_CHARACTER_SET.sample
+        else
+          ALPHANUMERIC_CHARACTER_SET.sample
+        end
+      }.join
+  end
+end
--- a/app/services/form_submission_service.rb
+++ b/app/services/form_submission_service.rb
@@ -16,7 +16,7 @@ class FormSubmissionService
     @requested_email_confirmation = @email_confirmation_form.send_confirmation == "send_email"
     @preview_mode = preview_mode
     @timestamp = submission_timestamp
-    @submission_reference = generate_submission_reference
+    @submission_reference = ReferenceNumberService.generate
 
     @mailer_options = MailerOptions.new(title: form_title,
                                         preview_mode: @preview_mode,
@@ -177,9 +177,5 @@ private
     return nil if [@form.support_url, @form.support_url_text].all?(&:blank?)
 
     "[#{@form.support_url_text}](#{@form.support_url})"
-  end
-
-  def generate_submission_reference
-    SecureRandom.base58(8).upcase
   end
 end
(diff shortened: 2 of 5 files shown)


app/services/reference_number_service.rb
DavidBiddle, line 1: If I'm calculating right this gives us `(24^3)*8*(24^3)*8*(24^2) = 7044820107264` combinations.

This [collision probability calculator](https://kevingal.com/apps/collision.html) suggests that for a form with 100,000 submissions, there will be a 0.071% chance of two reference numbers matching. That feels like it's probably an acceptable risk but I'd be interested in other people's thoughts

DavidBiddle: If we did 8 characters with numbers at the 2nd and 5th digits, the number of combinations will be 68719476736 and there would be a 7.017% probability of a collision in 100,000 submissions

DavidBiddle: We expect they'll primarily be used as:
- a way for form processors to match a form submission email with a GOV.UK Pay payment
- a way for form fillers and government organisations to talk unambiguously about a form submission when they need to talk to each other (e.g. if a form filler email's a department's support address to ask what the status of their submission is, or if the department rings the user to clarify something in their submission).

I think you're correct that it only really matters on a per-form basis. I picked 100,000 submissions as a best guess for the upper end of the number of submissions an individual form might receive in a shortish space of time (i.e. XL bullies form plus some buffer). 

I don't think a collision is necessarily catastrophic - there will likely be other things form processors can disambiguate on if they notice the duplicate references (submission time, user's name/email, etc) - but I think if they don't notice the duplicate references that increases the risk of a payment being assigned to the wrong submission or the wrong user's details being updated in a case management system somewhere.

lfdebrux: How long did it take the XL Bully Form to get to 10,000 submissions?

SamJamCul: Interesting, if we drop down to only 8 character strings, we get about a 40% chance of colliding with 100k. That feels less acceptable, although this is over the course of a year. 

image

DavidBiddle: @lfdebrux looks like it took about 6 weeks

DavidBiddle: I guess one thing to note is that we already know we want to start logging these reference numbers in a database at some point. Once we do that we can enforce uniqueness, and we could put some monitoring in to check if collisions are happening frequently

app/services/reference_number_service.rb
```diff
@@ -0,0 +1,17 @@
+class ReferenceNumberService
+  RANDOM_SOURCE_SET_1 = %w[A B C D E F G H J K L M N P R S T U V W X Y].freeze
+  RANDOM_SOURCE_SET_2 = %w[2 3 4 5 6 7 8 9].freeze
+  RANDOM_SOURCE_UNION = RANDOM_SOURCE_SET_1 + RANDOM_SOURCE_SET_2
```
DavidBiddle, line 4: Optional, but I think it might be easier to see what this does by giving these more specific names:

```suggestion
  ALPHABETICAL_CHARACTER_SET = %w[A B C D E F G H J K L M N P R S T U V W X Y].freeze
  NUMERIC_CHARACTER_SET = %w[2 3 4 5 6 7 8 9].freeze
  ALPHANUMERIC_CHARACTER_SET = ALPHABETICAL_CHARACTER_SET + NUMERIC_CHARACTER_SET
```

DavidBiddle (approved):
Looks good! Raised one suggestion on the names of the character set constants, but happy for this to be merged whether or not you think that's a good idea

chao-xian (approved):

app/services/reference_number_service.rb
SamJamCul: Sorted, good suggestion!

DavidBiddle (approved):

PR #666, opened 2024-03-12
Tidy up FormSubmissionService and some related files

chao-xian:

This PR is just to tidy up the code before tackling the core goal of introducing the refactoring noted in the Trello card: https://trello.com/c/nu6ooJai/1420-refactor-submission-email-methods


app/domain/notify_template_body_filter.rb
```diff
@@ -0,0 +1,56 @@
+class NotifyTemplateBodyFilter
```
lfdebrux, line 1: Could you explain a bit for me what you're thinking is as to why this would go in the `domain` folder rather than the `lib` folder? I'm trying to understand to help me make similar decisions in future.

chao-xian: Mmm this one I think I was overzealous with as it's _not_ to do with our domain and could probably stay in `lib`

chao-xian: From convo in Tech Weekly we'll leave this re-organisation for now and keep an eye on things. Instead we'll add some structure to the `/lib` directory as an intermediate step.

chao-xian:
I've moved a small set of classes into a module named `Flow` as these seem to relate to the flow a user goes through when filling in a form. It now does feel like this module should go into something relating to the domain of our system...

app/services/notify_template_body_filter.rb
DavidBiddle, line 1: Our other services in this folder are all called thing_it_does_service.rb - is it worth renaming this to keep to that convention?

DavidBiddle: Not very imaginative but maybe`notify_template_body_filter_service.rb`or just `notify_template_filter_service.rb`

chao-xian: I think I'll go with `notify_template_service.rb` and filtering inside of it. TY!

chao-xian: Actually I'm going to name this `NotifyTemplateFormatter` and move it to the app lib...

DavidBiddle (approved):
Refactor looks great! I've done some light testing of the journey locally, it still works and the emails still look OK 🎉

PR #713, opened 2024-04-16
Stop elements other than summary list going full width

DavidBiddle:

Trello card: https://trello.com/c/hv6KQQUr/1224-check-your-answers-page-width

We have logic to make the summary list full width if the form contains a long text question. This is intended, but the original implementation made the _whole page_ full width, rather than just the summary list.

This PR puts the summary list in a separate grid row from the other elements on the page, and applies the full width logic to this row only. 

#### Before
Screenshot of the check your answers form, where one of the answers is a long piece of text describing the issue this PR fixes. The heading, summary list, confirmation email form and declaration all expand to the full width of the page wrapper.

#### After
Screenshot of the check your answers form, where one of the answers is a long piece of text describing the issue this PR fixes. The summary list expands to the full width of the page wrapper. The heading, confirmation email form and declaration all remain limited to two-thirds of the page wrapper's width.

### DIFF
--- a/app/views/forms/check_your_answers/show.html.erb
+++ b/app/views/forms/check_your_answers/show.html.erb
@@ -6,24 +6,32 @@
   <% end %>
 <% end %>
 
-<div class="govuk-grid-row">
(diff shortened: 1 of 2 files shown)


thomasiles (approved):
nice 👍

PR #734, opened 2024-04-30
Rename form objects to have suffix "Input"

stephencdaly:

The suffix "Form" for form objects which represent the input fields on a HTML form was difficult to work with, as we have a "Form" entity represing a GOV.UK Forms form, which is used throughout the application.

Use the suffix "Input" for form objects, to hopefully avoid this point of confusion.

### DIFF
--- a/app/controllers/forms/check_your_answers_controller.rb
+++ b/app/controllers/forms/check_your_answers_controller.rb
@@ -4,16 +4,16 @@ module Forms
       return redirect_to form_page_path(current_context.form.id, current_context.form.form_slug, current_context.next_page_slug) unless current_context.can_visit?(CheckYourAnswersStep::CHECK_YOUR_ANSWERS_PAGE_SLUG)
 
       setup_check_your_answers
-      email_confirmation_form = EmailConfirmationForm.new
+      email_confirmation_input = EmailConfirmationInput.new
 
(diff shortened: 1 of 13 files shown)


lfdebrux (approved):
LGTM!

lfdebrux:
I've just realised that this change will affect the end to end tests, because they rely on the `name` of some of the `inputs` when requesting a confirmation email:

https://github.com/alphagov/forms-e2e-tests/blob/01233e44de56d85a87ee0b5f9323b41a3edb9179/spec/support/feature_helpers.rb#L272

We shouldn't really be relying on the `name` of this input, it should be possible to use the label instead, but we will need to change the end to end tests before merging this PR.

stephencdaly:
> I've just realised that this change will affect the end to end tests, because they rely on the `name` of some of the `inputs` when requesting a confirmation email:
> 
> https://github.com/alphagov/forms-e2e-tests/blob/01233e44de56d85a87ee0b5f9323b41a3edb9179/spec/support/feature_helpers.rb#L272
> 
> We shouldn't really be relying on the `name` of this input, it should be possible to use the label instead, but we will need to change the end to end tests before merging this PR.

Good spot! Thanks, I'll update this

lfdebrux (approved):

PR #746, opened 2024-05-14
Validate the Form ID Before Calling Forms-Api

danworth:
Ensure the form id in the request is at least valid alphanumeric characters and raise `ResourceNotFound` if not (this is the response of `ActiveResource::find` if no record is discovered). This prevents Forms-Runner from calling Forms-Api with ids containing special characters which causes unhandled `URI::InvalidURIError` from the constructed URI. This seems preferable to simply URI encoding whatever is provided and calling Forms-Api when we know it is impossible for a form to have special chars in it's id both now and in the future.

The form ids in Forms-Api are currently integers but there are plans to extend this to an alphanumeric format in future. Validating for alphanumeric characters fixes the current problem of encountering URI::InvalidURIError whilst keeping it flexible enough for future change.

Sentry issue behind addressed on Support: https://govuk-forms.sentry.io/issues/5343630524/?alert_rule_id=14126387&alert_type=issue&notification_uuid=7d4b7c82-b6fd-48c8-b4e7-d0b0e5bf51a4&project=6576261&referrer=slack

The exception occurred from calling forms-runner with a form id including special URI characters, which when interpolated into the ActiveResource path throws an `URI::InvalidURIError`.

Exception causing request (still happens in staging): 
```
https://submit.staging.forms.service.gov.uk/form/<id>
```

Invalid Id now handled with Not Found page in dev where this change is deployed: 
```
https://submit.dev.forms.service.gov.uk/form/<id>
```

### Notes ###
We had discussed the use of the `params` argument in https://api.rubyonrails.org/v3.2.1/classes/ActiveResource/Base.html#method-c-find but after further digging it doesn't seem able to be used with our need to suffix `/live|draft|archived` to the path. Also whilst attempting to implement it, the end result of sending something we know will never return a form, since we'll never use special characters in the form id, seemed undesirable and better to validate the user input as close to it entering our system as possible whilst remaining [somewhat liberal to what we accept](https://en.wikipedia.org/wiki/Robustness_principle).

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -7,6 +7,8 @@ class Form < ActiveResource::Base
   has_many :pages
 
   def self.find_with_mode(id:, mode:)
+    raise ActiveResource::ResourceNotFound.new(404, "Not Found") unless id.to_s =~ /^[[:alnum:]]+$/
+
     return find_draft(id) if mode.preview_draft?
     return find_archived(id) if mode.preview_archived?
     return find_live(id) if mode.live?
--- a/spec/models/form_spec.rb
+++ b/spec/models/form_spec.rb
@@ -129,6 +129,35 @@ RSpec.describe Form, type: :model do
         expect(form.live?).to eq(false)
       end
     end
+
+    context "when validating the provided form id" do
+      let(:response_data) { { id: "Alpha123", name: "form name", live_at: nil }.to_json }
+
+      before do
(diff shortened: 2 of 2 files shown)


jamie-o-wilkinson (approved):
LGTM 🌟

danworth:
Just to note, that I ran the e2e tests against this change in dev and they passed also....

PR #747, opened 2024-05-16
Remove 'reference_numbers_enabled' feature flag

SamJamCul:
This removes the reference numbers feature flag, as this feature is now on by default.

It also adds a 'dummy_feature' flag so that our settings tests don't collapse due to there being no features. I'm open to suggestions on how to better handle this though.

Trello card: https://trello.com/c/3ZPv1xaf/1438-remove-feature-flag-for-payment-links-and-reference-numbers


app/views/forms/submitted/_submitted_panel.html.erb
```diff
@@ -5,7 +5,7 @@
       { "app-panel--payment-required": current_context.form.payment_url.present? }
     )
 ) do %>
-  <% if FeatureService.enabled?(:reference_numbers_enabled) && current_context.get_submission_reference.present? %>
+  <% if current_context.get_submission_reference.present? %>
```
SamJamCul, line 8: I can't quite tell why this is here, I think it's making sure that if the session gets lost or something then we don't fall over trying to display a missing reference.

app/views/forms/submitted/submitted.html.erb
```diff
@@ -13,7 +13,7 @@
       <%= HtmlMarkdownSanitizer.new.render_scrubbed_markdown(@current_context.form.what_happens_next_markdown) %>
     <% end %>
 
-    <% if @current_context.form.payment_url.present? && FeatureService.enabled?(:reference_numbers_enabled) && @current_context.get_submission_reference.present? %>
+    <% if @current_context.form.payment_url.present? && @current_context.get_submission_reference.present? %>
```
SamJamCul, line 16: Same here, not sure if we need to check that the submission reference is present, as it's always generated as the feature's always on

config/settings.yml
```diff
@@ -1,5 +1,6 @@
 features:
-  reference_numbers_enabled: false
+  # placeholder feature flag when no actual flags are in use
+  dummy_feature: false
```
SamJamCul, line 3: This is here because removing the reference numbers flag kind of imploded the tests. Rather than remove all of the feature flag test bits, it seemed easier to keep the scaffolding around by having at least one feature present at all times.

app/views/forms/submitted/_submitted_panel.html.erb
DavidBiddle: Yeah, I think this was mainly in here for when we switched it the feature on - in case people already had the confirmation page open and refreshed it.

Kind of edge casey anyway, but especially so now that the feature has been on for a while, so we might be able to remove it. 

Alternately we might want to leave it as it is for now, and raise a ticket for replacing this with some more elegant behaviour (e.g. redirecting people back to the start of the form if there's nothing in the session).

config/settings.yml
DavidBiddle: I don't mind this but FYI we do have a way of keeping some of the scaffolding here and in the tests when there aren't any flags:
https://github.com/alphagov/forms-runner/blob/0a3528d252c7a6d92b6fa7735aeb2c90ec0d4ad9/config/settings.yml#L1
https://github.com/alphagov/forms-runner/blob/0a3528d252c7a6d92b6fa7735aeb2c90ec0d4ad9/spec/config/settings_spec.rb#L21-L27

DavidBiddle: Interested in what other people think on this, because the existing process for updating the tests is still a bit annoying and this definitely makes it easier

app/views/forms/submitted/_submitted_panel.html.erb
SamJamCul: I think save it for a different ticket then, if they lose their cookie then we can't really show them an appropriate submission page. This would be particularly important for the payment links bits too, so I'd rather leave this here to avoid showing them anything misleading.

config/settings.yml
chao-xian: This feature flag service is our own code and design, which means we should have that instruction documented somewhere already?

lfdebrux: I prefer what's in https://github.com/alphagov/forms-runner/commit/0a3528d252c7a6d92b6fa7735aeb2c90ec0d4ad9, it reads better to me. I appreciate it might be subjective though.

chao-xian: Wait sorry I forgot we use the config gem - so we don't need to include examples? Meaning David's proposal works. If we really wanted we can leave a comment in the settings file instead?

SamJamCul: I've changed it to the suggested one, if it's a railsy standard then it's much more suitable!

DavidBiddle (approved):
Works locally and code looks good!

PR #759, opened 2024-05-24
Introduce Rake tasks for pausing, unpausing, and getting the status of pipelines

AP-Hunt:

Trello card: https://trello.com/c/HFSmspAf/381-automatically-deploy-to-dev-environments-after-production

We're beginning to automatically deploy changes that are OK in production into dev environment(s). This means that dev environment pipelines will get run much more frequently, and the chances our test changes in the environment will get overridden are much higher. To counteract this we intend to pause the pipelines when we've deployed dev changes. 

This PR adds 3 Rake tasks to help with that:
* `rake pipeline:ENV:pause`
* `rake pipeline:ENV:unpause`
* `rake pipeline:ENV:status`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -54,7 +54,9 @@ gem "uk_postcode"
 # For structured logging
 gem "lograge"
 
+# For AWS interactions
 gem "aws-sdk-cloudwatch"
+gem "aws-sdk-codepipeline", "~> 1.72"
 
 group :development, :test do
   # See https://guides.rubyonrails.org/debugging_rails_applications.html#debugging-with-the-debug-gem
--- a/lib/tasks/pause_pipelines.rake
+++ b/lib/tasks/pause_pipelines.rake
@@ -1,0 +1,130 @@
+require "aws-sdk-codepipeline"
+
+namespace :pipeline do
+  %w[dev prod staging user-research].each do |env_name|
+    namespace env_name do
+      desc "Pause the pipeline in #{env_name}"
+      task pause: :environment do
+        pipeline_name = pipeline_name env_name
+
+        puts "Pausing pipeline #{pipeline_name}"
+
+        codepipeline = Aws::CodePipeline::Client.new
+        pipeline_definition = codepipeline.get_pipeline(name: pipeline_name)
+
+        unless pipeline_definition.pipeline.stages.any?
+          raise "Pipeline definition has no stages"
+        end
+
+        unless pipeline_definition.pipeline.stages.count > 1
+          raise "Pipeline has only one stage. It cannot be paused in a way that can be unpaused from the console if necessary."
+        end
+
+        second_stage = pipeline_definition.pipeline.stages[1]
+
+        default_reason = "Pipeline paused by #{ENV['USER']} using rake task"
+        reason = get_pause_reason(default_reason)
+        codepipeline.disable_stage_transition(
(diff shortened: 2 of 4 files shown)


lib/tasks/pause_pipelines.rake
```diff
@@ -0,0 +1,130 @@
...
+
+        unless pipeline_definition.pipeline.stages.any?
+          raise "Pipeline definition has no stages"
+        end
+
+        unless pipeline_definition.pipeline.stages.count > 1
+          raise "Pipeline has only one stage. It cannot be paused in a way that can be unpaused from the console if necessary."
+        end
+
+        second_stage = pipeline_definition.pipeline.stages[1]
```
jamie-o-wilkinson, line 23: So does this assume that all pipelines do something temporary (e.g. source the repo) in their first stage? It would be ideal if we could find the stage by action rather than relying on the position of the stage never changing. Ideal, but not essential 😄

lib/tasks/pause_pipelines.rake
```diff
@@ -0,0 +1,130 @@
...
+          raise "Pipeline definition has no stages"
+        end
+
+        unless pipeline_definition.pipeline.stages.count > 1
+          raise "Pipeline has only one stage. It cannot be unpaused because there is no transition from the source stage to unpause."
+        end
+
+        second_stage = pipeline_definition.pipeline.stages[1]
+
+        codepipeline.enable_stage_transition(
```
jamie-o-wilkinson, line 56: What if other stage transitions have been disabled manually? Would it be possible to find all paused transitions and enable each of them?

lib/tasks/pause_pipelines.rake
AP-Hunt: AWS CodePipeline requires that the pipeline contains at least two stages, the first of which must be a source stage. We always want the transition into the second stage to be the pausing point, regardless of what the second stage does.

https://docs.aws.amazon.com/codepipeline/latest/userguide/reference-pipeline-structure.html

We're able to pause the transition going into the first stage, but that's not an operation you can undo in the UI for some reason.

lib/tasks/pause_pipelines.rake
AP-Hunt: I pondered that too. I figured that if somebody had paused the pipeline in a different way to these helpers then they definitely had something specific in mind. I thought that undoing something that wasn't done by this would be an unexpected side effect.

Maybe we should look for other paused stages and alert the user to them after we've unpaused?

lib/tasks/pause_pipelines.rake
jamie-o-wilkinson: Cool, that makes sense. Thanks for the explanation!

Gemfile
```diff
@@ -86,3 +86,5 @@ group :test do
   # axe-core for running automated accessibility checks
   gem "axe-core-rspec"
 end
+
+gem "aws-sdk-codepipeline", "~> 1.72"
```
DavidBiddle, line 90: Might be worth moving this above the two `group` blocks for tidiness

AP-Hunt: Done :+1: I've grouped it with the other AWS SDK package

DavidBiddle (approved):
Code looks good and works for me locally 🎉

PR #762, opened 2024-05-28
Show error page when user tries to submit incomplete form

stephencdaly:

Trello card: https://trello.com/c/x8aQr5Y6/1501-it-is-possible-to-submit-an-incomplete-form-after-submitting-a-completed-form

It is possible for a user to submit an incomplete form after they have already submitted a form if they duplicate the tab with the  "Check your anwsers" page before submitting. If they then continue to answer at least one question in a new submission, they were able to go back to the already open "Check your answers" page and submit the incomplete form.

One way of reproducing this is as follows:

1. Create a form with more than one question, none of which are optional
2. Fill out the form until the check your answers page
3. Duplicate the browser tab with the check your answers page twice
4. In the last tab, submit the form
5. In the second tab, click change on one of the questions, answer the question and click continue once
6. In the first tab, submit the form

This commit causes an error page to be shown if the user manages to click the submit button on the check your answers page, but all  questions have not been answered. The error page has a link to go back to where they should be in the form.

There is a small chance that a user could see this error page if they haven't already submitted a form in the case that a new version of the form is made live with additional questions at the end when a  user is on the check your answers page. This is because we check whether the last question they answered is the last question in the form before we allow it to be submitted.

Sending an event to Sentry when a user attempts to submit an incomplete form has been removed as this was added as a temporary measure to be informed if this happened again before we fixed the bug,  but we continue to log an event to Splunk.

The error page looks like this:

Screenshot 2024-05-28 at 16 42 35

### DIFF
--- a/app/controllers/forms/check_your_answers_controller.rb
+++ b/app/controllers/forms/check_your_answers_controller.rb
@@ -13,31 +13,28 @@ module Forms
       email_confirmation_input = EmailConfirmationInput.new(email_confirmation_input_params)
       requested_email_confirmation = email_confirmation_input.send_confirmation == "send_email"
 
-      if email_confirmation_input.valid?
-        if current_context.form_submitted?
-          redirect_to error_repeat_submission_path(current_form.id)
-        else
-          unless current_context.can_visit?(CheckYourAnswersStep::CHECK_YOUR_ANSWERS_PAGE_SLUG)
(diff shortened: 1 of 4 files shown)


SamJamCul (approved):
Looks like a good fix, I've tried to find a way to break it and couldn't find anything.

PR #794, opened 2024-06-20
Add Google Analytics snippet

DavidBiddle:

Trello card: https://trello.com/c/7jvfS18K/1585-add-google-analytics-4-to-forms-runner

This card adds Google analytics to any page, as long as the flag (`Settings.analytics_enabled`) is `true`.

Largely copied from the analytics work Tom did in the product pages. In future tickets we'll also copy over the cookie consent stuff from that repo, but this ticket/PR is limited to adding the snippet and putting it behind the flag.

You can test whether GA is running on a particular page by opening your browser console and seeing if typing `window.dataLayer` returns anything - if it returns `undefined` it isn't running, if it returns an array of event objects it is running.

### DIFF
--- a/app/frontend/entrypoints/application.js
+++ b/app/frontend/entrypoints/application.js
@@ -1,3 +1,8 @@
 import { initAll } from 'govuk-frontend'
+import { installAnalyticsScript } from '../javascript/google-tag'
+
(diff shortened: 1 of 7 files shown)


stephencdaly (approved):
Looks good to me! Well tested code, and I did the suggested test with the `analytics_enabled` setting on and off and it works as expected.

PR #811, opened 2024-07-01
Update cookies page and add cookie consent form

DavidBiddle:

Trello card: https://trello.com/c/1e6Mzt5m/1586-update-cookies-page-for-ga4-for-the-runner

Makes updates to the cookies page to reflect the analytics changes - largely copied from forms-product-pages:
- updates the content to include information about analytics cookies
- adds a cookie consent form so that users can set their preferences from this page
- removes the cookie banner from the page (since the consent form covers the same functionality)

Includes some changes similar to those in the [cookie banner PR](https://github.com/alphagov/forms-runner/pull/796):

- putting the analytics content on this page behind the `analytics_enabled` feature flag
- moving some of the logic out of application.js and into the cookie consent form component
- putting the cookie consent form component JS in a ViewComponent with the cookie banner HTML, instead of in the frontend folder

### Screenshots
Before:
Cookies page, with a cookie banner at the top and a section for essential cookies. The cookie banner is the only way to set the preference for analytics cookies on this page.

After:
Cookies page, with no cookie banner at the top. The essential cookies section now includes an entry for analytics_consent, and there is a new section for analytics cookies. Under the analytics cookies section is a form for setting analytics cookies

### Testing instructions
You can see whether GA is being loaded by checking the `<head>` element to see if it contains the google tag manager script (`<script async="" src="https://www.googletagmanager.com/gtm.js?id=GTM-T2SJXKKQ"></script>`). You should only see this if you've accepted analytics cookies.

### DIFF
--- a/app/views/application/cookies.html.erb
+++ b/app/views/application/cookies.html.erb
@@ -25,29 +25,31 @@
         ["analytics_consent", "Saves your cookie consent settings", "1 year"],
       ]) %>
 
-      <h2 class="govuk-heading-l">Analytics cookies (optional)</h2>
+      <% if Settings.analytics_enabled %>
+        <h2 class="govuk-heading-l">Analytics cookies (optional)</h2>
 
-      <p>We use Google Analytics software to understand how people use GOV.UK Forms. This information helps us to improve our service.</p>
+        <p>We use Google Analytics software to understand how people use GOV.UK Forms. This information helps us to improve our service.</p>
 
-      <p>Google is not allowed to use or share our analytics data with anyone.</p>
+        <p>Google is not allowed to use or share our analytics data with anyone.</p>
 
-      <p>Google Analytics stores anonymised information about:</p>
+        <p>Google Analytics stores anonymised information about:</p>
 
-      <ul class="govuk-list govuk-list--bullet">
-        <li>the pages you visit</li>
-        <li>how long you spend on each page</li>
-        <li>how you arrived at the site</li>
-        <li>what you click on while you visit the site</li>
-        <li>the device and browser you use</li>
-      </ul>
+        <ul class="govuk-list govuk-list--bullet">
+          <li>the pages you visit</li>
+          <li>how long you spend on each page</li>
+          <li>how you arrived at the site</li>
+          <li>what you click on while you visit the site</li>
+          <li>the device and browser you use</li>
+        </ul>
 
-      <%= govuk_table(rows: [
-        ["Name", "Purpose", "Expires"],
-        ["_ga", "Helps us count how many people visit GOV.UK Forms by telling us if you’ve visited before.", "2 years"],
-        ["_gid", "Helps us count how many people visit GOV.UK Forms by telling us if you’ve visited before.", "24 hours"],
-        ["_gat_UA-[random number]", "Helps us reduce the number of unnecessary tracking requests we send to Google Analytics.", "1 minute"],
-      ]) %>
+        <%= govuk_table(rows: [
+          ["Name", "Purpose", "Expires"],
+          ["_ga", "Helps us count how many people visit GOV.UK Forms by telling us if you’ve visited before.", "2 years"],
+          ["_gid", "Helps us count how many people visit GOV.UK Forms by telling us if you’ve visited before.", "24 hours"],
+          ["_gat_UA-[random number]", "Helps us reduce the number of unnecessary tracking requests we send to Google Analytics.", "1 minute"],
+        ]) %>
 
+      <% end %>
     <% end %>
   </div>
 </div>
--- a/app/views/application/cookies.html.erb
+++ b/app/views/application/cookies.html.erb
@@ -1,7 +1,28 @@
 <% set_page_title("Cookies") %>
 
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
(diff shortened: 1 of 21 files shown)


app/views/application/cookies.html.erb
```diff
@@ -1,21 +1,55 @@
...
+
+      <p>We use cookies to make this form work.</p>
+
+      <p>Find out <a rel="external" href="https://ico.org.uk/your-data-matters/online/cookies/">how to manage cookies</a>.</p>
+
+      <h2 class="govuk-heading-l">Essential cookies</h2>
+
+      <p>We use essential cookies to make forms work.</p>
+
+      <%= govuk_table(rows: [
```
lfdebrux, line 22: Minor nitpick, but I noticed that in the forms-product-page app the tables listing the cookies have the text in the first column in bold.

app/views/application/cookies.html.erb
```diff
@@ -1,21 +1,55 @@
...
+
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-xl">Cookies</h1>
-    <p>Cookies are small files saved on your phone, tablet or computer when you visit a website.</p>
+    <%= render(CookieConsentFormComponent::View.new) do %>
+      <h1 class="govuk-heading-xl">Cookies</h1>
+      <p>Cookies are small files saved on your phone, tablet or computer when you visit a website.</p>
+
+      <p>We use cookies to make this form work.</p>
```
lfdebrux, line 14: Do we not need something here about how we use cookies to collect information about how the website is used? Similar to what is at https://www.forms.service.gov.uk/cookies.

app/views/application/cookies.html.erb
```diff
@@ -1,21 +1,55 @@
...
+        <ul class="govuk-list govuk-list--bullet">
+          <li>the pages you visit</li>
+          <li>how long you spend on each page</li>
+          <li>how you arrived at the site</li>
+          <li>what you click on while you visit the site</li>
+          <li>the device and browser you use</li>
+        </ul>
 
-    <p>We use essential cookies to make this form work. UK law does not require us to ask for consent for these cookies because they are essential for the service to function.</p>
+        <%= govuk_table(rows: [
```
lfdebrux, line 45: Does the table need a caption?

app/views/application/cookies.html.erb
```diff
@@ -1,21 +1,55 @@
 <% set_page_title("Cookies") %>
```
lfdebrux, line 1: For some reason on my machine this page has the title 'Cookies - GOV.UK', without the 'GOV.UK Forms' that I would expect.

DavidBiddle:
Moving back to draft - will resume after firebreak

app/views/application/cookies.html.erb
DavidBiddle: Good spot and not a nitpick - they're also semantically table headers in forms-product-page and they should be here too!

app/views/application/cookies.html.erb
DavidBiddle: I think @AnneCarr said this wording was OK when we drafted it - we're explaining the analytics in detail further down. But I've not got any objection to changing it if we think being explicit here is better

app/views/application/cookies.html.erb
DavidBiddle: This affects all of the generic pages (cookies, accessibility, error pages). Have pushed a change to fix this

app/views/application/cookies.html.erb
DavidBiddle: Good spot, have added 👍

spec/features/analytics/cookies_page_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+      end
+    end
+
+    context "when the user has rejected cookies" do
+      before do
+        within_fieldset "Do you want to accept analytics cookies?" do
+          choose "No"
+        end
+
+        click_button "Save cookie settings"
```
stephencdaly, line 59: For feature tests, should we be following the pattern of having `scenario`s that contain all actions and assertions in one block using BDD style rather than `contexts` with `it`s?

For this page, performance of the tests shouldn't be much of a concern but this might lead to bad patterns in other feature tests where slow actions are repeated before each assertion rather than just performed once.

Is there a particular reason that doesn't make sense for this test?

stephencdaly (approved):
Looks good to me! @lfdebrux 's comments have been addressed. I've tested locally and everything works as expected.

I've left a comment about the feature tests, but happy with what you've done if there's a good reason for this deviating from the standard pattern.

spec/features/analytics/cookies_page_spec.rb
DavidBiddle: A combination of:
- being vaguely paranoid about cookies being preserved across tests
- thinking of scenarios as user journeys, I figured it wouldn't be particularly common for a user to visit this pag emultiple times in one session varying all of the options

I think I could accomplish both of those with a few separate scenarios (one for each current context) though, so will change it to match the convention in other feature tests.

DavidBiddle: @stephencdaly have updated, thanks for the suggestion!

stephencdaly (approved):
Nice work!

PR #823, opened 2024-07-04
WIP: Refactor Form model to abstract away source of forms

thomasiles:

This contains the first steps towards merging forms-admin with forms-api. This step abstracts away the source of forms, such that we can switch between using ActiveResource and directly making requests to the API via a feature flag.

Trello card: https://trello.com/c/DKW2M1eK


thomasiles:
I'm closing this as we don't need it anymore, given the migration to use V2 of the API

PR #827, opened 2024-07-09
Use CurrentAttributes to store and populate log context

stephencdaly:

Trello card: https://trello.com/c/tjnJrmjR/7-improve-default-information-we-add-to-logs

We already use Lograge to add attributes about the request context, such as the `request_id`, `form_id` etc to the request logs for incoming requests to the app.

For other application logs, we use a class called EventLogger to log events as JSON, and pass around a `logging_context` object that we store on the `ApplicationController` and add to in various places throughout a request to add useful JSON fields to the log lines.

Simplify this approach by using an implementation of the `ActiveSupport::CurrentAttributes` abstract super class, which provides a thread-isolated attributes singleton, which resets automatically before and after each request. This implementation - `CurrentLoggingAttributes` stores the attributes we expect to be present on most requests.

Introduces a custom logger class - `ApplicationLogger` to use for application logs. This makes use of the `CurrentAttributes` implementation to extract common logging fields and appends them to the JSON object for a log event.

This way, we do not have to pass around a `logging_context` between different methods, which can be a bit confusing - and if we forget to do this or make a mistake with how we do it, it will cause us not to log the request context attributes.

This also means that Rails default logs, such as for outgoing requests to other apps will have all the requests context attributes.

We can still pass in log-line specific additional attributes as a hash when we log and these will be added to the JSON object that we log.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,7 +2,7 @@
 
 class ApplicationController < ActionController::Base
   before_action :set_request_id
-  before_action :set_logging_context
+  before_action :set_logging_attributes
   before_action :check_maintenance_mode_is_enabled
   after_action :add_robots_header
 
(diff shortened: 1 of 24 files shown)


danworth:
>This PR attempts to keep as much of the logging context the same as possible. In most cases, it just replaces adding to the logging_context hash we pass around by adding things to the CurrentLoggingAttributes, so attributes will appear on the same lines as before.

To help with the review please can you call out any changes to this, otherwise I feel subtle changes might be easily missed from reviewing the code alone. As mentioned Anne uses the `events` in her reporting so please confirm whether anything relating to that changes.

stephencdaly:
> > This PR attempts to keep as much of the logging context the same as possible. In most cases, it just replaces adding to the logging_context hash we pass around by adding things to the CurrentLoggingAttributes, so attributes will appear on the same lines as before.
> 
> To help with the review please can you call out any changes to this, otherwise I feel subtle changes might be easily missed from reviewing the code alone. As mentioned Anne uses the `events` in her reporting so please confirm whether anything relating to that changes.

Good point! The only thing that's intentionally changed here is that the `event` logs now include a message field which is just `message: "Form event"` for consistency. 

Other than that, I did make sure that everything stays the same. I've done my best to test this both with automated tests and pushing up to dev and testing it there, so I have pretty good confidence in this.

chao-xian (approved):
lgtm thank you @stephencdaly

DavidBiddle (dismissed):
Nice one - big improvement in readability! 🎉

chao-xian (approved):

PR #838, opened 2024-07-16
Worst pr

chao-xian:

Trello card: 


app/controllers/forms/base_controller.rb
```diff
@@ -7,12 +7,12 @@ def redirect_to_friendly_url_start
       LogEventService.log_form_start(logging_context) unless mode.preview?
     end
 
-    rescue_from ActiveResource::ResourceNotFound, StepFactory::PageNotFoundError do
+    rescue_from ActiveResource::ResourceNotFound, Flow::StepFactory::PageNotFoundError do
```
AP-Hunt, line 10: Can we import/open `Flow` so that we can continue to say `StepFactory::...` instead of `Flow::StepFactory::...`. It's adding noise.

app/lib/flow/context.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    def find_or_create(page_slug)
+      step = completed_steps.find { |s| s.page_slug == page_slug }
+      step || @step_factory.create_step(page_slug)
+    end
+
+    def save_step(step)
+      return false unless step.valid?
+
+      step.save_to_context(@form_context)
```
AP-Hunt, line 28: If all this method is doing is checking validity and calling a method on `step`, couldn't that be something `step.save_to_context` does itself?

AP-Hunt (commented):
The first commit here ought to be split in to many commits that tell a story of both why we've refactored this and how.

AP-Hunt:
..... The GitHub UI betrayed me

stephencdaly (commented):
I'm finding it quite difficult to work out whether the files that have been moved have changed at all, as they're showing up in the PR as deletions and additions. Would you be able to separate moving the files into its own commit, and if anything has changed in them, make these changes in a different commit please?

This PR would generally benefit from splitting it out into smaller commits that explain the different stages of the refactor so it's easier to follow. For example, splitting out moving the NotifyTemplateFormatter into its own class could be a commit.

Could you also improve the commit messages to explain what has changed an why?

jamie-o-wilkinson (commented):
This PR is quite large, is it possible to break it down into smaller chunks in individual commits? This means big changes can be reviewed bit by bit which is easier to digest, and lends itself to a more descriptive commit history. It would also be really helpful to give the PR more description.

(should be inline comment): Perhaps this move of the Context class could be in a single commit? Then any necessary changes can be made in the next commit.

(should be inline comment): It's nice this class has been moved into it's own file. Perhaps another candidate for a separate commit?

app/lib/notify_template_formatter.rb
```diff
@@ -0,0 +1,56 @@
...
+       prep_answer_text(page.show_answer_in_email)].join
+    }.join("\n\n---\n\n").concat("\n")
+  end
+
+  def prep_question_title(question_text)
+    "# #{question_text}\n"
+  end
+
+  def prep_answer_text(answer)
+    return "\\[This question was skipped\\]" if answer.blank?
```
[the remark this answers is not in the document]
chao-xian: I should do that as a subsequent PR... thanks @AP-Hunt !

PR #862, opened 2024-08-05
Generate CSV data for a submission

stephencdaly:

Trello card: https://trello.com/c/BjjG6uTU/1666-generate-csv-data-for-a-submission

If the `attach_csv_to_submission_email` feature flag is enabled, write the submission to a CSV in a temporary file.

In future, we will send the CSV to form processors using Notify.

We will also enhance the feature flag so it can be enabled on a form by form basis later.

The first column is the form reference, followed by a column per question where the header is the question name.

This uses the Ruby "csv" library, which offers basic CSV reading/writing functionality.

The outputted CSV file will use the default unix line endings and UTF-8 encoding.

### How to test

1. Comment out the `file.unlink` line in `app/services/form_submission_service.rb` locally so the temporary file we write the CSV to is not immediately deleted
2. Enable the `attach_csv_to_submission_email` feature flag
3. Make a form submission.
4. Look for the `"Wrote submission CSV"` log line. Copy the `path` field from the log line to get the path to the CSV file.
5. Inspect the CSV file by running `open <path_to_csv` or `cat <path_to_csv>`

### DIFF
--- a/app/services/form_submission_service.rb
+++ b/app/services/form_submission_service.rb
@@ -40,11 +40,15 @@ class FormSubmissionService
 
     unless @form.submission_email.blank? && @preview_mode
 
+      if FeatureService.enabled?("attach_csv_to_submission_email")
+        write_csv_file
+      end
+
       mail = FormSubmissionMailer
-      .email_confirmation_input(text_input: email_body,
-                                notify_response_id: @email_confirmation_input.submission_email_reference,
-                                submission_email: @form.submission_email,
-                                mailer_options: @mailer_options).deliver_now
+               .email_confirmation_input(text_input: email_body,
+                                         notify_response_id: @email_confirmation_input.submission_email_reference,
+                                         submission_email: @form.submission_email,
+                                         mailer_options: @mailer_options).deliver_now
 
       CurrentLoggingAttributes.submission_email_id = mail.govuk_notify_response.id
     end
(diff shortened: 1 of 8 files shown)


app/services/form_submission_service.rb
```diff
@@ -40,11 +40,15 @@ def submit_form_to_processing_team
 
     unless @form.submission_email.blank? && @preview_mode
 
+      if FeatureService.enabled?("attach_csv_to_submission_email")
```
jamie-o-wilkinson, line 43: I think this is good for now, but I guess in the future we might benefit from generalising the submission service a bit, and have something like a CSV submitter, JSON submitter, S3 submitter, etc., all of which behave like a generic "submitter"

DavidBiddle (approved):
Code looks good and works as described 🎉

spec/services/submission_csv_service_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  after do
+    test_file.unlink
+  end
+
+  describe "#write" do
+    it "writes submission to CSV file" do
+      service.write
+      expect(CSV.open(test_file.path).readlines).to eq(
+        [
+          ["Reference", "What is the meaning of life?", "What is your email address?"],
```
jamie-o-wilkinson, line 23: Excellent test data 🤖

jamie-o-wilkinson (approved):
Looks good to me as a starting point, nice one! 🥇

app/services/form_submission_service.rb
stephencdaly: Good point! Yeh let's chuck it in here for now as we'll still be sending in the email, but as soon as we start looking at the S3 upload then yeh it would probably be good to look at doing this.

spec/services/submission_csv_service_spec.rb
stephencdaly: Ah I can't take credit for this, I copied some existing test data 😆

chao-xian:
Lovely work - ty @stephencdaly

PR #867, opened 2024-08-06
Make methods in form_submission_service private

stephencdaly:

There was no need for these methods to be public, but they were called directly in tests.

Modify the tests to just call the `submit` method whilst maintaining the same tests for sending the submission/confirmation email.

**NOTE**: Hiding whitespace changes will make this PR a lot easier to review because contexts were moved around in the tests so indentation changed.

### DIFF
--- a/app/services/form_submission_service.rb
+++ b/app/services/form_submission_service.rb
@@ -31,6 +31,8 @@ class FormSubmissionService
     @submission_reference
   end
 
+private
+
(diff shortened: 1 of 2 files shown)


jamie-o-wilkinson (approved):
LGTM 👍

chao-xian (dismissed):
~Why are we removing the tests?~ Sorry it's early and I panic-hit the button.

PR #868, opened 2024-08-07
Use per form feature flagging for generating CSV

stephencdaly:

Trello card: https://trello.com/c/BjjG6uTU/1666-generate-csv-data-for-a-submission

Pass in the form to the FeatureService when checking whether the `attach_csv_to_submission_email` feature flag is enabled. This will allow enabling this flag on a form by form basis.

We don't think this requires additional tests - we test the feature flag when it is set globally in `form_submission_service_spec` and we test applying feature flags for specific forms in `feature_service_spec`.

### DIFF
--- a/app/services/form_submission_service.rb
+++ b/app/services/form_submission_service.rb
@@ -42,7 +42,7 @@ private
 
     unless @form.submission_email.blank? && @preview_mode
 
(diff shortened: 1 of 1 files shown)


jamie-o-wilkinson (approved):
This behaves as before (i.e. feature disabled) for me locally 👍

PR #871, opened 2024-08-08
Accept an array of form IDs for per form feature flagging

jamie-o-wilkinson:

This commit changes per form feature flagging to accept a comma separated list of form IDs for which a feature is enabled. This only applies if the feature flag itself is disabled.
    
This is to facilitate feature flagging per form separately in different environments.

All environment variables are strings, and because we want to be able to test the CSV submissions with users imminently we will just do the string parsing in the feature service for now.

It might be possible to persuade Terraform to provide an array to the application such that it is parsed as an array. We'll look at this next.

Trello card: https://trello.com/c/9XvRKwdK

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,17 +151,22 @@ features:
 
 And check with `FeatureService.enabled?("some.nested_feature")`.
 
-You can also set a feature for a specific form:
+#### Enabling feature flags for specific forms
+
+You can also enable a feature flag just for specific form(s) by providing a list of form IDs.
 
 ```yaml
 features:
   some_feature:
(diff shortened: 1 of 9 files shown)


stephencdaly (approved):
I paired on this - so someone else is welcome to review. But approving so we can merge so we can enable for a form for some user research today.

stephencdaly (approved):
I paired on this - so someone else is welcome to review. But approving so we can merge so we can enable for a form for some user research today.

DavidBiddle (approved):

PR #872, opened 2024-08-09
Change all url helpers to use keyword arguments

thomasiles:
Currently the feature tests call URL helpers, like form_page_path, with a mix of positional and keyword arguments .

An example of a positional argument call:
form_page_path("preview-draft", 2, form_response_data.form_slug, 1)

An example of a keyword call:
form_page_path(mode: "preview-draft", form_id: 2, form_slug: form_response_data.form_slug, page_slug: 1)

This is fine, but inconsistent.

This PR changes all calls to use the keyword form. This makes it clearer
which arguments are for which routing segment. The keyword argument is
clearer because there are 4 arguments which makes it hard to get right.

It also allows adding new optional parameters to the calls, which will
be useful in the future.

There maybe a better way to do this in general, rather than using the helpers with all params like this.

The application code doesn't need to be changed.
Somehow the helpers in tests don't work the same way as they do in app code. I
think this is partly because of settings like `default_url_options`, which isn't
used in the tests.

This doesn't currently matter though because our routes don't have lots of
optional params, which makes using positional arguments difficult.

This commit does the simplest thing to make it consistent.

Trello card: 

### DIFF
--- a/spec/requests/forms/base_controller_spec.rb
+++ b/spec/requests/forms/base_controller_spec.rb
@@ -129,7 +129,7 @@ RSpec.describe Forms::BaseController, type: :request do
     end
 
     it "renders the page with a link back to the form start page" do
-      expect(response.body).to include(form_page_path("form", 2, form_response_data.form_slug, 1))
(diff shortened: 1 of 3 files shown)


SamJamCul (approved):
Looks good, much better than what we had before

PR #873, opened 2024-08-09
Add "repeatable?" to Page

thomasiles:
- **Make Page factory work with ActiveResource**
- **Move Page specs into another describe**
- **Ensure Page factory is valid**
- **Change answer_settings tests to use factory**
- **Add Page#repeatable?**

### Refactor Page specs and add Page#repeatable?

Trello card: https://trello.com/c/3nMdZ3Pa/1676-13-implement-add-another-answer-in-forms-runner-for-single-question-only-13

This PR makes changes to the Pages factory and adds a new test for an attribute added by an API change:

See: https://github.com/alphagov/forms-api/pull/542

The commit messages describe the changes to the Page factory and tests.

### DIFF
--- a/spec/models/page_spec.rb
+++ b/spec/models/page_spec.rb
@@ -6,6 +6,18 @@ RSpec.describe Page, type: :model do
     expect(page).to be_valid
   end
 
+  describe "#answer_settings" do
+    it "returns an empty object for answer_settings when it's not present" do
+      page = build :page
+      expect(page).to have_attributes(answer_settings: {})
+    end
+
+    it "returns an answer settings object for answer_settings when present" do
+      page = build :page, answer_settings: { only_one_option: "true" }
+      expect(page.answer_settings.attributes).to eq({ "only_one_option" => "true" })
+    end
+  end
+
   describe "API call" do
     let(:response_data) do
       {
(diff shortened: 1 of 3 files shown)


spec/models/page_spec.rb
thomasiles, line 1: This looks horrible in the diff but we are just wrapping the existing tests in a describe block, unchanged.

spec/models/page_spec.rb
```diff
@@ -1,49 +1,51 @@
 require "rails_helper"
 
 RSpec.describe Page, type: :model do
```
thomasiles, line 3: This looks horrible in the diff but we are just wrapping the existing tests in a describe block, unchanged.

spec/models/page_spec.rb
```diff
@@ -1,49 +1,64 @@
...
+      page = build :page, answer_settings: { only_one_option: "true" }
+      expect(page.answer_settings.attributes).to eq({ "only_one_option" => "true" })
     end
   end
 
-  it "returns the page for a form" do
-    expect(described_class.find(1, params: { form_id: 2 })).to have_attributes(id: 1, question_text: "Question text", answer_type: "date")
-  end
+  describe "repeatable?" do
+    it "when attribute does not exist" do
```
DavidBiddle, line 22: I think this and the two `it` blocks below should be named to say something about what they expect - currently I think they're written to sound more like context names:

```suggestion
    it "returns false when attribute does not exist" do
```

DavidBiddle (commented):
One suggestion but PR looks good otherwise. Appreciated the commit messages which explain the refactors really clearly 🎉

DavidBiddle (approved):

PR #877, opened 2024-08-12
Prepare FeatureService for form external IDs

stephencdaly:

Trello card: 

We are introducing external IDs for forms, and replacing exposing the form database ID outside of forms-api with using this exernal ID.

This means that the `form.id` in runner will become the external ID, which is a String rather than an integer.

Existing forms will have an external ID equal to the internal ID cast to a String. This means that if we have enabled a feature flag for individual forms using the `enabled_for_form_ids` configuration, this will continue to work without having to update the value of the environment variable in forms deploy. We do, however need to make sure we are treating the value as a string.

We can get ready for the form.id becoming a String by converting both the value of `enabled_for_form_ids` if it is just a single number to a string and also converting the form.id to a string for comparison.

### DIFF
--- a/app/services/feature_service.rb
+++ b/app/services/feature_service.rb
@@ -14,7 +14,9 @@ module FeatureService
     if feature.enabled_for_form_ids.present?
       raise FormRequiredError, "Feature #{feature_name} requires form to be provided" if form.blank?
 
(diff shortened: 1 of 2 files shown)


stephencdaly:
Closing - can be re-opened if needed as part of the work to use external IDs

PR #882, opened 2024-08-14
Add a single repeatable step

thomasiles:
# Enable a single repeatable question to be created

Trello card: https://trello.com/c/3nMdZ3Pa/1676-13-implement-add-another-answer-in-forms-runner-for-single-question-only-13

There is a lot to consider in this branch and I'm trying to do it with minimal disruption/refactoring of the current code.

### DIFF
--- a/app/controllers/forms/check_your_answers_controller.rb
+++ b/app/controllers/forms/check_your_answers_controller.rb
@@ -50,11 +50,17 @@ module Forms
   private
 
     def page_to_row(page)
+      change_link = if page.repeatable?
+                      change_add_another_answer_path(page.form_id, page.form_slug, page.page_id)
+                    else
+                      form_change_answer_path(page.form_id, page.form_slug, page.page_id)
+                    end
+
       question_name = helpers.question_text_with_optional_suffix_inc_mode(page, @mode)
       {
         key: { text: question_name },
         value: { text: page.show_answer },
-        actions: [{ href: form_change_answer_path(page.form_id, page.form_slug, page.page_id), visually_hidden_text: question_name }],
+        actions: [{ href: change_link, visually_hidden_text: question_name }],
       }
     end
 
--- a/app/controllers/forms/check_your_answers_controller.rb
+++ b/app/controllers/forms/check_your_answers_controller.rb
@@ -9,7 +9,7 @@ module Forms
     end
 
     def show
-      return redirect_to form_page_path(current_context.form.id, current_context.form.form_slug, current_context.next_page_slug) unless current_context.can_visit?(CheckYourAnswersStep::CHECK_YOUR_ANSWERS_PAGE_SLUG)
+      return redirect_to form_page_path(current_context.form.id, current_context.form.form_slug, current_context.next_page_slug, nil) unless current_context.can_visit?(CheckYourAnswersStep::CHECK_YOUR_ANSWERS_PAGE_SLUG)
 
       setup_check_your_answers
       email_confirmation_input = EmailConfirmationInput.new
--- a/app/controllers/forms/check_your_answers_controller.rb
+++ b/app/controllers/forms/check_your_answers_controller.rb
@@ -77,8 +77,7 @@ module Forms
     end
 
     def setup_check_your_answers
-      previous_step = current_context.previous_step("check_your_answers")
-      @back_link = form_page_path(current_context.form.id, current_context.form.form_slug, previous_step)
+      @back_link = back_link
       @rows = check_your_answers_rows
       @form_submit_path = form_submit_answers_path
 
@@ -87,6 +86,14 @@ module Forms
       end
 
       answers_need_full_width
+    end
+
+    def back_link
+      previous_step = current_context.previous_step(CheckYourAnswersStep::CHECK_YOUR_ANSWERS_PAGE_SLUG)
+
+      if previous_step.present?
+        previous_step.repeatable? ? add_another_answer_path(form_id: current_context.form.id, form_slug: current_context.form.form_slug, page_slug: previous_step.page_slug) : form_page_path(current_context.form.id, current_context.form.form_slug, previous_step.page_id)
+      end
     end
   end
 end
--- a/app/controllers/forms/page_controller.rb
+++ b/app/controllers/forms/page_controller.rb
@@ -37,6 +37,10 @@ module Forms
       page_slug = params.require(:page_slug)
       @step = current_context.find_or_create(page_slug)
 
+      if @step.respond_to?(:answer_id)
+        @step.answer_id = answer_id
+      end
+
       @support_details = current_context.support_details
     end
 
--- a/app/controllers/forms/page_controller.rb
+++ b/app/controllers/forms/page_controller.rb
@@ -51,6 +51,7 @@ module Forms
     def setup_instance_vars_for_view
       @is_question = true
       @question_edit_link = "#{Settings.forms_admin.base_url}/forms/#{@step.form_id}/pages/#{@step.page_slug}/edit/question"
+      @save_url = save_url
     end
 
     def changing_existing_answer
@@ -93,6 +94,10 @@ module Forms
 
     def is_first_page?
       current_context.form.start_page == @step.id
+    end
+
+    def save_url
+      save_form_page_path(@step.form_id, @step.form_slug, @step.id, changing_existing_answer: @changing_existing_answer, answer_id:)
     end
   end
 end
--- a/app/controllers/forms/page_controller.rb
+++ b/app/controllers/forms/page_controller.rb
@@ -70,10 +70,16 @@ module Forms
     end
 
     def next_page
-      if @changing_existing_answer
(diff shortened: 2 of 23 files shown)


config/routes.rb
```diff
@@ -22,9 +22,19 @@
       post "/#{CheckYourAnswersStep::CHECK_YOUR_ANSWERS_PAGE_SLUG}" => "forms/check_your_answers#submit_answers", as: :form_submit_answers
       get "/submitted" => "forms/submitted#submitted", as: :form_submitted
       get "/privacy" => "forms/privacy_page#show", as: :form_privacy
-      get "/:page_slug/change" => "forms/page#show", as: :form_change_answer, defaults: { changing_existing_answer: true }
-      get "/:page_slug" => "forms/page#show", constraints: { page_slug: Flow::StepFactory::PAGE_SLUG_REGEX }, as: :form_page
-      post "/:page_slug" => "forms/page#save", as: :save_form_page
+      get "/:page_slug/change(/:answer_id)" => "forms/page#show", as: :form_change_answer, defaults: { changing_existing_answer: true, answer_id: 1 }
```
lfdebrux, line 25:
```suggestion
      get "/:page_slug(/:answer_id)/change" => "forms/page#show", as: :form_change_answer, defaults: { changing_existing_answer: true, answer_id: 1 }
```

app/models/step.rb
```diff
@@ -24,59 +24,47 @@ def state
...
 
   def params
-    @question.attribute_names.concat([selection: []])
+    question.attribute_names.concat([selection: []])
   end
 
-  def valid?
-    @question.valid?
-  end
+  delegate :valid?, to: :question
```
lfdebrux, line 46:
```suggestion
  delegate :valid?, :show_answer, :show_answer_in_email, :question_text, :hint_text, :answer_settings, to: :question
```

app/models/repeatable_step.rb
```diff
@@ -0,0 +1,78 @@
...
+    super
+    @questions = []
+  end
+
+  def repeatable?
+    true
+  end
+
+  def parent_question
+    Step.instance_method(:question).bind(self).call
```
lfdebrux, line 18: This needs more explanation in the code

app/models/repeatable_step.rb
```diff
@@ -0,0 +1,78 @@
...
+
+  def next_answer_id
+    questions.length + 1
+  end
+
+  def show_answer
+    if questions.present?
+      safe_join(
+        questions.map.with_index(1) do |question, index|
+          content_tag(:p, sanitize("#{index}. #{question.show_answer}"))
```
lfdebrux, line 66: Can we make this an `<ol>`?

app/models/repeatable_step.rb
```diff
@@ -0,0 +1,78 @@
+class RepeatableStep < Step
+  include ActionView::Helpers::SanitizeHelper
+  include ActionView::Helpers::TagHelper
+  include ActionView::Helpers::OutputSafetyHelper
+
+  attr_accessor :answer_id, :questions
```
lfdebrux, line 6: Do we want to rename this to `index`?

```suggestion
  attr_accessor :answer_index, :questions
```

app/controllers/forms/page_controller.rb
```diff
@@ -37,12 +37,21 @@ def prepare_step
...
+      if @step.respond_to?(:answer_id)
+        @step.answer_id = answer_id
+      end
+
       @support_details = current_context.support_details
     end
 
+    def answer_id
+      params.fetch(:answer_id, 1).to_i
+    end
```
lfdebrux, line 49: I'm a bit nervous about the idea that non-repeatable pages have an `answer_id`

config/routes.rb
```diff
@@ -22,9 +22,19 @@
...
+
+      get "/:page_slug/add-another-answer" => "forms/add_another_answer#show", constraints: { page_slug: Flow::StepFactory::PAGE_SLUG_REGEX }, as: :add_another_answer
+      post "/:page_slug/add-another-answer" => "forms/add_another_answer#save", constraints: { page_slug: Flow::StepFactory::PAGE_SLUG_REGEX }, as: :save_add_another_answer
+      get "/:page_slug/change-add-another-answer" => "forms/add_another_answer#show", constraints: { page_slug: Flow::StepFactory::PAGE_SLUG_REGEX }, as: :change_add_another_answer, defaults: { changing_existing_answer: true }
```
lfdebrux, line 29:
```suggestion
      get "/:page_slug/add-another-answer/change" => "forms/add_another_answer#show", constraints: { page_slug: Flow::StepFactory::PAGE_SLUG_REGEX }, as: :change_add_another_answer, defaults: { changing_existing_answer: true }
```

app/controllers/forms/check_your_answers_controller.rb
```diff
@@ -82,5 +87,13 @@ def setup_check_your_answers
 
       answers_need_full_width
     end
+
+    def back_link
+      previous_step = current_context.previous_step(CheckYourAnswersStep::CHECK_YOUR_ANSWERS_PAGE_SLUG)
+
+      if previous_step.present?
+        previous_step.repeatable? ? add_another_answer_path(form_id: current_context.form.id, form_slug: current_context.form.form_slug, page_slug: previous_step.page_slug) : form_page_path(current_context.form.id, current_context.form.form_slug, previous_step.page_id)
+      end
```
lfdebrux, line 96: Can this be part of the flow context?

app/controllers/forms/page_controller.rb
```diff
@@ -51,18 +60,26 @@ def changing_existing_answer
...
+        else
+          check_your_answers_path(form_id: current_context.form.id, form_slug: current_context.form.form_slug)
+        end
       elsif @step.routing_conditions.any?
         calculate_page_routing
+      elsif @step.repeatable? && repeating?
+        add_another_answer_path(form_id: current_context.form.id, form_slug: current_context.form.form_slug, page_slug: @step.page_slug)
       else
         form_page_path(@step.form_id, @step.form_slug, @step.next_page_slug)
       end
```
lfdebrux, line 85: Can this be part of the flow context?

app/lib/flow/journey.rb
```diff
@@ -15,7 +15,7 @@ def initialize(form_context:, step_factory:)
     def find_existing_step(page_slug)
       step = @step_factory.create_step(page_slug)
       step.load_from_context(@form_context) unless @form_context.get_stored_answer(step).nil?
-    rescue ActiveModel::UnknownAttributeError
+    rescue ActiveModel::UnknownAttributeError, ArgumentError
```
lfdebrux, line 18: Can we create our own exceptions for these?

lfdebrux (commented):
Some suggestions for stylistic changes

app/controllers/forms/check_your_answers_controller.rb
thomasiles: yes, I think the flow context is the best place for back links and next_page.

I think it makes sense to do that as part of a wider refactor though. I think the context should return a step for the back link, which the controller then converts to a URL.

Keeping URLs out of the context will hopefully make it easier to test and a bit more flexible.

app/controllers/forms/page_controller.rb
thomasiles: yes, similar to the question about back_link

app/models/repeatable_step.rb
thomasiles: yes, I think that makes sense.

app/models/repeatable_step.rb
thomasiles: This does feel a little hacky. Can you think of a better name for it? It's kind of the prototype or base question for this Step but I'm not sure either of those make it clearer.

Would changing the constructor for RepeatableStep make it clearer?

```ruby
def initialize(question:, page:, form_id:, form_slug:, next_page_slug:, page_slug:)
super(question: question_prototype, page:, form_id:, form_slug:, next_page_slug:, page_slug:)
@questions = []
@question_protoype = question
```

lfdebrux: Is it better to do `.dup` on a question instance rather than `.new` on a class? Personally I'd suggest having an instance variable called `@question_class` and call `@question_class.new` when we want to create a new question.

Also, it just occurred to me, because we inherit from `Step` there should be an instance variable `@question` that we can use instead of calling the superclass?

```suggestion
    @question
```

We could use that in the constructor as well

```
def initialize(...)
  super
  @questions = []
  @question_protoype = @question # or @question_class = @question.class
end
```

thomasiles: Thats a good idea, I'll give the instance variable a go.

I started off using question.class.new. A new question doesn't include the values taken from the page to configure the question. Question text, hint, guidance etc all get configured when the question is created.

app/controllers/forms/page_controller.rb
lfdebrux: Let's leave this refactor for later

lfdebrux (approved):
Amazing stuff, thanks @thomasiles and @SamJamCul!

PR #896, opened 2024-08-23
1718 stop form fillers from adding more than x answers

SamJamCul:

Trello card: https://trello.com/c/qAJNULMm/1718-stop-form-fillers-from-adding-more-than-x-answers

When adding another answers, you should hit a limit that's set in `RepeatableStep::MAX_ANSWERS` now. This will alter the add_another_answer form so that it only submits with "no", thus continuing to the next question. 

This felt like the simplest way of getting users onto the next question, as it mimics them selecting 'no' to adding another answer, and removes the 'yes' option. 

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -124,6 +124,7 @@ en:
           other: You have added %{count} answers
           zero: You have added no answers
         radios_legend: Do you need to add another answer?
+        max_answers: You cannot add another answer to this question as you've entered the maximum of %{max_answers}
     back: Back
   helpers:
     hint:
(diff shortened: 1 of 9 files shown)


config/locales/en.yml
```diff
@@ -123,6 +124,7 @@ en:
           one: You have added one answer
           other: You have added %{count} answers
           zero: You have added no answers
+        max_answers: You cannot add another answer to this question as you've entered the maximum of %{max_answers}
```
lfdebrux, line 127:
```suggestion
        max_answers: You cannot add another answer to this question as you’ve entered the maximum of %{max_answers}
```

lfdebrux (dismissed):
Works for me, nice one!

lfdebrux (approved):

PR #900, opened 2024-08-29
WIP: Add remove answer

thomasiles:

Trello card: 


app/views/forms/remove_answer/show.html.erb
```diff
@@ -0,0 +1,103 @@
...
+        inline: true %>
+
+        <%= f.govuk_submit %>
+      <% when @step.question.is_a?(Question::Address) %>
+        <% set_page_title(form_title(form_name: @current_context.form.name, page_name: t('.address_heading'), mode: @mode, error: @step.question&.errors&.any?)) %>
+
+        <h1 class="govuk-heading-l"><%= t('.address_heading') %></h1>
+
+        <p class="govuk-inset-text">
+          <%= @step.question.show_answer %>
```
lfdebrux, line 49: Currently the address is not showing exactly as in the design:

Screenshot 2024-09-05 at 16 22 12

It's formatted as one line, rather than being multiple lines as in https://app.mural.co/t/gaap0347/m/gaap0347/1687946839058/9e1b89cd76af287c05c82df1cdb307a830324315?wid=0-1722946940424

How hard would this be to change?

lfdebrux (commented):
Some initial comments, so far it looks pretty good! I've been playing with it locally as well as reading the code.

config/locales/en.yml
```diff
@@ -127,6 +132,17 @@ en:
...
+      show:
+        address_heading: Remove an address
+        address_radios_legend: Are you sure you want to remove this address?
+        long_answer_heading: Remove an answer
+        long_radios_legend: Are you sure you want to remove this answer?
+        national_insurance_number_heading: Remove National Insurance number
+        national_insurance_number_radios_legend: Are you sure you want to remove this National Insurance number?
+        phone_number_heading: Remove phone number
+        phone_number_radios_legend: Are you sure you want to remove this phone number?
+        short_answer_legend: Are you sure you want to remove '%{answer_text}'?
```
lfdebrux, line 145:
```suggestion
        short_answer_legend: Are you sure you want to remove ‘%{answer_text’?
```

SamJamCul: Sorted!

config/routes.rb
```diff
@@ -51,6 +51,13 @@
            constraints: page_constraints,
            defaults: page_answer_defaults
 
+      get "/:page_slug/:answer_index/remove" => "forms/remove_answer#show",
+          as: :form_remove_answer,
+          constraints: page_constraints.merge(answer_constraints)
+      post "/:page_slug/:answer_index/remove" => "forms/remove_answer#delete",
```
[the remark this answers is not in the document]
SamJamCul: Updated this so that it's a delete method

app/controllers/forms/remove_answer_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+  class RemoveAnswerController < PageController
+    def show
+      @remove_answer_input = RemoveAnswerInput.new
+    end
+
+    def delete
+      @remove_answer_input = RemoveAnswerInput.new(remove_answer_input_params)
+
+      if @remove_answer_input.invalid?
+        return render :show
```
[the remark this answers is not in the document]
SamJamCul: Added this and updated test accordingly

app/views/forms/remove_answer/show.html.erb
SamJamCul: I think I might need help working out how to do this, I can't seem to inject any html or newline characters to try and split them out. It seems like we don't do this anywhere else either :/

lfdebrux: Okay, let's not worry about this for now then, we can just mark it down as a change we didn't make for this ticket and make a new one.

app/controllers/forms/add_another_answer_controller.rb
```diff
@@ -36,10 +36,16 @@ def add_another_path
 
     def rows
       @step.questions.map.with_index(1) do |question, answer_index|
+        actions = [{ text: t("forms.add_another_answer.rows.change"), href: form_change_answer_path(@step.form_id, @step.form_slug, @step.page_slug, answer_index:), visually_hidden_text: "answer #{answer_index}" }]
```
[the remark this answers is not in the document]
SamJamCul: Good idea, I'll try and splice it out

lfdebrux:
Not sure if you spotted this already, but I just found that with a form with an optional question that can be answered more than once, removing the answer from the check you answers page isn't working as I would expect.

What I did was filled out the form with all answers, then when I got to the check your answers page I clicked the change action for the optional question and then clicked the remove action for the sole answer and confirmed that I wanted to remove the answer.

What I expected to happen was to then be taken back to the check your answers page, or at least the add another answer page, but instead what happened was that I was taken to the question page for the repeatable question. If I didn't put an answer in the optional question I then had to go through the rest of the form again to get back to the check your answers page.

This screencast shows an example with a short form, but I've seen the same behaviour in a much longer form:

https://github.com/user-attachments/assets/f291d09d-a7fc-4420-b51f-b09dc4ab6122

SamJamCul:
> Not sure if you spotted this already...
@lfdebrux We've now added the `changing_existing_answer` param to all the relevant links, so it should be maintained while flicking through the remove answer pages.

lfdebrux (approved):

PR #901, opened 2024-08-29
Separate name fields in csv

DavidBiddle:

Trello card: https://trello.com/c/PhJ9djZ7/1732-tweak-csv-output-so-the-name-fields-are-separated-as-expected

Separates out the name field into separate columns in thge CSV output. To do this it adds a new method `show_answer_in_csv` to the question model. This returns a hash with one key/value pair per column. 

Touches quite a few files but most of those are adding tests.

### DIFF
--- a/app/models/question/name.rb
+++ b/app/models/question/name.rb
@@ -50,7 +50,11 @@ module Question
     end
 
     def show_answer_in_email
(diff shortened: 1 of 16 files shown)


chao-xian (approved):
Thanks for adding the tests!

PR #906, opened 2024-09-02
Add space to some phone numbers in CSV output

DavidBiddle:

Trello card: https://trello.com/c/PoAbnOwy/1730-make-phone-number-show-correctly-in-csv-file

Adds logic to tweak the CSV output for phone numbers. This is to work around the default Excel behaviour, which is to interpret purely numeric phone numbers as numbers rather than strings when opening the CSV fileand format them without any leading zeroes at the beginning.

The CSV will now display the number with a space after the 5th digit for any numbers which are:
- entirely numeric (i.e. don't have any spaces, extensions or '+' characters), and
- begin with a zero

This will force Excel to interpret the numbers as strings, which will preserve their formatting.

### DIFF
--- a/spec/models/question/phone_number_spec.rb
+++ b/spec/models/question/phone_number_spec.rb
@@ -8,13 +8,21 @@ RSpec.describe Question::PhoneNumber, type: :model do
   let(:is_optional) { false }
   let(:question_text) { Faker::Lorem.question }
 
-  let(:valid_phone_numbers) do
+  let(:valid_numeric_phone_numbers) do
+    %w[
+      07123123123
+      457123123123
+      07123123
+      071231231231231
+    ]
+  end
+
+  let(:valid_non_numeric_phone_numbers) do
     [
       "+447123 123 123",
       "+407123 123 123",
       "+1 7123 123 123",
       "+447123123123",
-      "07123123123",
       "01234 123 123 --()+ ",
       "01234 123 123 ext 123",
       "01234 123 123 x123",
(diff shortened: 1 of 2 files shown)


spec/models/question/phone_number_spec.rb
```diff
@@ -74,9 +86,27 @@
       expect(question.show_answer).to eq "07123123123"
     end
 
-    it "shows the answer in show_answer_in_csv" do
-      question.phone_number = "07123123123"
-      expect(question.show_answer_in_csv).to eq(Hash[question_text, "07123123123"])
+    context "when the phone number has only numeric characters and begins with 0" do
```
stephencdaly, line 89: Is it worth adding another test for when the phone number already has a space before the 5th digit? In this case we won't insert an additional space. Just so that the tests document this behaviour.

stephencdaly (approved):
Looks good to me. We agreed on slack that whilst it might be desirable to format UK numbers depending on their type https://www.area-codes.org.uk/formatting this is a bit more complex to do due to fact we're quite permissive with the phone numbers we allow meaning we can't use an existing gem easily.

For now, inserting a space just to stop spreadsheet software from dropping the leading 0 is the intention, and we can get feedback on whether form processors want more advanced formatting.

I've just left a suggestion for an additional test we might want to add.

spec/models/question/phone_number_spec.rb
DavidBiddle: Good shout, have added!

stephencdaly: Ah I meant, if the phone number is e.g. "0113 496 0000", where the space is after the 4th digit, it will keep the number how it is.

DavidBiddle: Ah, that makes more sense! Have updated

PR #914, opened 2024-09-04
Limit Notify attachment filename to 100 characters

stephencdaly:

Trello card: 

Notify has a limit of 100 characters for filenames for files attached to emails. We include the form name in the filename, which caused an exception to be thrown for long form names that would cause the filename length to exceed 100 characters.

Truncate the filename so it will be less than or equal to 100 characters. Truncate the filename at the last full word in the form name rather than partway through a word.


DavidBiddle (approved):
Code looks good and works on my machine 🎉

PR #918, opened 2024-09-05
Retry sending submission email without file upon error

stephencdaly:

Trello card: https://trello.com/c/myozrQtS/1743-bug-notify-error-due-to-csv-filename-exceeding-100-characters

When the NotificationClient raises a `Notifications::Client::BadRequestError` upon attempting to deliver the submission email when a CSV file is attached, try the request again without the CSV file attached.

The intention of this is to allow submissions to go through if there is an unexpected error with the CSV file attachment. This could be due to the file failing a virus scan, or another unanticipated issue with the file we have attached.

Send the exception to Sentry and log at error level when this happens, as we want to be notified if we fail to attach the CSV file.

### DIFF
--- a/app/services/form_submission_service.rb
+++ b/app/services/form_submission_service.rb
@@ -40,19 +40,13 @@ private
       raise StandardError, "Form id(#{@form.id}) is missing a submission email address"
     end
 
-    csv_attached = false
     unless @form.submission_email.blank? && @preview_mode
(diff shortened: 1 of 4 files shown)


DavidBiddle (approved):
Code looks good to me, and it works on my machine when I comment out the file name truncation.

Good idea on updating the configs!

jamie-o-wilkinson (approved):

PR #927, opened 2024-09-11
Update add another title and error text

thomasiles:
### Update the text for the title and error message of the add another page

Trello card: https://trello.com/c/CzswSAoa/1725-3-review-form-filler-content-for-single-repeatable-step

This PR updates the content for the add another page.

## Update
[Charlotte has said that the text should be "Select yes if you need to add another answer" and that other instances of this pattern should be updated to remove the capitalisation and quotes around yes](https://trello.com/c/CzswSAoa/1725-3-review-form-filler-content-for-single-repeatable-step#comment-66eaa040002be7b1a7af359e). I've changed the text for this error and the one which appears on the submission email form.

Updated screenshots of both:

image

image

## Outdated - see above

**Note** the trello card gave the text as "Select yes if you need to add another answer". I've capitalised and add quotes around _yes_. This makes it match the error the form filler sees if they don't answer the question about confirmation emails before submitting the form.

It's easy to change it if thats not what we want.

image

image


DavidBiddle (approved):
Good spot, code looks good, commits are well documented and it works on my machine 🎉

lfdebrux (approved):

PR #942, opened 2024-09-17
Add deprecation warnings for API endpoints that we think we don't use

lfdebrux:

Trello card: https://trello.com/c/7unK4mxB/1775-53dev-iterate-designs-for-single-repeatable-question-mvp-following-ur 

We want to be certain that forms-runner is not using any of the `/forms/:id`, `/forms/:id/pages`, or `/forms/:form_id/pages/:id` API endpoints.

This is because we plan to deprecate those endpoints as part of moving to version 2 of the API (see [ADR 034]).

This PR adds a deprecation warning to the `.find` methods for the `Page` and `Form` class.

Note that this also causes a warning when `Form.find#pages` is used if the form response doesn't include the pages; however apart from one explicit test I don't think we call API endpoints that don't include the pages (but we'll find out when we release this change!).

[ADR 034]: https://github.com/alphagov/forms/blob/5469de94bbd7e657a0b888608e4549c9ba30087a/ADR/ADR034-api-v2-for-forms.md

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -19,4 +19,10 @@ class ErrorsController < ApplicationController
   def maintenance
     render "errors/maintenance", formats: :html
   end
+
+  def deprecated
+    form = Form.find params.require(:form_id)
+    page = Page.find params.require(:page_id), params: { form_id: params.require(:form_id) }
+    redirect_to form_page_path(form_id: form.id, form_slug: form.form_slug, page_slug: page.id, mode: :form)
+  end
 end
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -6,6 +6,17 @@ class Form < ActiveResource::Base
 
   has_many :pages
 
+  def self.deprecator
+    Rails.application.deprecators[:forms_api]
+  end
+
+  def self.find(scope, **options)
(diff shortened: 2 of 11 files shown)


lfdebrux:
I've tested this in dev by visiting submit.dev.forms.service.gov.uk/deprecated?form_id=71&page_id=159.

The deprecation warning message appears in Splunk (see https://gds.splunkcloud.com/en-GB/app/gds-543-forms/search?q=search%20index%3D%22gds_dsp_dev_forms%22%20DEPRECATION&display.page.search.mode=smart&dispatch.sample_ratio=1&workload_pool=&earliest=1726635934.262&latest=1726635944.263&sid=1726636301.719842), but more importantly I don't see any deprecation warnings from running the end to end tests.

thomasiles (approved):
Looks good!

The deprecations are reassuring.

As an added check, I've used [this splunk query](https://gds.splunkcloud.com/en-GB/app/gds-543-forms/search?sid=1726656683.26340) in the past:

```
index="*production*" | search requested_by="forms-runner*" | stats values(action)
```

Which outputs:
```
show_archived
show_draft
show_live
```

to give a bit of reassurance that the runner is only using what we think it is.

PR #946, opened 2024-09-18
Allow using v2 API to load forms

lfdebrux:

Trello card: https://trello.com/c/OOYCWoTa/1779-8-change-forms-runner-to-use-v2-api 

This PR makes it so that if we enable the `api_v2` feature flag forms-runner uses the new v2 API.

Change `Api::V1::FormSnapshotRepository` to use new `Api::V2::FormDocumentResource` and `Api::V1::Converter` classes to allow this app to retrieve forms using the v2 API without changes to the rest of the codebase. Basically, we get the v2 form documents and then immediately convert them to the older form snapshot format so we can keep using the same models.

This behaviour is currently behind a feature flag and not enabled, so it should be safe to deploy this change to production.

I have tested the change locally with the feature flag enabled and the end to end tests pass, however the unit tests need more work to pass, which I will do in a future PR.


config/settings/development.yml
```diff
@@ -0,0 +1,2 @@
+features:
+  api_v2: true
```
thomasiles, line 2: I think this is a typo - it should be `v2_api`.

lfdebrux: Good spot, thank you

lfdebrux: Actually I might change the feature flag to be properly named `api_v2`, so it's consistent with other places in our codebase...

lfdebrux: I went ahead and changed it, the feature flag should now be `api_v2` everywhere

thomasiles (approved):
Sorry I meant to leave the rest of the review beyond the typo yesterday but I got distracted.

I really like this approach, it's very clever! It really reduces the risks and allows us to migrate the runner incrementally.

My only comment was going to be that some sort of end to tests to cover the whole new journey would be useful as a sanity check that nothing has broken.

I've seen you already have another PR which covers it.

I tested it yesterday without the changes to the name of the feature flag and didn't find any issues.

Amazing 🏅

app/services/api/v1/converter.rb
```diff
@@ -0,0 +1,26 @@
+class Api::V1::Converter
+  def to_api_v1_form_snapshot(form_document)
```
chao-xian, line 2: just a style note here: the `to_api_` prefix is superfluous as the inference that this is the `API::V1::Converter` class so a method that gives `form_snapshot(form_document)` would have been more succinct - similar for `page(step)` below. Or for a bit of clarity `v1_form_snapshot` would be ok too. Nbd as I understand this class wont be around too long.

PR #947, opened 2024-09-19
Use submission type on form

jamie-o-wilkinson:

Trello card: https://trello.com/c/DQCipW8q

This PR refactors the form submission service to use the `submission_type` stored on the Form model instead of the per form CSV submission feature flag. Since the per form feature flagging is no longer needed and we recommend not using it in future, we've removed it from the feature service.

We'll remove associated terraform from forms-deploy in a subsequent PR.

### DIFF
--- a/app/services/feature_service.rb
+++ b/app/services/feature_service.rb
@@ -1,26 +1,10 @@
-module FeatureService
-  class FormRequiredError < StandardError; end
+class FeatureService
+  class << self
+    def enabled?(feature_name)
+      return false if Settings.features.blank?
 
-  def self.enabled?(feature_name, form = nil)
-    return false if Settings.features.blank?
-
-    segments = feature_name.to_s.split(".")
-    feature = Settings.features.dig(*segments)
-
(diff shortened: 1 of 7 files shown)


app/services/feature_service.rb
```diff
@@ -1,26 +1,10 @@
...
-    if feature.enabled_for_form_ids.present?
-      raise FormRequiredError, "Feature #{feature_name} requires form to be provided" if form.blank?
-
-      return feature.enabled_for_form_ids == form.id if feature.enabled_for_form_ids.is_a?(Integer)
-
-      form_ids = feature.enabled_for_form_ids.split(",").collect(&:strip)
-
-      return form_ids.include?(form.id.to_s)
+      segments = feature_name.to_s.split(".")
+      segments.reduce(Settings.features) { |config, segment| config[segment] }
```
lfdebrux, line 7: Let's use `dig` instead of `reduce` here:

```suggestion
      Settings.features.dig(*segments)
```

lfdebrux (approved):
LGTM

PR #963, opened 2024-09-25
Support repeated answers in submission CSV

stephencdaly:

Trello card: https://trello.com/c/n6Jpj46Q/1822-support-add-another-single-answer-into-csv-processor

When there is a repeated step in the completed form - i.e. when a question that allows multiple answers has been answered multiple times, include a column in the submission CSV for each repeated answer. The column header is suffixed by the answer index, like ' - Answer 1'.

For example,

'What is the meaning of life - Answer 1, What is the meaning of life - Answer 2'

When the question has separate attributes for different parts of the answer, like a name type question can, the suffix comes after the suffix for the attribute name.

For example,

'What is your name - First name - Answer 1, What is your name - Last name - Answer 1'

### DIFF
--- a/spec/factories/models/question/text.rb
+++ b/spec/factories/models/question/text.rb
@@ -1,0 +1,14 @@
+FactoryBot.define do
+  factory :text, class: "Question::Text" do
+    question_text { Faker::Lorem.question }
+    hint_text { nil }
+    is_optional { false }
+    page_heading { nil }
+    guidance_markdown { nil }
+    text { nil }
+
+    trait :with_answer do
+      text { Faker::Lorem.sentence }
+    end
+  end
+end
--- a/spec/factories/models/repeatable_steps.rb
+++ b/spec/factories/models/repeatable_steps.rb
@@ -1,0 +1,12 @@
+FactoryBot.define do
+  factory :repeatable_step, class: "RepeatableStep" do
+    page { association :page }
+    sequence(:page_slug) { |n| "page-#{n}" }
+    sequence(:form_slug) { |n| "form-#{n}" }
+    form_id { 1 }
+    question { build(:full_name_question) }
+    next_page_slug { nil }
+
+    initialize_with { new(question:, page:, form_id:, form_slug:, next_page_slug:, page_slug:) }
+  end
+end
--- a/app/models/repeatable_step.rb
+++ b/app/models/repeatable_step.rb
@@ -85,6 +85,18 @@ class RepeatableStep < Step
     end
   end
 
+  def show_answer_in_csv
+    if questions.present?
+      header_values_hash = {}
+      questions.each.with_index(1) do |question, index|
+        question.show_answer_in_csv.each do |header, value|
+          header_values_hash["#{header} - Answer #{index}"] = value
+        end
+      end
+      header_values_hash
+    end
+  end
+
   def remove_answer(answer_index)
     questions.delete_at(answer_index - 1)
     if questions.empty?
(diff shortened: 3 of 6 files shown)


spec/factories/models/repeatable_steps.rb
```diff
@@ -0,0 +1,14 @@
...
+    page { association :page }
+    sequence(:page_slug) { |n| "page-#{n}" }
+    sequence(:form_slug) { |n| "form-#{n}" }
+    form_id { 1 }
+    question { build(:full_name_question) }
+    next_page_slug { nil }
+    questions { build_list(:full_name_question, 2) }
+    answer_index { 1 }
+
+    initialize_with { new(question:, page:, form_id:, form_slug:, next_page_slug:, page_slug:) }
```
lfdebrux, line 12: Why do we need to define `initialize_with`, doesn't the default work?

spec/factories/models/repeatable_steps.rb
```diff
@@ -0,0 +1,14 @@
+FactoryBot.define do
+  factory :repeatable_step, class: "RepeatableStep" do
+    page { association :page }
+    sequence(:page_slug) { |n| "page-#{n}" }
+    sequence(:form_slug) { |n| "form-#{n}" }
+    form_id { 1 }
+    question { build(:full_name_question) }
+    next_page_slug { nil }
+    questions { build_list(:full_name_question, 2) }
+    answer_index { 1 }
```
lfdebrux, line 10: Setting `questions` and `answer_index` feels like something that should be part of a trait or sub-factory rather than the default factory, as those are things that are normally managed by the class initializer.

spec/factories/models/question/text.rb
```diff
@@ -0,0 +1,10 @@
+FactoryBot.define do
+  factory :text, class: "Question::Text" do
+    question_text { Faker::Lorem.question }
+    hint_text { nil }
+    is_optional { false }
+    page_heading { nil }
+    guidance_markdown { nil }
+    text { Faker::Lorem.sentence }
```
lfdebrux, line 8: It feels like the default for `text` should be `nil`?

spec/factories/models/repeatable_steps.rb
stephencdaly: Without it, it fails - due to the initialise method having keyword parameters.

spec/factories/models/repeatable_steps.rb
stephencdaly: Yeh makes sense, I can move to traits

spec/factories/models/question/text.rb
stephencdaly: Yeh, makes sense - I'll add a trait for having an answer

spec/factories/models/repeatable_steps.rb
stephencdaly: I've just set them to nil as my tests were setting the questions themselves anyway so I don't need a trait for now.

lfdebrux: Ah, can we just remove the lines altogether? I am just concerned about the factory somehow overriding the implementation behaviour which is to set `question = [@question.dup]` in https://github.com/alphagov/forms-runner/blob/8dc52ef014d535847b83b403d8181299408d8ee8/app/models/repeatable_step.rb#L14

stephencdaly: Ah gotcha, yeh sure

stephencdaly: Done

lfdebrux (approved):

PR #965, opened 2024-09-25
Display optional questions in CSV

stephencdaly:

Trello card: https://trello.com/c/n6Jpj46Q

This PR makes it so that we always include columns for optional questions in the submission CSV, with a blank value if they are unanswered.

For optional questions that can have multiple answers, if no answer is given, we want to include columns for the first answer, with the suffix " - Answer 1" and with blank value, to mirror what we do in the email.

### DIFF
--- a/app/models/question/name.rb
+++ b/app/models/question/name.rb
@@ -54,7 +54,22 @@ module Question
     end
 
     def show_answer_in_csv
-      attributes_with_values.to_h { |attribute| ["#{question_text} - #{friendly_name_for_attribute(attribute)}", send(attribute)] }
+      header_values_hash = {}
+      if needs_title?
+        add_attribute_to_hash(:title, header_values_hash)
+      end
+
+      if is_full_name?
+        add_attribute_to_hash(:full_name, header_values_hash)
+      else
+        add_attribute_to_hash(:first_name, header_values_hash)
+        if include_middle_name?
+          add_attribute_to_hash(:middle_names, header_values_hash)
+        end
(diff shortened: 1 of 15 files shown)


app/models/question/name.rb
```diff
@@ -54,7 +54,22 @@ def show_answer_in_email
...
+
+      if is_full_name?
+        header_values_hash[csv_header_name_for_attribute(:full_name)] = full_name.to_s
+      else
+        header_values_hash[csv_header_name_for_attribute(:first_name)] = first_name.to_s
+        if include_middle_name?
+          header_values_hash[csv_header_name_for_attribute(:middle_names)] = middle_names.to_s
+        end
+        header_values_hash[csv_header_name_for_attribute(:last_name)] = last_name.to_s
+      end
```
chao-xian, line 70: Can we extract this to a method pls? Kinda like https://refactoring.guru/consolidate-conditional-expression

stephencdaly: I'm not entirely sure which bit you mean, as the linked example isn't talking about quite the same thing. Which bits would you suggest extracting?

stephencdaly: I've pushed up a minor simplification to extract e.g. 

```
header_values_hash[csv_header_name_for_attribute(:last_name)] = last_name.to_s
```

to 

```
add_attribute_to_hash(:last_name, header_values_hash)
```

Are you suggesting to extract everything within the else block to a method?

chao-xian (approved):

PR #993, opened 2024-10-15
Remove code using V1 API

lfdebrux:

We are no longer using the V1 API code anywhere, and are happy with the functioning of the V2 API for forms-runner. We can delete this code.


DavidBiddle (approved):
Code looks good and no issues when running locally 🎉

PR #1003, opened 2024-10-18
Change Step to keep a reference to its associated models

lfdebrux:

Steps belong to forms and have a page. Currently we keep the IDs of these associated objects as attributes in a step, along with some other bits of data from those records that we need, but this isn't super helpful if we need other details about those other objects.

Given that in forms-runner we will always fetch the form and build the question before instantiating a step, we can just keep references as an instance attribute, with little downside.


thomasiles (approved):
looks neater, nice one!

PR #1004, opened 2024-10-18
Implement submission to s3

stephencdaly:

Trello card: https://trello.com/c/ToXPf8zb/

Implement uploading file containing submission data to an S3 bucket specified for a form if the `submission_type` is `s3`. 

To do this, we assume an AWS role that is defined per environment. We expect the bucket to have granted write permissions for this AWS role.

### How to test

1. Update `aws_s3_submissions.iam_role_arn` in the settings.yml file to be the ARN of the role with permissions to upload to the s3 bucket in the deploy environment.
2. Run the forms-api rake task to configure one of your local forms to send submissions to S3:

```
rake "forms:set_submission_type_to_s3[1, deploy-account-bucket-name, deploy-account-id]"
```

3. Start form-runner locally in an authenticated aws shell for the dev environment. This allows the app to assume the AWS role for writing to the S3 bucket:

```
gds aws forms-dev-readonly -- bundle exec rails s
```

4. Submit the form. Check the S3 bucket in the deploy account and verify that a submission CSV file has been uploaded.

### DIFF
--- a/app/services/form_submission_service.rb
+++ b/app/services/form_submission_service.rb
@@ -91,7 +91,7 @@ private
 
   def deliver_submission_email_with_csv_attachment
     Tempfile.create do |file|
-      SubmissionCsvService.new(
+      CsvSubmissionService.new(
         current_context: @current_context,
         submission_reference: @submission_reference,
         timestamp: @timestamp,
--- a/app/services/form_submission_service.rb
+++ b/app/services/form_submission_service.rb
@@ -39,16 +39,20 @@ private
     if @form.submission_type == "s3"
       upload_submission_csv_to_s3
     else
-      if !@preview_mode && @form.submission_email.blank?
-        raise StandardError, "Form id(#{@form.id}) is missing a submission email address"
-      end
+      submit_form_with_email_submission_type
+    end
+  end
 
-      unless @form.submission_email.blank? && @preview_mode
-        if @form.submission_type == "email_with_csv"
-          deliver_submission_email_with_csv_attachment_with_fallback
-        else
-          deliver_submission_email
-        end
+  def submit_form_with_email_submission_type
+    if !@preview_mode && @form.submission_email.blank?
+      raise StandardError, "Form id(#{@form.id}) is missing a submission email address"
+    end
+
+    unless @form.submission_email.blank? && @preview_mode
+      if @form.submission_type == "email_with_csv"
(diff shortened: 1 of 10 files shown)


app/services/s3_submission_service.rb
```diff
@@ -0,0 +1,45 @@
...
+  def initialize(file_path:,
+                 form_id:,
+                 s3_bucket_name:,
+                 s3_bucket_aws_account_id:,
+                 timestamp:,
+                 submission_reference:)
+    @file_path = file_path
+    @form_id = form_id
+    @bucket = s3_bucket_name
+    @expected_bucket_owner = s3_bucket_aws_account_id
```
chao-xian, line 14: How about `receiving_bucket_owner`?

app/services/form_submission_service.rb
```diff
@@ -36,15 +36,19 @@ def submit
...
-        deliver_submission_email_with_csv_attachment_with_fallback
-      else
-        deliver_submission_email
+      unless @form.submission_email.blank? && @preview_mode
+        if @form.submission_type == "email_with_csv"
+          deliver_submission_email_with_csv_attachment_with_fallback
+        else
+          deliver_submission_email
+        end
       end
```
DavidBiddle, line 52: Optional: The logic in this method has quite a bit of nested logic. It might make sense to move the email-related logic into its own method, so that the `submit_form_to_processing_team` is only responsible for choosing the correct submission method.

DavidBiddle (dismissed):
Code looks good and commit messages are clear :tada:

I've raised one optional comment, but will leave it up to you whether you think it's worth doing.

I haven't managed to test this locally as I've not got s3 bucket details to use at the moment

app/services/form_submission_service.rb
chao-xian: Yeah I was going to say similar and that we should at least refactor in a separate PR to improve/clean up the code.

stephencdaly: Great suggestion, have extracted this to a method. I'll spend a little bit of time looking at this more to see if there's a slightly bigger refactor that could be done without too much effort - but if I do that it would be in another PR.

DavidBiddle (approved):

chao-xian:
I also don't have a bucket to test with so maybe one for @alice-carr to double check?

alice-carr (approved):
I've tested this locally and it's all working as expected with the submission coming through to the bucket 👍

PR #1011, opened 2024-10-22
Add tests for Step model

thomasiles:

To add support for branching questions with two branches to the Forms Runner, the #next_page_after_routing method needs to be changed.

This PR adds a spec file for the Step model.

Trello card: https://trello.com/c/vKfjfFSA/1922-add-support-for-branching-question-with-two-branches-to-forms-runner

### Add tests to make room for future improvements

The Step model doesn't currently have any unit tests.

This makes changing the class harder:
- it's tricky to know we haven't broken something the code relies on
- new methods are less likely to have tests written for them if there is no existing spec file to add to

This commit adds a spec file for the Step model.

The tests are written to match the existing code and the functionality it currently has.

There are ways we could improve the Step model in the future which would make the tests easier to write and maintain.

### Future improvements to the model and tests

These tests will be the base on which improvements can be made.

For example, we could:
- make some methods in the Step model private
- remove any methods which are not used in the code
- improve the way the Step model is instantiated to make it easier to
  configure and test
- add a model for conditions, which is currently an implicit class
  created by ActiveResource
- clear up the factories to make them easier to use in tests, rather
  than using doubles

### DIFF
--- a/spec/models/step_spec.rb
+++ b/spec/models/step_spec.rb
@@ -1,0 +1,256 @@
+require "rails_helper"
+
+RSpec.describe Step do
+  subject(:step) do
+    described_class.new(
+      question:,
+      page:,
+      form_id: 3,
+      form_slug: "test-form",
+      next_page_slug: "next-page",
+      page_slug: "current-page",
+    )
+  end
+
+  let(:question) { instance_double(Question::Text, serializable_hash: {}, attribute_names: %w[name], valid?: true) }
+  let(:page) { build(:page, id: 2, position: 1, routing_conditions: []) }
+  let(:form_context) { instance_double(Flow::FormContext) }
+
+  describe "#initialize" do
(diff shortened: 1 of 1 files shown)


spec/models/step_spec.rb
```diff
@@ -0,0 +1,256 @@
...
+      expect(step.page_slug).to eq("current-page")
+      expect(step.form_id).to eq(3)
+      expect(step.form_slug).to eq("test-form")
+      expect(step.next_page_slug).to eq("next-page")
+      expect(step.page_number).to eq(1)
+      expect(step.routing_conditions).to eq([])
+    end
+  end
+
+  describe "#==" do
```
SamJamCul, line 32: What's that?

SamJamCul (approved):
Looking good, this was a real thorn in the runner

PR #1017, opened 2024-10-24
Refactor form submission service

stephencdaly:

We've recently added a fair bit of additional logic to FormSubmissionService to deal with attaching CSV files to emails and also sending submissions to an S3 bucket. Refactor FormSubmissionService and related services to attempt to divide up the concerns into more manageable classes. 

### DIFF
--- a/app/services/form_submission_service.rb
+++ b/app/services/form_submission_service.rb
@@ -120,12 +120,12 @@ private
   end
 
   def write_submission_csv(file)
-    CsvGenerator.new(
+    CsvGenerator.write_submission(
       current_context: @current_context,
       submission_reference: @submission_reference,
       timestamp: @timestamp,
       output_file_path: file.path,
-    ).write
+    )
   end
 
   def form_title
--- a/app/services/form_submission_service.rb
+++ b/app/services/form_submission_service.rb
@@ -120,7 +120,7 @@ private
   end
 
   def write_submission_csv(file)
-    CsvSubmissionService.new(
+    CsvGenerator.new(
       current_context: @current_context,
       submission_reference: @submission_reference,
       timestamp: @timestamp,
--- a/app/services/form_submission_service.rb
+++ b/app/services/form_submission_service.rb
@@ -16,45 +16,31 @@ class FormSubmissionService
     @timestamp = submission_timestamp
     @submission_reference = ReferenceNumberService.generate
 
-    @mailer_options = MailerOptions.new(title: form_title,
-                                        preview_mode: @preview_mode,
-                                        timestamp: @timestamp,
-                                        submission_reference: @submission_reference,
-                                        payment_url: @form.payment_url_with_reference(@submission_reference))
-
     CurrentLoggingAttributes.submission_reference = @submission_reference
   end
 
   def submit
     submit_form_to_processing_team
     submit_confirmation_email_to_user
+
     @submission_reference
   end
 
 private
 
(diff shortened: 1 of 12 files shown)


spec/services/notify_submission_service_spec.rb
```diff
@@ -0,0 +1,196 @@
...
+  let(:service) { described_class.new(current_context:, notify_email_reference:, mailer_options:) }
+  let(:form) do
+    build(:form,
+          id: 1,
+          name: "Form 1",
+          submission_email:,
+          payment_url:)
+  end
+  let(:current_context) { OpenStruct.new(form:, completed_steps: [step], support_details: OpenStruct.new(call_back_url: "http://gov.uk")) }
+  let(:request) { OpenStruct.new({ url: "url", method: "method" }) }
```
lfdebrux, line 13: This variable never appears to be used? 🤔

lfdebrux (approved):
These changes look good to me, and the end to end tests pass on my machine. I have two tiny comments, but nothing important.

Thanks for taking the time to do this!

spec/services/s3_submission_service_spec.rb
```diff
@@ -2,96 +2,98 @@
...
+  end
   let(:s3_bucket_name) { "a-bucket" }
   let(:s3_bucket_aws_account_id) { "23423423423423" }
   let(:s3_bucket_region) { "eu-west-1" }
   let(:submission_reference) { Faker::Alphanumeric.alphanumeric(number: 8).upcase }
   let(:timestamp) do
     Time.use_zone("London") { Time.zone.local(2022, 9, 14, 8, 0o0, 0o0) }
   end
+  let(:current_context) { OpenStruct.new(form:, completed_steps: [step]) }
+  let(:step) { OpenStruct.new({ question_text: "What is the meaning of life?", show_answer_in_email: "42" }) }
```
lfdebrux, line 24: One more thing to consider, if you have time, I wonder if it would be possible to use `instance_double` instead of `OpenStruct` mocks in these tests; the advantage there is that `rspec-mocks` will then be able to check whether we're accidentally stubbing methods that don't exist.

app/services/form_submission_service.rb
```diff
@@ -16,126 +16,70 @@ def initialize(current_context:, email_confirmation_input:, preview_mode:)
...
     else
-      submit_form_with_email_submission_type
-    end
-  end
-
-  def submit_form_with_email_submission_type
-    if !@preview_mode && @form.submission_email.blank?
-      raise StandardError, "Form id(#{@form.id}) is missing a submission email address"
+      notify_submission_service.submit_form_to_processing_team
     end
```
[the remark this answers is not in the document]
stephencdaly: Really good suggestion, have done this.

spec/services/notify_submission_service_spec.rb
stephencdaly: Good spot, have removed

spec/services/s3_submission_service_spec.rb
stephencdaly: Agreed, use of OpenStruct here isn't very nice. I've replaced this to use the factories instead - does that seem ok to you?

lfdebrux (approved):

spec/services/s3_submission_service_spec.rb
lfdebrux: Yes using factories is great I think!

PR #1018, opened 2024-10-24
Add default condition support to routing

thomasiles:
### Add default to Step#next_page_slug_after_routing

Trello card: https://trello.com/c/vKfjfFSA/1922-add-support-for-branching-question-with-two-branches-to-forms-runner

To support more complex routing in forms, form creators will be able to add a "default" condition to a page to cover the case when the user does not select an option.

This commit allows the second condition to be set as the default. To do so the condition must have an empty answer value.

The code change is maybe overly strict about conditions and the order they must be in.

It's just as easy to allow multiple conditions with a default, for example:

```ruby
  def next_page_slug_after_routing
     return next_page_slug if routing_conditions.empty?

    matching_condition = find_matching_condition
    default_condition = find_default_condition

    if matching_condition
      goto_condition_page_slug(matching_condition)
    elsif default_condition
      goto_condition_page_slug(default_condition)
    else
      next_page_slug
    end
  end

  def find_matching_condition
    routing_conditions.find { |condition| condition.answer_value == question.selection }
  end

  def find_default_condition
    routing_conditions.find { |condition| condition.answer_value.blank? }
  end
  ```

In order to keep this change focused on the current problem, the code only handles the current case of two conditions with a default.

When changes to the admin are complete it might be clearer how we will represent different routing conditions.

The routing conditions should probably be given their own model and the evaluation logic might be moved to it's own class, outside of the service.

Tackling those at this point is probably too early.


lfdebrux (approved):
🙌

PR #1019, opened 2024-10-25
Change key name for submission uploaded to s3

stephencdaly:

Trello card: https://trello.com/c/ToXPf8zb/

Change the key name so that:

- The root folder is pluralized to "form_submissions/"
- If the form is a preview, the root folder is "test_form_submissions/"
- The form ID is a folder rather than a prefix to the filename
- The timestamp is in UTC and does not contain any special characters, as colons (:) and plus signs (+) are not in AWS's list of safe characters for key names.

### DIFF
--- a/app/services/s3_submission_service.rb
+++ b/app/services/s3_submission_service.rb
@@ -1,11 +1,13 @@
 class S3SubmissionService
   def initialize(current_context:,
                  timestamp:,
-                 submission_reference:)
+                 submission_reference:,
+                 preview_mode:)
     @current_context = current_context
     @form = current_context.form
     @timestamp = timestamp
     @submission_reference = submission_reference
+    @preview_mode = preview_mode
   end
 
   def submit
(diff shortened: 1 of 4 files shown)


app/services/s3_submission_service.rb
```diff
@@ -58,6 +60,8 @@ def assume_role
   end
 
   def key_name
-    "form_submission/#{@form.id}_#{@timestamp}_#{@submission_reference}.csv"
+    folder = @preview_mode ? "test_form_submissions" : "form_submissions"
+    formatted_timestamp = @timestamp.utc.strftime("%Y%m%dT%H%MZ")
```
lfdebrux, line 64: Is it worth having seconds as well?

stephencdaly: Oops yes I meant to and forgot to add that in after copying this from somewhere!

stephencdaly: Have updated, thanks!

lfdebrux (approved):

chao-xian (approved):

PR #1048, opened 2024-11-13
Add 'None of the above' option to autocomplete

DavidBiddle:

Trello card: https://trello.com/c/5prDL8Vy/1974-implement-none-of-the-above-pattern-for-long-list?filter=label:Feature%20team%20one,label:Tech%20Support,label:X%20feature%20teams

Includes:

- Does a bit of work to simplify our existing 'None of the above' logic (specifically, removing some coversion between strings and symbols)
- Adds 'None of the above' to the end of the autocomplete list if the question supports it.

We discussed adding synonyms for the new option, but this PR doesn't include that. This is because it's more complex than it initially appeared - it relies on the dfe-autocomplete library's custom sorting logic, which we're not using in our app at the moment. We can still do this if we want to, but there are some tradeoffs involved that mean we should make a card, refine it and implement it in a separate PR if synonyms are needed.

### Screenshot
An autocomplete question where the list of options now has 'None of the above' as the last entry in the list


app/components/question/selection_component/view.rb
```diff
@@ -36,14 +36,14 @@ def divider
       def none_of_the_above_radio_button
         return nil unless question.is_optional?
 
-        option = form_builder.govuk_radio_button :selection, :none_of_the_above.to_s, label: { text: "None of the above" }
+        option = form_builder.govuk_radio_button :selection, I18n.t("page.none_of_the_above"), label: { text: "None of the above" }
```
stephencdaly, line 39: Do you know why the label doesn't use a translation, isn't that the bit that it's most important to be able to translate?

stephencdaly: I also wonder whether conversely, we don't want to use a translation for the value. If we add translations for another language, wouldn't we still want the submission email to contain the English? This is a question for when we actually come to translate, probably.

stephencdaly (approved):
Looks good. I've tested locally and works as expected for radios, check boxes and autocomplete.

Just left a question about translations but it isn't a blocker.

stephencdaly (approved):
Nice work!

PR #1055, opened 2024-11-18
Add id to submission reference number

alice-carr:
When the submission type of a form is S3, we save the form to a specified bucket, using the submission reference number as part of the key. When running end to end tests, we want to be able to submit a form, then look into an S3 bucket to find the submission.

Adding an ID here allows the end to end tests to easily find the reference number and use it to look for the submitted form in S3.

[Trello card](https://trello.com/c/64CM50hZ/1921-add-an-e2e-test-for-s3-submissions)

this relates to this PR: https://github.com/alphagov/forms-e2e-tests

### DIFF
--- a/app/views/forms/submitted/_submitted_panel.html.erb
+++ b/app/views/forms/submitted/_submitted_panel.html.erb
@@ -7,6 +7,6 @@
 ) do %>
   <% if current_context.get_submission_reference.present? %>
     <%= t('form.submitted.your_reference') %><br>
-    <strong><%= current_context.get_submission_reference %></strong>
+    <strong id="submission-reference"><%= current_context.get_submission_reference %></strong>
   <% end %>
 <% end %>


app/views/forms/submitted/_submitted_panel.html.erb
```diff
@@ -7,6 +7,6 @@
 ) do %>
   <% if current_context.get_submission_reference.present? %>
     <%= t('form.submitted.your_reference') %><br>
-    <strong><%= current_context.get_submission_reference %></strong>
+    <strong id="app-panel--reference-number"><%= current_context.get_submission_reference %></strong>
```
stephencdaly, line 10: I'm not sure id's need to follow this pattern with `app-panel--` - I think that's more how we do class names but also no harm in it doing this either so not a blocker.

stephencdaly (approved):

app/views/forms/submitted/_submitted_panel.html.erb
stephencdaly: Yeh no pattern I can see, so this is fine to go in like this.

lfdebrux: The naming of `app-panel--` follows [Block Element Modifier style](https://getbem.com/), just for interest.

I'd recommend just calling the id here `submission-reference`, to match the method we're calling

```suggestion
    <strong id="submission-reference"><%= current_context.get_submission_reference %></strong>
```

alice-carr: thanks Laurence, I've updated that

lfdebrux (approved):

PR #1085, opened 2024-12-05
Upload to s3 for file upload questions

stephencdaly:

Trello card: https://trello.com/c/7CmjPY64/

Add a new `update_answer` method to question_base to set the answer attributes on the question. This is overridden by the `file` model to perform the upload to S3 when the answer is saved.

We do not want to save the file itself to the session, so we set this to `nil` on the model as this model is used to set the answer data on the session. Instead, set `original_filename` which is the name the file was uploaded with and `uploaded_file_key` which is the S3 object key on the session.

Make it so that we display the `original_filename` back to the user on the "Check your answers" page.

We generate a unique key for the file in S3 using UUID to ensure there is no chance of 2 files being uploaded with the same key.

### DIFF
--- a/app/models/question/file.rb
+++ b/app/models/question/file.rb
@@ -1,5 +1,43 @@
 module Question
   class File < Question::QuestionBase
     attribute :file
+    attribute :original_filename
+    attribute :uploaded_file_key
+
+    def show_answer
+      original_filename
+    end
+
+    def update_answer(params)
+      file_param = params[:file]
+      file = file_param.tempfile
+
+      key = file_upload_s3_key(file)
+      upload_to_s3(file, key)
+
+      # we don't want to store the file itself on the session
+      self.file = nil
+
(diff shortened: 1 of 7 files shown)


DavidBiddle (dismissed):
Code looks good!

Got it working locally with the spike bucket and have seen my file get uploaded successfully. The real bucket doesn't allow me to upload which I assume is by design 😄

It's probably worth adding a feature test for this work, but I don't think that's necessary for the current iteration.

stephencdaly:
> Code looks good!
> 
> Got it working locally with the spike bucket and have seen my file get uploaded successfully. The real bucket doesn't allow me to upload which I assume is by design 😄
> 
> It's probably worth adding a feature test for this work, but I don't think that's necessary for the current iteration.

Good idea, I've added a feature test

DavidBiddle (approved):
Thanks for adding that feature test!

PR #1090, opened 2024-12-09
rename aws s3 submission variable

alice-carr:

this commit renames the aws s3 submission iam role variable. This is so that we a have a top level of aws in the settings.yml file for when we add more aws variables in the future

[Trello card](https://trello.com/c/snn0svAW/2052-rename-awss3submissions-runner-setting)


cadmiumcat (approved):
👍

PR #1103, opened 2024-12-17
Update to rails 8.0.1

DavidBiddle:

Trello card: 

Updated by:

- taking the initial Dependabot update commit
- running `bin/rails app:update` and reviewing the changes against RailsDiff and the commit history to avoid overwriting our deliberate changes

See individual commit messages for details.


config/environments/test.rb
```diff
@@ -32,33 +30,18 @@
...
   config.active_support.deprecation = :stderr
 
-  # Raise exceptions for disallowed deprecations.
-  config.active_support.disallowed_deprecation = :raise
-
-  # Tell Active Support which deprecation messages to disallow.
-  config.active_support.disallowed_deprecation_warnings = [
-    "/forms/:id endpoint",
-    "/forms/:id/pages endpoint",
-  ]
```
stephencdaly, line 60: Is this no longer needed/supported?

DavidBiddle: It doesn't look like the API for `ActiveSupport::Deprecation` has changed and the [configuration guide](https://guides.rubyonrails.org/configuring.html#config-active-support-disallowed-deprecation) still lists these config settings. I'm not sure why the script removced this but I'll readd it because I think it should still be supported.

DavidBiddle: Done!

chao-xian: @SamJamCul @thomasiles does this relate to the stuff Laurence added for API migration

stephencdaly (approved):
Changes looks sensible and I can't see any issues when running locally

PR #1106, opened 2024-12-18
Attach file uploads to emails

stephencdaly:

Trello card: https://trello.com/c/JEWV7ew5/

Attach files for file upload questions to the email when sending the email using SES. This gets the file from S3 using the key stored against the answer in the session.

Files are stored in memory before the email is sent, but this is unavoidable as we need to read the raw content of the file to pass to ActionMailer to construct the raw email.

This assumes that all files are uploaded with unique filenames - which may not be the case in reality. We will need to ensure that files have unique names but this is not addressed in this PR.

### Testing

It is currently only possible to test this by pushing the branch to the dev environment. I have tested this and got an email with the file attached:

Screenshot 2024-12-18 at 12 11 18


DavidBiddle (approved):
Changes look good, and I've tested them successfully in dev with and without CSVs enabled 🎉

PR #1155, opened 2025-01-23
Stable csv columns

thomasiles:

Trello card: https://trello.com/c/GxKKN67X/2038-implement-csv-formatting-for-routing?filter=member:tomiles3

This PR changes the CSV output for notify and S3 to include headers for all questions in a form. For questions which the filler has not answered (because they have taken other routes through a form with routing) the value will be blank.

The teams using CSV output have been told and have agreed that we can go ahead and release this change.

See the trello ticket for more details.


lfdebrux (approved):
Nice!

PR #1159, opened 2025-01-24
Bump base image

chao-xian:
https://trello.com/c/Q1UuCEPI/2664-ruby-version-update-to-341

Bumps core dependencies and base image in Dockerfile.
- Bumps Ruby from 3.3.4 to 3.4.1
- Bumps Node.js from 20.12.1 to 22.11.0
- Bumps Alpine Linux from 3.19 to 3.21

Trello card: 


spec/controllers/application_controller_spec.rb
```diff
@@ -50,7 +50,7 @@ def raise_error_with_cause
 
       expect(log_lines[0].keys).to include("rescued_exception_trace")
       expect(log_lines[0]["rescued_exception_trace"]).to include(
-        "spec/controllers/application_controller_spec.rb:11:in `raise_error'",
+        "spec/controllers/application_controller_spec.rb:11:in 'raise_error'",
```
lfdebrux, line 53: Not sure if it's worth changing the commit message, but I thought I'd just flag that this change is because of a change in Ruby 3.4's error message formatting: https://bugs.ruby-lang.org/issues/16495

chao-xian: ooh thank you!

chao-xian: Done

DavidBiddle (dismissed):

DavidBiddle (approved):

PR #1160, opened 2025-01-24
Play back what file has been uploaded to users, and allow them to change file

stephencdaly:

Trello card: https://trello.com/c/mfhj3ufe/2068-play-back-what-file-has-been-uploaded-to-users-and-allow-them-to-change-file

This PR adds a review file page that allows a form filler to view the file they have uploaded, and optionally remove it. This is implemented usinga new `ReviewFileController` with new routes for showing the page and removing the file. This also requires tweaking the redirects in the `PageController` to take the user to the review file page:
- after uploading a file
- when editing an answer from check your answers
- when navigating to the file upload question directly with a file already uploaded

This PR will be followed by one which alters the flow slightly, to add a confirmation of removal screen. This will allow us to use the summary list component as intended by the design system (we think!).
### Screenshots

Screenshot 2025-01-28 at 16 47 23
Screenshot 2025-01-29 at 15 06 30
Screenshot 2025-01-28 at 16 48 41

### DIFF
--- a/app/components/question/file_review_component/view.html.erb
+++ b/app/components/question/file_review_component/view.html.erb
@@ -13,7 +13,8 @@
       row.with_cell(text: question.original_filename)
       row.with_cell do |cell|
         form_with url: @remove_file_url, method: :post do |form|
-          form.govuk_submit t('forms.review_file.show.remove'), secondary: true
+          button_content = "#{t('forms.review_file.show.remove')} <span class='govuk-visually-hidden'>#{t('forms.review_file.show.hidden_text')}</span>".html_safe
+          form.govuk_submit button_content, secondary: true
         end
       end
     end
--- a/app/components/question/file_review_component/view.html.erb
+++ b/app/components/question/file_review_component/view.html.erb
@@ -1,0 +1,25 @@
+<% if question.page_heading.present? %>
+  <h2 class="govuk-heading-m"><%= question_text_with_extra_suffix %></h2>
+<% else %>
+  <h1 class="govuk-heading-l"><%= question_text_with_extra_suffix %></h1>
+<% end %>
+
+<%= hint_text %>
+
+<%= govuk_table(classes: "govuk-!-margin-top-7") do |table|
+  table.with_caption(size: 'm', text: t('forms.review_file.show.table_heading'))
+  table.with_body do |body|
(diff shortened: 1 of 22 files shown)


stephencdaly (commented):
I worked on a lot of this but I've checked over everything including the new changes and it all looks good and works for me. I can't approve as I raised the PR.

app/components/question/file_review_component/view.html.erb
```diff
@@ -0,0 +1,23 @@
...
+<%= govuk_table(classes: "govuk-!-margin-top-7") do |table|
+  table.with_caption(size: 'm', text: t('forms.review_file.show.table_caption'))
+  table.with_body do |body|
+    body.with_row do |row|
+      row.with_cell(text: t('forms.review_file.show.table_header'), header: true)
+      row.with_cell(text: question.original_filename)
+      row.with_cell do |cell|
+        form_with url: @remove_file_url, method: :post do |form|
+          button_content = "#{t('forms.review_file.show.remove')} <span class='govuk-visually-hidden'>#{t('forms.review_file.show.table_header')}</span>".html_safe
+          form.govuk_submit button_content, secondary: true
```
DavidBiddle, line 14: I think you can simplify this by supplying a `visually_hidden_suffix` (https://github.com/x-govuk/govuk-components/blob/0be048cd603fcf434e8fcd4144b179561fedeadc/app/helpers/govuk_link_helper.rb#L28)

```suggestion
          button_content = "#{t('forms.review_file.show.remove')} <span class='govuk-visually-hidden'>#{t('forms.review_file.show.table_header')}</span>".html_safe
          form.govuk_submit t('forms.review_file.show.remove'), secondary: true, visually_hidden_suffix: t('forms.review_file.show.table_header')
```

(not blocking, since I know this button is going to disappear soon)

DavidBiddle (approved):
Code looks good, commits are well-explained, and I see the expected behaviour on my machine 🎉

app/components/question/file_review_component/view.html.erb
stephencdaly: Revolving this conversation so this PR can be merged. As this is going to change to a summary list, the component should add the hidden text to the action link for us.

PR #1167, opened 2025-01-28
WIP: Fix the issue which prevented the CSV including all questions in

thomasiles:
### Fix the issue which prevented the CSV including all questions in

Trello card: 

This is a draft PR which fixed the [CSV issue which caused an incident for a particular form when released](https://github.com/alphagov/forms-runner/pull/1155).

When deployed, an exception was thrown when creating the CSV. This happened because the form contained a checkbox question which was skipped due to routing.

The CSV generator asked the checkbox question to produce an answer. A blank string should have been produced but it failed when trying to access a method which didn't exist on nil.

This PR adds unit tests and changes the selection question type to stop it happening.

It also adds a very basic feature tests which uses a form which triggers the issue.

I don't think the feature tests should be written with a JSON literal like this, it was a WIP.

This PR isn't finished and needs more work:
- to verify the issue is fixed for more form types
- sort the test out. 

### DIFF
--- a/app/models/question/selection.rb
+++ b/app/models/question/selection.rb
@@ -48,6 +48,8 @@ module Question
     end
 
     def selection_without_blanks
+      return [] if selection.nil?
+
       selection.reject(&:blank?)
     end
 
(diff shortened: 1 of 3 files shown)


thomasiles:
this has been replaced by #1208

PR #1168, opened 2025-01-28
Enable starting the Rails server using the AWS IAM role for forms-runner

AP-Hunt:

Trello card: https://trello.com/c/l4CIGYi2/596-help-ft-team-1-test-upload-email-locally

In Amazon ECS, forms-runner has credentials for an IAM role that grants it permission to do things like upload files to S3 and send emails via AWS SES. Doing these things locally requires the same set of permissions.

To attain those permissions, developers can now assume the same role as is used by Amazon ECS in the development account by starting the Rails server with the ASSUME_DEV_IAM_ROLE environment variable.

I've not been able to test this much beyond starting the server successfully, so I'd appreciate if somebody with a working `forms-runner` setup could give it a quick prod.

I'm reasonably sure it's all correct from seeing the server start up correctly. Setting environment variables is all that's required to make everything else work.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -59,6 +59,7 @@ gem "aws-sdk-cloudwatch"
 gem "aws-sdk-codepipeline", "~> 1.93"
 gem "aws-sdk-s3"
 gem "aws-sdk-sesv2"
+gem "aws-sdk-sts"
 
 # For sending submissions as CSV
 gem "csv"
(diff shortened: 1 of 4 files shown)


stephencdaly (commented):
I had to make a few changes to get it working for me, were you able to get it working without the modifications I've suggested.

After these changes, it all works and is clearly explained.

AP-Hunt:
> I had to make a few changes to get it working for me, were you able to get it working without the modifications I've suggested?

I was able to assume the readonly role and then start the server successfully yeah -- but I think the server start probably wasn't as successful as it looked if you were seeing things that were just wrong. The perils of testing by "doesn't look wrong"!

stephencdaly (approved):
Nice work! Thanks for doing this

PR #1173, opened 2025-01-30
Update content around file type validation

DavidBiddle:

Trello card: https://trello.com/c/Nr7OQkk2/2092-limit-the-file-types-that-can-be-uploaded

Makes some tweaks to the content we show to form fillers on file upload questions:

- alphabetises the list of file types in the validation message
- adds some new guidance around what file types the user can supply
- puts the new guidance and the existing guidance on file size together in one details component

#### Before
The upload page with a validation error. The error reads 'The selected file must be a CSV, JPEG, JPG, PNG, XLSX, DOC, DOCX, PDF, JSON, ODT, RTF or TXT'. Below the input there is guidance that says 'Your file must be smaller than 7MB.'

### DIFF
--- a/app/components/question/file_component/view.html.erb
+++ b/app/components/question/file_component/view.html.erb
@@ -9,5 +9,3 @@
                                   },
                                   accept: Question::File::FILE_TYPES.join(", ")
 %>
-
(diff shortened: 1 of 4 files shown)


stephencdaly (approved):
Code and test looks good to me, and the content appears as in the design and the trello ticket.

PR #1178, opened 2025-02-04
Update SES email template with GOV.UK branding and content

DavidBiddle:

Trello card: https://trello.com/c/MD1dVOix/2070-update-ses-email-template-to-match-designs-for-file-upload

Updates the SES email template to match the latest designs.

This PR:
- configures ActionMailer preview to make developing templates easier
- uses Notify's template HTML in the mailer layout so that we get the same header and styles as our Notify emails
- adds some content from the design:
  -  '(attached to this email)' suffix to file questions
  -  'Check that this data looks safe' text
- also includes a couple of some small cleanup changes to the markup

See individual commit messages for more details.

### Screenshot
image

### DIFF
--- a/app/views/aws_ses_form_submission_mailer/submission_email.html.erb
+++ b/app/views/aws_ses_form_submission_mailer/submission_email.html.erb
@@ -1,11 +1,11 @@
 <p>
-  This is a completed "<%= @mailer_options.title %>" form.
+  <%= I18n.t("mailer.submission.title", title: @mailer_options.title) %>
 </p>
 <p>
-  This form was submitted at <%= @mailer_options.timestamp.strftime("%l:%M%P").strip %> on <%= @mailer_options.timestamp.strftime("%-d %B %Y") %>
+  <%= I18n.t("mailer.submission.time", time: @mailer_options.timestamp.strftime("%l:%M%P").strip, date: @mailer_options.timestamp.strftime("%-d %B %Y") ) %>
 </p>
 <p>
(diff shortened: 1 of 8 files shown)


DavidBiddle:
> [53 New issues](https://sonarcloud.io/project/issues?id=alphagov_forms-runner&pullRequest=1178&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)

This number is correct but the issues are unavoidable.  For compatibility reasons email markup makes heavy use of layout tables and inline styles, neither of which we would use in website HTML. It's not very maintainable (but we don't ewxpect to have to change this often). We should look at simplifying the template in the future, but for now we're copying what we already get from Notify.

app/views/aws_ses_form_submission_mailer/submission_email.html.erb
DavidBiddle, line 1: We should make the quotes on 
```
This is a completed "<%= @mailer_options.title %>" form.
```
curly

DavidBiddle: We may as well do it as part of this work so I'll pick it up on Monday unless anyone needs to merge this while I'm off

DavidBiddle: Done

stephencdaly (approved):
All the changes look good to me, and I've tested locally and the email looks as expected.

PR #1179, opened 2025-02-04
Add remove file confirmation

jamie-o-wilkinson:

This PR adds a confirmation page for removing an uploaded file. This allows us to use a summary list in the review file page, and better fits with patterns used elsewhere in GOV.UK Forms (e.g. add another answer). This necessitates updating the delete action in `ReviewFileController` to accept an input object and conditionally delete the file.

There are a couple of refactoring bits I'd like to do, but probably in a subsequent PR unless reviewers think it should be done in this one:
- tidying up translation names and reducing duplication
- reducing duplication in `ReviewFileController`
- refactoring remove file and remove answer input objects into a single input object used by both
- suggestions by @stephencdaly:
  - fix back link on confirmation page
  - improve confirmation action name
  - move view to remove_file directory

<details><summary>Screenshots</summary>

Screenshot 2025-02-05 at 11 00 52
Screenshot 2025-02-05 at 11 12 40
Screenshot 2025-02-05 at 11 12 49
Screenshot 2025-02-05 at 11 12 56
Screenshot 2025-02-05 at 11 13 07

</details> 

### DIFF
--- a/app/controllers/forms/review_file_controller.rb
+++ b/app/controllers/forms/review_file_controller.rb
@@ -9,10 +9,21 @@ module Forms
     end
 
     def delete
-      @step.question.delete_from_s3
-      current_context.clear_stored_answer(@step)
+      @remove_file_input = RemoveFileInput.new(remove_file_input_params)
 
-      redirect_to redirect_after_delete_path, success: t("banner.success.file_removed")
+      if @remove_file_input.invalid?
+        back_link(@step.page_slug)
+        @remove_file_url = remove_file_path(form_id: @step.form_id, form_slug: @step.form_slug, page_slug: @step.page_slug, changing_existing_answer:)
+        return render :confirmation, status: :unprocessable_entity
+      end
+
+      if @remove_file_input.remove_file?
+        @step.question.delete_from_s3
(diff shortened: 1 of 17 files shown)


app/controllers/forms/review_file_controller.rb
```diff
@@ -4,23 +4,44 @@ class ReviewFileController < PageController
...
+
+      redirect_to review_file_path(form_id: @step.form_id, form_slug: @step.form_slug, page_slug: @step.page_slug, changing_existing_answer:)
     end
 
     def continue
       redirect_to next_page
     end
 
+    def confirmation
+      back_link(@step.page_slug)
```
stephencdaly, line 34: I think this back link should go back to the review file page, currently it's going to the previous question.

stephencdaly (commented):
The changes look good to me and work locally - with a small issue where the back link goes to the wrong page, which I've left a comment about.

I've also added a suggestion for another possible refactor for the controller but this can be done in a subsequent PR if you agree with the suggestion.

stephencdaly (approved):
Agreed this is fine to merge to get it in before sprint demo. The back link can be fixed in a subsequent PR as no end users can see this yet.

PR #1182, opened 2025-02-05
Add development.yml file with settings for testing file upload

stephencdaly:

Trello card:  https://trello.com/c/xs3nkHZ3/2117-document-testing-file-upload-from-local-to-s3-and-ses

To test file upload locally we need to connect to the AWS dev account by opening an aws shell. This development.yml file adds all the settings required to upload files to S3 and send emails using SES using the aws dev account.


DavidBiddle (approved):
Looks good, and matches what I've got working in my `development.local.yml` file

PR #1183, opened 2025-02-05
Fix back link on file removal confirmation page

jamie-o-wilkinson:

The back link on the file removal confirmation page previously linked to the upload file page, when it should link to the review file page. This PR fixes this. I've also added a `setup_confirmation` method to reduce code duplication.


stephencdaly (approved):
Looks good and works for me, nice refactor too!

PR #1184, opened 2025-02-05
Replace RemoveAnswerInput and RemoveFileInput with RemoveInput

jamie-o-wilkinson:

The `RemoveAnswerInput` and `RemoveFileInput` input objects were identical, so I've replaced them with a common `RemoveInput` input object.


stephencdaly (approved):
Nice refactor! I double checked that everything still looks right for both usages and it does.

PR #1190, opened 2025-02-06
Refactor file removal into RemoveFileController

jamie-o-wilkinson:

The responsibilities of the ReviewFileController were getting muddled with reviewing and deleting an uploaded file. Moving removal to a separate controller is clearer and is more idiomatic.


stephencdaly (approved):
Really nice refactor!

When testing this I spotted that the error summary if you don't select a radio button option on the remove confirmation page isn't at the top of this page, but that wasn't introduced by this PR so can be fixed separately

PR #1204, opened 2025-02-11
Log an error when file upload key is not found in S3

stephencdaly:

Trello card: https://trello.com/c/ZzVCrkDn/2115-test-policy-on-the-s3-buckets-that-clear-up-files-after-30-days

Log at error level with the logging context field `s3_object_key` when a `NoSuchKey` error is raised by the request to get an object from S3. This will allow us to find out the key for the file that was not found to help us diagnose the cause of this error.

Re-raise the error, which will cause the user to see a generic technical difficulties page, and the error will be sent to Sentry.

We only expect this error to be raised when a user has managed to keep the same form filling session open for 30 days and our lifecycle rules for the S3 bucket have deleted the file.


jamie-o-wilkinson (approved):
All looks good to me 👍

PR #1205, opened 2025-02-12
Add initial submissions table

jamie-o-wilkinson:

Trello card: https://trello.com/c/lW3IqOhY

We need to store submissions in order to be able to replay failed submission deliveries. This PR adds an initial submissions table with timestamps only, so that we can get our pipelines working. The next step is to get the AWS pipelines working as they should.


chao-xian (approved):
I did not deserve a co-author credit on the first few commits but ty

PR #1208, opened 2025-02-13
Fix CSV generation to include all questions including skipped checkbox questions

thomasiles:
Trello card: https://trello.com/c/GxKKN67X/2038-implement-csv-formatting-for-routing

This PR makes submission CSVs include headers and values for all questions in a form. It fixes an issue which caused production issues when previously enabling this feature.

See:
[The original PR which caused submissions to fail for a production form](https://github.com/alphagov/forms-runner/pull/1155)
[The PR used during the incident to revert the behaviour](https://github.com/alphagov/forms-runner/pull/1158)
[The incident document](https://docs.google.com/document/d/1M3usd9hEwpnVSnh2WP77Qf6uGak7pASfQqe5AQCIwc4/edit?tab=t.0#heading=h.p99426yo0rbv)

To prevent similar issues happening when deploying this PR the feature test for CSV generation has been adapted to cover the case.

I cloned the production form locally and tested it locally to reproduce the issue and confirm that the fix works. I then made a minimal tests case which is embedded in the feature test to show it's been fixed. The commits should be read in order and hopefully make it clear.

### DIFF
--- a/spec/features/fill_in_and_submit_form_with_csv_spec.rb
+++ b/spec/features/fill_in_and_submit_form_with_csv_spec.rb
@@ -3,30 +3,16 @@ require "rails_helper"
 feature "Fill in and submit a form with a CSV submission", type: :feature do
   let(:steps) do
     [
-      build(:v2_question_page_step, :with_text_settings, id: 1, question_text:, next_step_id: 2),
-      build(:v2_question_page_step, :with_text_settings, id: 2, is_optional: true, question_text: "An optional question?", next_step_id: 3),
-      build(:v2_question_page_step, :with_selections_settings, id: 3, question_text: "A routing question?", routing_conditions: [DataStruct.new(routing_page_id: 3, check_page_id: 3, answer_value: "Option 1", goto_page_id: nil, skip_to_end: true, validation_errors: [])]),
-      build(:v2_question_page_step, :with_text_settings, id: 4, question_text: "a question skipped through routing"),
-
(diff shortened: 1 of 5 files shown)


spec/features/fill_in_and_submit_form_with_csv_spec.rb
```diff
@@ -109,12 +76,14 @@ def and_i_submit_my_form
...
   end
 
-  def then_an_email_submission_should_have_been_sent
+  def and_an_email_submission_should_have_been_sent
+    expect(ActionMailer::Base.deliveries.first).to be_present
+
     delivered_email = ActionMailer::Base.deliveries.first
 
     expected_content = [
-      ["Reference", "Submitted at", question_text, "An optional question?", "A routing question?"],
```
stephencdaly, line 116: Did we remove the optional question because this case is tested sufficiently by including a skipped question and updating the test to check that the header is there for this in the next commit?

stephencdaly (approved):
Looks good and works as expected when I tested locally

spec/features/fill_in_and_submit_form_with_csv_spec.rb
thomasiles: tbh, I removed it because I was focused on the skipped selection case.

I didn't add it back in because I'm not sure it does add anything. This is related to my question in tech standup about adding more examples for form submission. There are a lot of different combinations/settings for forms which  might cause an issue for features.

I don't think this can cover them all and I didn't think optional questions were adding much.

I could add it back in though as it's not doing any harm?

stephencdaly: You're right that it doesn't add anything - we've got a unit test that covers optional questions in the CsvGenerator spec so no need to have a specific assertion on it in the feature test.

lfdebrux (commented):
This looks good to me. Thank so much for improving the tests and writing the comprehensive commit messages.

PR #1221, opened 2025-02-20
Store submission data to handle bounced submission emails

jamie-o-wilkinson:

Trello card: https://trello.com/c/lW3IqOhY

This PR adds a new `Submission` model, and uses it to store submission data when submission emails are sent via AWS SES. In order to enable this, we also added a method to the `SessionAnswerStore` to retrieve answers from the session.

In future, we may make all submission types asynchronous so may persist submission data for all submission types then, as opposed to just for the `AwsSesSubmissionService`.


chao-xian (approved):
I can't see any issues and trust it's all good anyway as 3 senior developers who all worked on this!

PR #1237, opened 2025-03-05
Delete submissions after 7 days

stephencdaly:

Trello card: https://trello.com/c/koB73lTl/

Add a scheduled task which is run by Solid Queue to delete submission which:
* have an `updated_at` date of earlier that 7 days ago
* have a `mail_message_id` present, indicating we have successfully initialised sending the submission email

We are using the `updated_at` date so that when a submission bounces, we will delete it 7 days after the submission is re-attempted. We'll look at ensuring we don't delete bounced submissions until they've been successfully retried in a different PR.

If a submission has answered file upload questions, we delete the uploaded files from S3.

This job will be run on an hourly schedule.

If deleting a submission fails for any reason, we send the error to Sentry and continue deleting other submissions.

### Testing

To test running the recurring job locally, uncomment the lines in `development.rb` to set Solid Queue as the adapter for Active Job.

Start Solid Queue by running `ASSUME_DEV_IAM_ROLE=true ./bin/jobs` while in an AWS shell, so it is possible to delete files from the S3 bucket.

If you see errors being thrown when Solid Queue tries to run the job, look at https://github.com/rails/rails/issues/38560
I found this made it work for me:
`PGGSSENCMODE="disable" ASSUME_DEV_IAM_ROLE=true ./bin/jobs`

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -37,6 +37,8 @@ file_upload:
   poll_scan_status_wait_milliseconds: 500
   poll_scan_status_max_attempts: 20
 
+retain_submissions_for_seconds: 604800 # 7 days
+
 maintenance_mode:
   # When set to true, All pages will render 'Maintenance mode' message
   enabled: false
--- a/app/jobs/delete_submissions_job.rb
+++ b/app/jobs/delete_submissions_job.rb
@@ -1,0 +1,34 @@
+class DeleteSubmissionsJob < ApplicationJob
+  queue_as :default
+
+  def perform(*_args)
+    delete_submissions_updated_before_time = Settings.retain_submissions_for_seconds.seconds.ago
+    submissions_to_delete = Submission.where(updated_at: ..delete_submissions_updated_before_time)
+                            .where.not(mail_message_id: nil)
+
+    submissions_to_delete.find_each { |submission| delete_submission_data(submission) }
+  end
+
+  def delete_submission_data(submission)
+    files = submission.journey.completed_file_upload_questions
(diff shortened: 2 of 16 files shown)


config/settings.yml
```diff
@@ -37,6 +37,8 @@ file_upload:
   poll_scan_status_wait_milliseconds: 500
   poll_scan_status_max_attempts: 20
 
+retain_submissions_for_seconds: 604800
```
DavidBiddle, line 40: Optionally it might be worth adding a comment here to make it obvious what this is in days

DavidBiddle (dismissed):
Raised one optional thing but looks good overall 🎉 

Tested locally and saw it working - I did have to set the `PGGSSENCMODE="disable"` environment variable to run the job, is it worth documenting that in the code somewhere?

stephencdaly:
> Raised one optional thing but looks good overall 🎉
> 
> Tested locally and saw it working - I did have to set the `PGGSSENCMODE="disable"` environment variable to run the job, is it worth documenting that in the code somewhere?

Yeh if that's going to be a problem for everyone we should document it. Possibly adding something to the README for running Solid Queue locally is worthwhile? I might do that as a small separate PR if that's ok

jamie-o-wilkinson (approved):
All looks good, nice one! I'm just gonna test it locally but I see @DavidBiddle has already done that so won't block merging 👍

config/settings.yml
stephencdaly: Good idea, have added

DavidBiddle (approved):

PR #1248, opened 2025-03-10
Store form document on submission

stephencdaly:

Trello card: https://trello.com/c/MK2CoKeB/2753-ensure-we-can-send-submission-emails-when-form-is-changed

Store the form document JSON blob on the submission in the database and use this when reading the form to construct the Journey for generating the submission email in the SendSubmissionJob, and for deleting uploaded files from S3 in the DeleteSubmissionsJob.

We need to do this rather than getting the form from forms-api as the form may have been updated after the submission was made, meaning that the answers may no longer match up with the questions in the most recent version of the form.

To account for submissions that currently exist without the form_document stored, read the form from forms-api if the form_document is not present. This fallback can be removed once we have automatically deleted all submissions that don't have the form_document stored.

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -4,6 +4,7 @@ class Form < ActiveResource::Base
   self.include_format_in_path = false
 
   has_many :pages
+  attr_accessor :document_json
 
   def last_page
     pages.find { |p| !p.has_next_page? }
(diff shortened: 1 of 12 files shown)


DavidBiddle (approved):
Code looks good and I can still see the file details in the email when running locally 🎉

stephencdaly:
@DavidBiddle thanks for the review! I just spotted some tests could be updated so I've pushed 2 extra commits

DavidBiddle (approved):

PR #1249, opened 2025-03-10
Improve logging for jobs

stephencdaly:

Use CurrentAttributes to set attributes that will be logged for all log lines when a submission is being processed. This means that we don't need to manually add logging attributes and anything logged by nested calls gets the same attributes.

Have added a separate `CurrentJobLoggingAttributes` class to distinguish from attributes available for HTTP requests. Also renamed `CurrentLoggingAttributes` -> `CurrentRequestLoggingAttributes` to make this distinction even clearer.

Log the message_id for the SES message as `mail_message_id` to be consistent with what it's called in the database and code, instead of `submission_email_id` which is how we log it for submissions using Notify.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,7 +2,7 @@
 
 class ApplicationController < ActionController::Base
   before_action :set_request_id
-  before_action :set_logging_attributes
+  before_action :set_request_logging_attributes
   before_action :check_maintenance_mode_is_enabled
   after_action :add_robots_header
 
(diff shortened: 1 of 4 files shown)


app/controllers/application_controller.rb
```diff
@@ -27,13 +27,13 @@ def check_maintenance_mode_is_enabled
   end
 
   def set_logging_attributes
```
DavidBiddle, line 29: For additional consistency you could rename all of these to `set_request_logging_attributes`, but I'm not sure how much extra clarity that adds in practice so I'll leave it up to you

thomasiles (dismissed):
This looks good and makes sense 🏅

app/controllers/application_controller.rb
stephencdaly: Not sure it's strictly necessary because as it's in a controller we know we're in the context of a request, but happy to update anyway

DavidBiddle: Good point - probably not worth touching all the files it's in then!

stephencdaly: Too late, already done it

DavidBiddle (dismissed):

stephencdaly:
Sorry, had to resolve some merge conflicts. Could I get another review please?

DavidBiddle (approved):

PR #1252, opened 2025-03-11
Add number suffix to identical file names

DavidBiddle:

When more than one file with the same name has been uploaded in the same form journey, we want to disambiguate them by adding a suffix to some of them before attaching them to the SES email.

In this implementation the first file with a given name will keep the original filename, and subsequent files with the same name will be suffixed like so:
- `my_file.png`
- `my_file_1.png`
- `my_file_2.png`
- `my_file_3.png`

We currently only allow 4 files but this implementation should support any number if we increrase that in future.

This PR also includes a change to truncate filenames to 255 characters or fewer - this is the maximum filename macOS supports by default, Windows supports 256 by default and S3 allows 1024 so I think it makes sense to choose the smallest of these options. 

### Testing instructions
- Create a form with at least 2 file upload questions in your local forms-admin.
- Run the runner in an AWS shell with the IAM role assumed
- Upload files with the same name for every file question in the form (the easiest way is to use the same file every time)
- Check that the submission email includes multiple attachments, and the names are suffixed as detailed above

### DIFF
--- a/app/lib/flow/journey.rb
+++ b/app/lib/flow/journey.rb
@@ -24,6 +24,8 @@ module Flow
       @form = form
       @step_factory = StepFactory.new(form:)
       @completed_steps = generate_completed_steps
+
+      populate_file_suffixes
     end
 
     def find_or_create(page_slug)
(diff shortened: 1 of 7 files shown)


stephencdaly (dismissed):
Really neat solution. It's well tested, and I've checked it works locally.

I have a small question about the filename limit for Windows. From what I can see the limit is 256 for the path + filename - whereas for mac the limit is just for the filename.

If this is the case, do we need to make sure that the filename length is somewhat less than 256 characters to account for the length of the directory part of the file wherever it's downloaded? I wonder whether that's why Notify picked a limit of 100 characters.

DavidBiddle:
@stephencdaly Good spot. The [Windows API maximum file path limitation documentation](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation) suggests that it's possible for the path to be longer if a registry key has been set for a particular app, but we can't guarantee that so it's 256.

Reducing the limit makes sense to me, and 100 characters is still a very long filename so it seems like a reasonable figure to go with. I'll update that now

stephencdaly (approved):

PR #1253, opened 2025-03-11
Add suffix to filename in CSV

DavidBiddle:

Uses the filename suffixes added in https://github.com/alphagov/forms-runner/pull/1252 in the CSV data as well as the email.

This was omitted from the original PR (thanks @stephencdaly for spotting!)


stephencdaly (approved):
Looks good and I've tested locally and it works as expected

PR #1257, opened 2025-03-12
Support optional file questions in csv

DavidBiddle:

The change in #1253 meant that when a user left a file input blank, the CSV generation would break (because it expects a hash but the code was returning `nil`).

This PR updates it and adds some missing tests that would've caught the issue sooner.


stephencdaly (approved):
I've tested locally and it works as expect, and there's now good test coverage for this.

PR #1261, opened 2025-03-13
Show exit pages

thomasiles:
### Add support for exit pages within the runner

Trello card: https://trello.com/c/pfNYulxC/2151-add-support-for-exit-pages-to-forms-runner

Exit pages are given a url like this: "/preview-live/3/branch-route-form/11/exit". "/exit" is added on to the end of the page_slug.

We can change exit to anything else easily, I'll post something in slack and we make the change in another PR. It's just changing the route definition, so shouldn't be hard.

This PR doesn't add events when an exit page is viewed. I'll create another card which captures that so we can discuss what's best with Anne.

image

Note: I mucked up the markdown in the API seed for the exit pages. The above will only show correctly in the preview when[ the PR for fixing the seed](https://github.com/alphagov/forms-api/pull/713) is merged into the API.

### DIFF
--- a/app/controllers/forms/exit_pages_controller.rb
+++ b/app/controllers/forms/exit_pages_controller.rb
@@ -1,0 +1,10 @@
+module Forms
+  class ExitPagesController < PageController
+    def show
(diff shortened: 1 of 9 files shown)


lfdebrux (approved):
This is great, really clear breakdown of the commits and helpful commit messages. I've tested locally, it works for me.

PR #1262, opened 2025-03-13
Handle submission email bounces and complaints

jamie-o-wilkinson:

Trello card: https://trello.com/c/wMIZWtPX

This adds a new job to process bounce and complaint notifications off the bounces and complaints SQS queue. For bounces, we set a new attribute `mail_status` to indicate there is a bounce. In a subsequent PR we'll use this in the delete job to prevent us deleting submissions that need to be replayed.


app/jobs/receive_submission_bounces_and_complaints_job.rb
```diff
@@ -0,0 +1,91 @@
...
+  REGION = "eu-west-2".freeze
+  QUEUE_NAME = "submission_email_ses_bounces_and_complaints_queue".freeze
+  MAX_NUMBER_OF_MESSAGES = 10
+  POLLING_PERIOD = 20
+
+  def perform
+    CurrentJobLoggingAttributes.job_id = job_id
+
+    sts_client = Aws::STS::Client.new(region: REGION)
+    queue_url = "https://sqs.#{REGION}.amazonaws.com/#{sts_client.get_caller_identity.account}/#{QUEUE_NAME}"
```
stephencdaly, line 16: This seems fine, but I wonder is it better for the URL to be passed in as an environment variable? There's probably not significant advantages of doing that though. The main thing I can think of is if we want to mock it out using localstack in any contexts but we don't currently have any plans to do that, or if we have multiple environments in the same AWS account.

Probably leave it as this and we can change later if necessary, unless infra folk have any strong opinions.

app/jobs/receive_submission_bounces_and_complaints_job.rb
```diff
@@ -0,0 +1,91 @@
...
+      ses_event_type = ses_message["eventType"]
+
+      CurrentJobLoggingAttributes.sqs_message_id = sqs_message_id
+      CurrentJobLoggingAttributes.mail_message_id = ses_message_id
+
+      process_bounce(ses_message_id) if ses_event_type == "Bounce"
+      process_complaint(ses_message_id) if ses_event_type == "Complaint"
+
+      sqs_client.delete_message(queue_url: queue_url, receipt_handle: receipt_handle)
+    rescue StandardError => e
```
stephencdaly, line 58: I might be wrong, but I think this will rescue an exception raised in the function and break out of the `each` loop so we won't process further messages in the batch. Do we want to extract a method `process_message` out to process a single message and rescue any exception for that individual message?

app/jobs/receive_submission_bounces_and_complaints_job.rb
```diff
@@ -0,0 +1,91 @@
...
+  def process_messages(sqs_client, queue_url, messages)
+    messages.each do |message|
+      sqs_message_id = message.message_id
+      receipt_handle = message.receipt_handle
+      sns_message = JSON.parse(message.body)
+      ses_message = JSON.parse(sns_message["Message"])
+      ses_message_id = ses_message["mail"]["messageId"]
+      ses_event_type = ses_message["eventType"]
+
+      CurrentJobLoggingAttributes.sqs_message_id = sqs_message_id
```
stephencdaly, line 51: Could we add the SQS message ID to the logging context earlier up as soon as we have it? This means that if the message is malformed and an error is raised parsing it we'll get the message ID on the log line for the error.

app/jobs/receive_submission_bounces_and_complaints_job.rb
```diff
@@ -0,0 +1,91 @@
...
+      Sentry.capture_exception(e)
+    ensure
+      CurrentJobLoggingAttributes.reset
+    end
+  end
+
+  def process_bounce(ses_message_id)
+    submission = Submission.find_by(mail_message_id: ses_message_id)
+
+    raise "Submission not found for SES message ID: #{ses_message_id}" if submission.nil?
```
stephencdaly, line 69: if we use `find_by!` it will raise `ActiveRecord::RecordNotFound`  so we won't need to do this explicitly. Or do we not get as much detail on the error if we do that?

app/jobs/receive_submission_bounces_and_complaints_job.rb
```diff
@@ -0,0 +1,91 @@
...
+      form_id: submission.form_id,
+      submission_reference: submission.reference,
+      job_id:,
+    })
+  end
+
+  def process_complaint(ses_message_id)
+    submission = Submission.find_by(mail_message_id: ses_message_id)
+
+    set_submission_logging_attributes(submission)
```
stephencdaly, line 87: Could we move finding the submission up as it's duplicated in `process_bounce` and `process_complaint`?

app/jobs/receive_submission_bounces_and_complaints_job.rb
```diff
@@ -0,0 +1,91 @@
...
+  def process_bounce(ses_message_id)
+    submission = Submission.find_by(mail_message_id: ses_message_id)
+
+    raise "Submission not found for SES message ID: #{ses_message_id}" if submission.nil?
+
+    set_submission_logging_attributes(submission)
+
+    submission.update!(mail_status: "bounced")
+
+    Rails.logger.info("Submission email bounced - #{self.class.name}")
```
stephencdaly, line 75: Should this use the `EventLogger` just for consistency. I'm not sure how useful it is as I guess we care most about the events for the form filler when we're searching splunk in this way, but I've used it in `SendSubmissionJob` and `DeleteSubmissionsJob` for the events there so it might be good to be consistent.

spec/jobs/receive_submission_bounces_and_complaints_job_spec.rb
```diff
@@ -0,0 +1,196 @@
...
+        expect(log_lines).to include(hash_including(
+                                       "level" => "INFO",
+                                       "message" => "Submission email bounced - ReceiveSubmissionBouncesAndComplaintsJob",
+                                       "form_id" => form_with_file_upload.id,
+                                       "submission_reference" => reference,
+                                       "job_id" => @job_id,
+                                     ))
+      end
+
+      context "when there is no submission found" do
```
stephencdaly, line 106: It might be nice to add in a test that it will continue processing other messages if there is an error processing one message.

spec/jobs/receive_submission_bounces_and_complaints_job_spec.rb
```diff
@@ -0,0 +1,196 @@
...
+        let(:ses_bounce_message_body) { { "mail" => { "messageId" => "mismatched-message-id" }, "eventType": "Bounce" } }
+
+        it "sends an error to Sentry" do
```
stephencdaly, line 109: Should we have an assertion that `:delete_message` is not called?

stephencdaly (commented):
All looks really good. I've left a number of comments but they're all pretty minor. Happy to chat through any of them.

app/jobs/receive_submission_bounces_and_complaints_job.rb
jamie-o-wilkinson: I think that sounds like a good idea, but like you say, probably best to do when we need it for e.g. localstack.

app/jobs/receive_submission_bounces_and_complaints_job.rb
jamie-o-wilkinson: I think this is due to some really confusing rubocop behaviour. I did have the code inside the loop wrapping explicitly in a begin rescue ensure block, and rubocop insists on correcting it to this. I will test that the loop continues, I agree it looks like it wouldn't...

stephencdaly: Yeh I've had rubocop do this before and had to extract a method to get the behaviour I wanted

jamie-o-wilkinson: I've added a test to check it continues processing subsequent messages, and it seems to. So the behaviour is correct, but the code looks weird. The key thing is the rescue ensure blocks are still within the loop block.

app/jobs/receive_submission_bounces_and_complaints_job.rb
jamie-o-wilkinson: Ooh yeah, nice one. Will change that now.

app/jobs/receive_submission_bounces_and_complaints_job.rb
jamie-o-wilkinson: I'm going to park this for now so we can get this moving 😄

spec/jobs/receive_submission_bounces_and_complaints_job_spec.rb
```diff
@@ -0,0 +1,224 @@
...
+
+          expect(sqs_client).not_to have_received(:delete_message)
+        end
+      end
+
+      context "when updating submission fails" do
+        let(:messages) { [sqs_bounce_message, sqs_bounce_message] }
+
+        before do
+          # allow(Submission).to receive(:find_by).and_raise(StandardError, "Test error")
```
stephencdaly, line 131: Commented out code

spec/jobs/receive_submission_bounces_and_complaints_job_spec.rb
```diff
@@ -0,0 +1,224 @@
...
+        it "continues processing subsequent messages" do
+          expected_loop_count = messages.length - 1
+
+          perform_enqueued_jobs
+
+          expect(sqs_client).to have_received(:delete_message).exactly(expected_loop_count).times
+        end
+      end
+
+      it "alerts to Sentry that there was a bounced delivery" do
```
stephencdaly, line 177: Would it make sense to move this `it` and the one below to be above the nested contexts?

stephencdaly (commented):
Just some minor stylistic comments otherwise looks great!

db/schema.rb
```diff
@@ -22,7 +22,9 @@
     t.integer "form_id"
     t.jsonb "answers"
     t.string "mode"
+    t.string "mail_status", default: "pending", null: false
```
stephencdaly, line 25: When I run `./bin/setup` locally, it wants to move `mail_status` to be below `form_document` in this file. I guess this is dependant on the order the migrations are run in, does it make sense to move this this line down?

Would it also make sense to regenerate your migration, as there is a migration with a later date that's already been run on production?

jamie-o-wilkinson: I tried this, but because the migrations have run in this order on dev already, it's going to be a faff to get the migrations on dev happy with switching the order...that said perhaps we'll have the same problem on other environments because the `form_document` was added first according to those environments...

stephencdaly (approved):

PR #1263, opened 2025-03-13
Include page heading in Check Your Answers rows for file upload questions

DavidBiddle:

Trello card: https://trello.com/c/wMIZWtPX/2160-include-page-heading-in-cya-page-when-guidance-is-present

Adds a caption with the page heading to the Check Your Answers page when the question has guidance.

At the moment this will only show for File questions, but if we're happy with this we can apply t to all questions with guidance in future.

#### Before
A Check your answers summary list with two 'Upload your file (optional)' keys and one 'What is your date of birth' key

#### After
The same Check your answers summary list, but one of the file upload keys now has a caption above it saying 'Driving Licence' and the other has a caption saying 'passport'

### DIFF
--- a/app/controllers/forms/check_your_answers_controller.rb
+++ b/app/controllers/forms/check_your_answers_controller.rb
@@ -53,11 +53,11 @@ module Forms
   private
 
     def page_to_row(page)
-      question_name = page.question.question_text_with_optional_suffix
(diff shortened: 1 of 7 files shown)


stephencdaly (approved):
Nice changes and simplifications. I've tested locally and it works as expected. I agree with your reasoning for removing the hidden mode string.

PR #1267, opened 2025-03-14
Send metrics for background jobs

stephencdaly:

Trello card: https://trello.com/c/sSfVKcbD/2143-set-up-monitoring-and-observability-for-ses-file-upload-submissions

Send CloudWatch metrics for:
* When a submission is successfully sent, send a metric for the time since it was scheduled to be sent. We will add an alert on this metric if the time is longer than we think is reasonable.
* When sending a submission fails. We can add an alert if there is a spike in failures.
* When the job to delete submissions starts. We can add an alert if the job isn't run within the expected timeframe. We will add a similar metric for the job to process bounces from the SQS queue once that has been added.


app/services/cloud_watch_service.rb
jamie-o-wilkinson, line 1: I really like that for the job generic events we just used a dimension for the job name 👍

jamie-o-wilkinson (approved):
Code looks good to me and it has been tested against dev. Nice one!

PR #1273, opened 2025-03-17
Handle success notifications

jamie-o-wilkinson:

Trello card: https://trello.com/c/Qf0gkFUE

This PR adds an Active Job to process delivery notifications from SES. We update the `mail_status` on the `Submission` and logs the successful delivery. There is quite a bit of duplicated code between `ReceiveSubmissionDeliveriesJob` and `ReceiveSubmissionBouncesAndComplaintsJob` which could in future be extracted into an SQS related class at some point in the future, but I decided not to do this at this stage since there are only two uses so far.

### DIFF
--- a/app/jobs/receive_submission_deliveries_job.rb
+++ b/app/jobs/receive_submission_deliveries_job.rb
@@ -1,0 +1,74 @@
+class ReceiveSubmissionDeliveriesJob < ApplicationJob
+  require "aws-sdk-sqs"
+  require "aws-sdk-sts"
+
+  queue_as :background
+
+  REGION = "eu-west-2".freeze
+  SQS_QUEUE_NAME = "submission_email_ses_successful_deliveries_queue".freeze
+  MAX_NUMBER_OF_MESSAGES = 10
+  POLLING_PERIOD = 20
+
+  def perform
+    CurrentJobLoggingAttributes.job_id = job_id
+
+    sts_client = Aws::STS::Client.new(region: REGION)
+    queue_url = "https://sqs.#{REGION}.amazonaws.com/#{sts_client.get_caller_identity.account}/#{SQS_QUEUE_NAME}"
+    sqs_client = Aws::SQS::Client.new(region: REGION)
+
+    loop do
+      messages = receive_messages(sqs_client, queue_url)
+
+      break if messages.empty?
+
+      process_messages(sqs_client, queue_url, messages)
+    end
+  end
+
+private
+
+  def receive_messages(sqs_client, queue_url)
+    response = sqs_client.receive_message(
+      queue_url: queue_url,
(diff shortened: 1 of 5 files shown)


spec/jobs/receive_submission_deliveries_job_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+        it "sends an error to Sentry" do
+          allow(Sentry).to receive(:capture_exception)
+
+          perform_enqueued_jobs
+
+          expect(Sentry).to have_received(:capture_exception)
+        end
+      end
+
+      it "deletes the SQS message" do
```
stephencdaly, line 148: It might be nice to move this `it` above the nested context for readability.

stephencdaly (commented):
Left some comments about potential code sharing but this isn't 100% necessary at this point. But might be nice to do sooner than later as it will make any changes we might need to make in future easier.

There were some additional tests/assertions that were added for bounces/complaints that should probably be added here.

app/jobs/receive_submission_deliveries_job.rb
```diff
@@ -0,0 +1,74 @@
...
+
+      receipt_handle = message.receipt_handle
+      sns_message = JSON.parse(message.body)
+      ses_message = JSON.parse(sns_message["Message"])
+      ses_message_id = ses_message["mail"]["messageId"]
+      CurrentJobLoggingAttributes.mail_message_id = ses_message_id
+
+      ses_event_type = ses_message["eventType"]
+
+      submission = Submission.find_by!(mail_message_id: ses_message_id)
```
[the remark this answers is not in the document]
jamie-o-wilkinson: Similar to above, there is a lot of (for now) intentional duplication.

app/jobs/receive_submission_deliveries_job.rb
```diff
@@ -0,0 +1,74 @@
...
+
+      process_delivery(submission) if ses_event_type == "Delivery"
```
stephencdaly, line 57: If the `eventType` isn't "Delivery" should we be deleting the message or should we instead throw an error? This shouldn't happen under regular circumstances but if we're checking the type at all maybe we shouldn't ignore it if it isn't what we expect.

stephencdaly (approved):
Sorry spotted one small thing about checking the eventType but not a priority so this can be merged as is and we can think more about this afterwards if necessary.

app/jobs/receive_submission_deliveries_job.rb
jamie-o-wilkinson: Good thought, I suppose the same is true of the bounces and complaints job. A check that the `eventType` is what is expected and raising an error if not sounds best?

stephencdaly (approved):

PR #1278, opened 2025-03-18
Add background job to send queue metrics

stephencdaly:

Trello card: https://trello.com/c/sSfVKcbD/2143-set-up-monitoring-and-observability-for-ses-file-upload-submissions

This PR:

* Assigns all jobs to named queues.
* Adds a job to send a metric for the length of the submissions queue to CloudWatch. This job runs as the highest priority to ensure we are sending regular metrics.

I did consider adding a separate worker for running background jobs so these would not block sending submissions if there was a bug that caused them to be long-running. However, we run a minimum of 6 forms-runner ECS tasks on production so it is unlikely we'd run into capacity issues. We have alerts that will tell us if we're struggling to keep up with demand.

## Testing

I've tested this by connecting my local environment to the AWS dev account:

Screenshot 2025-03-18 at 13 12 23

### DIFF
--- a/app/jobs/delete_submissions_job.rb
+++ b/app/jobs/delete_submissions_job.rb
@@ -1,5 +1,5 @@
 class DeleteSubmissionsJob < ApplicationJob
-  queue_as :default
+  queue_as :background
 
   def perform(*_args)
     CloudWatchService.log_job_started(self.class.name)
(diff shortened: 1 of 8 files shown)


jamie-o-wilkinson (approved):
All looks good to me 🥇

PR #1281, opened 2025-03-20
Add rake task to retry bounced submissions

jamie-o-wilkinson:

Add a rake task to retry submissions sent via SES that bounced. This accepts a form id as a parameter, and will enqueue each submission for that form that has a mail status of "bounced".

Trello card: https://trello.com/c/Qf0gkFUE

### DIFF
--- a/lib/tasks/submissions.rake
+++ b/lib/tasks/submissions.rake
@@ -1,0 +1,19 @@
+namespace :submissions do
+  desc "Retry bounced submissions"
+  task :retry_bounced_submissions, %i[form_id] => :environment do |_, args|
+    form_id = args[:form_id]
+
+    usage_message = "usage: rake submissions:retry_bounced_submissions[<form_id>]".freeze
+    abort usage_message if form_id.blank?
+
+    submissions_to_retry = Submission.where(form_id: form_id)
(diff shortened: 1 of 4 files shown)


lib/tasks/submissions.rake
```diff
@@ -0,0 +1,19 @@
...
+    abort usage_message if form_id.blank?
+
+    submissions_to_retry = Submission.where(form_id: form_id)
+                                      .where(mail_status: "bounced")
+
+    Rails.logger.info "#{submissions_to_retry.length} submissions to retry for form with ID: #{form_id}"
+
+    submissions_to_retry.each do |submission|
+      Rails.logger.info "Retrying submission with reference #{submission.reference} for form with ID: #{form_id}"
+      SendSubmissionJob.perform_later(submission)
```
stephencdaly, line 16: I think something should reset the `mail_status` to `pending` so we'll definitely delete the submission down the line. Possibly this should be the `SendSubmissionJob` when it updates the `mail_message_id`

jamie-o-wilkinson: I've added a commit which does this as you suggest. Good spot!

stephencdaly (approved):

PR #1282, opened 2025-03-20
Add improved error reporting for email formatters

lfdebrux:

When there is an error formatting the content for a submission email it is currently hard to figure out which answer the formatting failed on from the logs we have (see for example https://govuk-forms.sentry.io/issues/6426798718/); this makes it hard to be confident about what the cause might have been.

This commit adds the page ID for the answer that formatting failed for to a new FormattingError exception. The exception cause and backtrace should include information about the specific error that occurred when formatting, so we should still have that information as well.


thomasiles (approved):
nice cleanup and looks like it will be useful in production 👍

PR #1285, opened 2025-03-24
Add plain text part to submission email

DavidBiddle:

Trello card: https://trello.com/c/2DMKRjiO/2170-include-plain-text-version-for-emails-sent-using-ses

Updates the SES mailer template with a plaintext template, so that the email will be a multipart email with HTML and plaintext parts.

### Testing instructions
You can use the mailer preview (http://localhost:3001/rails/mailers/aws_ses_form_submission_mailer/submission_email) to see how the email should look in HTML and plaintext forms.

If running the product in a local AWS shell, you can see the real thing by submitting a file upload form. In Gmail you can then use '[Show original](https://support.google.com/mail/answer/29436?hl=en-GB#zippy=)' to check whether you can see the plaintext content in the email code.

The content should look something like this:

```
This is a completed “Form 1” form.

This form was submitted at 3:28pm on 24 March 2025

GOV.​UK Forms reference number: JCHF7487

Check that this data looks safe before you use it

---

## Upload your file

image.jpg (attached to this email)

---

## You cannot reply to this email

If you need to contact the person who completed this form, you’ll need to contact them directly.
If you’re experiencing a technical issue with this form, contact the GOV.​UK Forms team (https://www.forms.service.gov.uk/support) with details of the issue and the form it relates to.
```

### DIFF
--- a/app/lib/ses_email_formatter.rb
+++ b/app/lib/ses_email_formatter.rb
@@ -3,7 +3,7 @@ class SesEmailFormatter
 
   class FormattingError < StandardError; end
 
-  def build_question_answers_section(completed_steps)
+  def build_question_answers_section_html(completed_steps)
     completed_steps.map { |page|
       [prep_question_title(page),
        prep_answer_text(page)].join
--- a/app/lib/ses_email_formatter.rb
+++ b/app/lib/ses_email_formatter.rb
@@ -18,6 +18,8 @@ class SesEmailFormatter
     }.join(H_RULE_PLAIN_TEXT)
   end
 
+private
+
   def prep_question_title_html(page)
     "<h2>#{remove_heading_hashes(prep_question_title_plain_text(page))}</h2>"
   end
--- a/app/lib/ses_email_formatter.rb
+++ b/app/lib/ses_email_formatter.rb
@@ -21,7 +21,7 @@ class SesEmailFormatter
 private
 
   def prep_question_title_html(page)
-    "<h2>#{remove_heading_hashes(prep_question_title_plain_text(page))}</h2>"
+    "<h2>#{prep_question_title_plain_text(page)}</h2>"
   end
 
   def prep_answer_text_html(page)
@@ -31,7 +31,7 @@ private
   end
 
   def prep_question_title_plain_text(page)
-    "## #{page.question_text}"
+    page.question_text
   end
 
   def prep_answer_text_plain_text(page)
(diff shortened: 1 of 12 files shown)


app/lib/ses_email_formatter.rb
```diff
@@ -19,15 +19,11 @@ def build_question_answers_section_plain_text(completed_steps)
   end
 
   def prep_question_title(page)
```
stephencdaly, line 23: Would it be helpful to rename these methods to `prep_question_title_html` and `prep_answer_text_html` too to follow the other renames? I was slightly confused when reviewing this before I checked out the branch to look at it locally.

app/lib/ses_email_formatter.rb
```diff
@@ -1,36 +1,63 @@
...
   end
 
   def prep_answer_text(page)
+    "<p>#{convert_newlines_to_html(prep_answer_text_plain_text(page))}</p>"
+  rescue StandardError
+    raise FormattingError, "could not format answer for question page #{page.id}"
+  end
+
+  def prep_question_title_plain_text(page)
+    "## #{page.question_text}"
```
stephencdaly, line 34: Do we want to have hashes for headings? Will the email client be able to process the markdown?

Looks like the plain version of the Notify emails don't differentiate headings from other content.

stephencdaly (approved):
Minor style comment and a question about using markdown but nothing to stop this being merged.

app/lib/ses_email_formatter.rb
DavidBiddle: Good point, have updated!

DavidBiddle: Good suggestion - have updated!

app/lib/ses_email_formatter.rb
DavidBiddle: I wasn't 100% sure on this - I wasn't sure whether we'd want some plaintext way of differentiating the heading from the other text. I did it markdown-style because it felt like a good default plaintext formatting convention, but I wouldn't expect plaintext mail clients to do anything with it.

I've updated it to make it plaintext since that matches Notify more closely, we can revisit later if we think formatting would be helpful. 

Eventually it might make sense for us to do what Notify do, and generate the email content as markdown and then convert it into plaintext and html versions with different markdown formatters. Probably not necessary unless we start adding more complex formatting to the show_answer_in_email method

stephencdaly (approved):

PR #1287, opened 2025-03-24
Attachment sent status

chao-xian:

Trello card: https://trello.com/c/ZKseq4XJ/2157-add-e2e-test-for-sending-file-as-an-email-attachment

## What

Adds an endpoint that returns if a submission was emailed (via SES) for a given reference.

## Why

As part of e2e testing of File Upload. We have no way of checking that the email with attachment was sent from the e2e testing point of view. 

Having a way to check the status of submissions could also feed into monitoring.

### DIFF
--- a/app/controllers/submission_status_controller.rb
+++ b/app/controllers/submission_status_controller.rb
@@ -1,0 +1,15 @@
+class SubmissionStatusController < ApplicationController
+  def status
+    if Submission.emailed?(submission_params[:reference])
+      head :no_content
+    else
+      head :not_found
+    end
+  end
+
+private
+
+  def submission_params
+    params.permit(:reference)
+  end
+end
--- a/app/controllers/submission_status_controller.rb
+++ b/app/controllers/submission_status_controller.rb
@@ -1,4 +1,5 @@
 class SubmissionStatusController < ApplicationController
+  before_action :authenticate_client
   def status
     if Submission.emailed?(submission_params[:reference])
       head :no_content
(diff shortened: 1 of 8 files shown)


stephencdaly (approved):
Code looks good and it works for me locally.

Is the idea we'll add in a shared secret for authentication we can check against? Do we want to add that in before deploying this?

chao-xian:
@stephencdaly have added restriction as discussed. Related PR https://github.com/alphagov/forms-deploy/pull/1482

app/controllers/submission_status_controller.rb
```diff
@@ -9,6 +10,12 @@ def status
 
 private
 
+  def authenticate_client
+    authenticate_or_request_with_http_token do |token, _options|
+      ActiveSupport::SecurityUtils.secure_compare(token, Settings.submission_status_api[:secret])
```
stephencdaly, line 15: nice

spec/controllers/submission_status_controller_spec.rb
```diff
@@ -2,31 +2,47 @@
...
-        expect(response).to have_http_status :not_found
+          expect(response).to have_http_status :not_found
+        end
       end
     end
 
-    context "when a submission does not exist" do
-      it "returns a 404 status" do
-        get :status, params: { reference: "999" }
+    context "when no authorization token is included in request" do
```
stephencdaly, line 41: Is it worth having one more test for when the authorisation token doesn't match, just for completeness?

config/settings.yml
```diff
@@ -24,6 +24,9 @@ sentry:
   environment: local
   filter_mask: "[Filtered (client-side)]"
 
+submission_status_api:
+  secret: changeme
```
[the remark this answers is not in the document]
chao-xian: Added to `development.yml`

stephencdaly (approved):
Works as expected locally. Nice work 🎉

PR #1292, opened 2025-03-25
Use sent at date for deleting submissions

stephencdaly:

Trello card: https://trello.com/c/Qf0gkFUE/2155-handle-bounced-submission-emails

We were relying on the updated_at to decide when to delete a Submission. However, we're now updating the Submission record when we get notification via SNS that a submission has been successfully delivered. We don't know how long it will take to get and process this notifaction, and we don't want to delay deleting submission past our 7 day retention period.

Instead, store a `sent_at` date and use that to decide when to delete submissions.


jamie-o-wilkinson (approved):
Code looks good and works locally.

PR #1294, opened 2025-03-26
Add mail_status enum to constrain possible values

jamie-o-wilkinson:

We want to prevent `mail_status` from being set to any string, and to document in code the possible values of `mail_status`. This is done using an enum in the `Submission` model.


stephencdaly (approved):
Nice refactor. Left a suggestion that we can use the scopes that ActiveRecord provides for the enum, but not strictly necessary

lfdebrux (dismissed):
Nice! 🙌

stephencdaly (approved):

PR #1298, opened 2025-03-27
Put CSV file in folder for submissions to s3

stephencdaly:

Trello card: https://trello.com/c/6GaKpHgH/2172-decide-how-we-will-handle-including-uploaded-files-in-submissions-to-s3

To get ready for supporting file upload for S3 submissions, put the CSV file in a "folder" in S3 for the submission, which will also contain the associated uploaded files if the form has file upload questions, when we're supporting this.

The structure will look like:

```
form_submissions/<form_id>/
|- <timestamp>_<reference>/
| |- an_uploaded_file.jpg
| |- another_file_maybe.pdf
| |- form_submission.csv
```


DavidBiddle (approved):
Looks good! 🎉

PR #1300, opened 2025-03-31
Include `preview` flag on log lines

stephencdaly:

Include a `preview` flag that indicates whether the form is being filled out in preview mode or not. Most events we don't log in preview mode, but it's useful to know for the logs that we do have so we can exclude preview submissions from certain dashboards and reports.

Include it as a string rather than a boolean because otherwise we strip it out in the logger if it's set to false.


DavidBiddle (approved):
Code looks good and can see the new attribute in my local logs 🎉

PR #1313, opened 2025-04-04
Add failing spec for emails that should be invalid

chao-xian:
Our current validation checks on email addresses don't match with what Notify check. Adding their set of test examples[1] to reproduce the problem.

[1] https://github.com/alphagov/notifications-utils/blob/main/tests/recipient_validation/test_email_address.py#L27C5-L56C45

Trello card: 


chao-xian:
This got included in work by @lfdebrux and merged. closing this PR

PR #1325, opened 2025-04-11
Send uploaded files with S3 submissions

stephencdaly:

Trello card: https://trello.com/c/hqPJsvS9/2194-support-file-upload-for-submissions-to-s3

When the `submission_type` is set to `s3` for a form, and the form has answered file upload questions, send the uploaded files to the s3 bucket set on the form.

Do this by copying the files to the target bucket and then deleting them from our S3 bucket. Do not copy the object tagging when we do the copy as the form processor does not need the tag added by GuardDuty and to copy the tagging we would need to ask form processors to allow the s3:PutObjectTagging permission for our IAM role.

Copy the files before uploading the submission S3 to the target bucket so that form processors are able to add automations on the CSV file being added and be sure that any uploaded files referenced in the CSV are already present.

### DIFF
--- a/app/lib/csv_generator.rb
+++ b/app/lib/csv_generator.rb
@@ -4,12 +4,13 @@ class CsvGenerator
   CSV_EXTENSION = ".csv".freeze
   CSV_FILENAME_PREFIX = "govuk_forms_".freeze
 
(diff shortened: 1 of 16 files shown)


DavidBiddle (approved):
Code changes look good, and have tested successfully locally for both s3 and email submission 🎉

PR #1326, opened 2025-04-11
Add CSV file information to submission emails

jamie-o-wilkinson:

Trello card: https://trello.com/c/C8KYvyW8

We need to add CSV related info to the submission emails sent via AWS SES, so that forms that are receiving answers in CSV files receive submission emails with the right content once we switch over to using AWS SES for all submission emails.

<details><summary>Mailer previews</summary>
HTML:

Screenshot 2025-04-14 at 10 15 21

Plain-text:
```
This is a test of the “Form 1” form.

This form was submitted at 9:15am on 14 April 2025

GOV.​UK Forms reference number: GKCW5J8V

You can check that a payment has been made by using this reference number to search transactions within GOV.​UK Pay.

Check that this data looks safe before you use it

---

## What's your email address?

forms@example.gov.uk

---

CSV file of these answers

my_answers.csv (attached to this email)

---

You cannot reply to this email

If you need to contact the person who completed this form, you’ll need to contact them directly.
If you’re experiencing a technical issue with this form, contact the GOV.​UK Forms team (https://www.forms.service.gov.uk/support) with details of the issue and the form it relates to.
```
</details> 

### DIFF
--- a/app/mailers/aws_ses_form_submission_mailer.rb
+++ b/app/mailers/aws_ses_form_submission_mailer.rb
@@ -3,11 +3,12 @@ class AwsSesFormSubmissionMailer < ApplicationMailer
           reply_to: Settings.ses_submission_email.reply_to_email_address,
           delivery_method: Rails.configuration.x.aws_ses_form_submission_mailer["delivery_method"]
 
-  def submission_email(answer_content_html:, answer_content_plain_text:, submission_email_address:, mailer_options:, files:)
+  def submission_email(answer_content_html:, answer_content_plain_text:, submission_email_address:, mailer_options:, files:, csv_filename: nil)
     @answer_content_html = answer_content_html
     @answer_content_plain_text = answer_content_plain_text
     @mailer_options = mailer_options
     @subject = email_subject
+    @csv_filename = csv_filename
 
     files.each do |name, file|
       attachments[name] = file
--- a/app/services/aws_ses_submission_service.rb
+++ b/app/services/aws_ses_submission_service.rb
@@ -32,16 +32,17 @@ private
       write_submission_csv(file)
 
       files = files.merge({ csv_filename => File.read(file.path) })
-      deliver_submission_email(files)
+      deliver_submission_email(files, csv_filename)
     end
   end
 
-  def deliver_submission_email(files)
+  def deliver_submission_email(files, csv_filename = nil)
     mail = AwsSesFormSubmissionMailer.submission_email(answer_content_html:,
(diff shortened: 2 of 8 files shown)


app/mailers/aws_ses_form_submission_mailer.rb
```diff
@@ -3,11 +3,12 @@ class AwsSesFormSubmissionMailer < ApplicationMailer
           reply_to: Settings.ses_submission_email.reply_to_email_address,
           delivery_method: Rails.configuration.x.aws_ses_form_submission_mailer["delivery_method"]
 
-  def submission_email(answer_content_html:, answer_content_plain_text:, submission_email_address:, mailer_options:, files:)
+  def submission_email(answer_content_html:, answer_content_plain_text:, submission_email_address:, mailer_options:, files:, csv_filename: nil)
```
jamie-o-wilkinson, line 6: We could potentially avoid adding this parameter by extracting the CSV filename from the list of files, if it is there. That wouldn't be robust to a file being uploaded with the same name as the generated CSV, and wasn't explicit, so I preferred this. Open to suggestions though!

app/services/aws_ses_submission_service.rb
```diff
@@ -41,7 +44,8 @@ def deliver_submission_email(files)
                                                        answer_content_plain_text:,
                                                        submission_email_address: @form.submission_email,
                                                        mailer_options: @mailer_options,
-                                                       files:).deliver_now
+                                                       files:,
+                                                       csv_filename: @csv_filename).deliver_now
```
stephencdaly, line 48: It feels slightly neater if we were to pass `csv_filename` in as a parameter to the method. Setting it as an instance variable makes me feel a bit weird as we can't be so sure that its value has been set when we access it as it isn't set by the initialize method. It's very minor but I think having it a parameter would be slightly easier to follow.

app/mailers/aws_ses_form_submission_mailer.rb
stephencdaly: I think this is probably the best way of doing it

stephencdaly (approved):
Left a very minor style comment, but it's good to go if you disagree. Tested locally and it works as expected.

app/services/aws_ses_submission_service.rb
jamie-o-wilkinson: Yeah that's much neater, not sure why I didn't do that in the first place. Thanks for the suggestion!

stephencdaly (approved):

PR #1345, opened 2025-04-23
Show error message if the file is empty

stephencdaly:

Trello card: https://trello.com/c/VB6S8I1d/2197-add-error-messages-for-form-fillers-if-the-file-they-select-is-empty

Validate that files aren't empty, and if they are show a validation error to the user.

Screenshot 2025-04-23 at 16 42 35


DavidBiddle (approved):
Code looks good and the validation works for me locally 🎉

PR #1367, opened 2025-05-02
Switch to GOV.UK Rebrand

DavidBiddle:

Trello card: https://trello.com/c/L6O6609n/2322-switch-over-to-new-govuk-brand

PR introducing [v5.11.0 of govuk-frontend](https://github.com/alphagov/govuk-frontend/releases/tag/v5.11.0) and [v5.11.0 version of govuk-components](https://rubygems.org/gems/govuk-components/versions/5.11.0). Goes through all the steps on the govuk-frontend release page to make sure that all of the rebrand changes will be present.


jamie-o-wilkinson (approved):
Code looks good and works as expected, in review apps and in dev. 

Minor thing: I think the PR description is out of date? It refers to 5.10.x versions instead of 5.11.x, and also says:

> Note: We don't intend to merge this as-is, it's just for testing the rebrand work. It needs to be deployed to an environment

PR #1384, opened 2025-05-09
Use the AWS ActiveJob SQS adapter

theseanything:
We currently are using a scheduled solid queue job to run a custom poller implemetation for SQS to process messages. We do this for two ActiveJobs `ReceiveSubmissionBouncesAndComplaintsJob` and `ReceiveSubmissionDeliveriesJob`.

We can use the native [SQS adapter for ActiveJob](https://github.com/aws/aws-activejob-sqs-ruby) and remove custom poller implementation. It provides better support for configuration, multiple workers, back pressure, message visibility_timeout. It also allows to process messages closer to realtime, rather than every 10 mins.

Before we can adopt the AWS ActiveJob SQS we need to wait for support for processing events not added by ActiveJob. This has been raised in [an issue](https://github.com/aws/aws-activejob-sqs-ruby/issues/21) and already has [PR raised to add support](https://github.com/aws/aws-activejob-sqs-ruby/pull/22).

We also needed to add a Puma plugin to spawn the poller processes for the SQS Adapter. This plugin implemenation could be contributed back to the aws-activejob-sqs library.

In production we should run these worker processes in a separate workload to the main rails server (something we need to fix with our Solid Queue worker processes). However the puma plugin may still be helpful if running locally.
This pull request introduces significant changes to integrate `aws-activejob-sqs` for handling SQS queues, replacing the previous custom implementation. It includes updates to dependencies, job classes, configuration files, and the addition of a custom Puma plugin to manage the SQS poller process. Below is a summary of the most important changes grouped by theme.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -65,8 +65,11 @@ gem "aws-sdk-cloudwatch"
 gem "aws-sdk-codepipeline", "~> 1.97"
 gem "aws-sdk-s3"
 gem "aws-sdk-sesv2"
-gem "aws-sdk-sqs"
-gem "aws-sdk-sts"
+
+# TODO: Install gem from RubyGems when events support is released
+# https://github.com/aws/aws-activejob-sqs-ruby/pull/22
+gem "aws-activejob-sqs"
+gem "aws-sdk-rails"
 
 # For sending submissions as CSV
 gem "csv"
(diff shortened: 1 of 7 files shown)


lfdebrux:
@theseanything is this PR something you're still working on, or can it be closed?

PR #1387, opened 2025-05-12
Review apps: add Terraform in support of forms-runner review apps

AP-Hunt:

Trello card: https://trello.com/c/4r1IFARH/675-add-review-apps-to-forms-runner

This Terraform is cloned from the equivalent code in forms-admin, and adjusted for the needs of forms-runner. The code duplication is a choice we made, because these two sets of Terraform code are only similar, and not the same bar some parameter differences. Trying to share code between the two would cause more problems than it solves.


sarahseewhy (approved):
This looks solid. There's nothing that I can see which needs changing. Code is clear and I appreciate all the comments and documentation. Thank you 🪄

PR #1388, opened 2025-05-12
Display submitted at time in local time

stephencdaly:

Trello card: https://trello.com/c/4EnxrmLq/2864-submitted-at-time-is-in-utc-in-submission-emails-after-switching-to-ses

Emails sent using SES were displaying the submitted at time in UTC rather than in local time. This is because we are using the created_at time for the submission when processing the SendSubmissionJob, which is stored as UTC.

We need to convert the UTC time to the time we've configured to display submissions in, which is London time.


DavidBiddle (approved):
Code looks good and works for me locally 🦖 🦕 🎉

PR #1394, opened 2025-05-14
Log when file upload details component is clicked

DavidBiddle:

Trello card: https://trello.com/c/0mqWd67h/2177-log-when-what-files-can-i-upload-details-component-is-clicked

Adds a new tracker for Gogle Analytics that will push an event to the dataLayer if a details component is clicked:

```js
{
  event: 'details_opened',
  url: "https://submit.forms.service.gov.uk/form/5616/apply-for-a-juggling-licence-advanced/24586",
  text: "What files can I upload?"
}
```

This will help us see how often these bits of guidance are being used.

### DIFF
--- a/app/frontend/javascript/utils/google-analytics/index.js
+++ b/app/frontend/javascript/utils/google-analytics/index.js
@@ -96,3 +96,20 @@ export function attachExternalLinkTracker () {
     })
   })
 }
+
+export function attachDetailsOpenTracker () {
+  const detailsComponents = document.querySelectorAll('details')
+  window.dataLayer = window.dataLayer || []
+  detailsComponents.forEach(function (detailsElement) {
+    detailsElement.addEventListener('toggle', function (event) {
+      const summary = detailsElement.querySelector('summary')
+      if (detailsElement.open) {
+        window.dataLayer.push({
+          event: 'details_opened',
+          url: window.location,
+          text: summary.textContent
+        })
+      }
+    })
+  })
+}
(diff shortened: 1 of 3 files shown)


app/frontend/javascript/utils/google-analytics/index.js
```diff
@@ -96,3 +96,20 @@ export function attachExternalLinkTracker () {
...
   })
 }
+
+export function attachDetailsOpenTracker () {
+  const detailsComponents = document.querySelectorAll('details')
+  window.dataLayer = window.dataLayer || []
+  detailsComponents.forEach(function (detailsElement) {
+    detailsElement.addEventListener('toggle', function (event) {
+      const summary = detailsElement.querySelector('summary')
+      if (detailsElement.open) {
```
lfdebrux, line 106: Rather than finding the element and checking its state, can we use the `event.newState` attribute?

DavidBiddle: Ooh, good shout! Will update

DavidBiddle: Argh doesn't look like jsdom has implemented this so it makes the tests sad again 😢

lfdebrux (approved):

app/frontend/javascript/utils/google-analytics/index.js
DavidBiddle: Not a waste of time! I'll raise an issue for jsdom

DavidBiddle: https://github.com/jsdom/jsdom/issues/3879

PR #1408, opened 2025-05-16
Remove notify submission delivery

theseanything:


spec/models/current_request_logging_attributes_spec.rb
```diff
@@ -3,7 +3,6 @@
 RSpec.describe CurrentRequestLoggingAttributes, type: :model do
   subject(:current) { described_class.new }
 
-  let(:submission_email_id) { "a-submission-email-id" }
   let(:submission_email_reference) { "a-submission-email-ref" }
```
stephencdaly, line 6: `submission_email_reference` can also be remove?

stephencdaly (commented):
All the changes make sense, thanks for looking at this.

I think we can delete a few more things:
- app/lib/notify_template_formatter.rb (I think one method is still in use but the rest can go)
- app/mailers/form_submission_mailer.rb

theseanything:
> All the changes make sense, thanks for looking at this.
> 
> I think we can delete a few more things:
> 
> * app/lib/notify_template_formatter.rb (I think one method is still in use but the rest can go)
> * app/mailers/form_submission_mailer.rb

Thanks @stephencdaly - would it be possible to get this PR merged, and remove the other bits in a subsequent PR(s)?

spec/models/current_request_logging_attributes_spec.rb
theseanything: Yup - good catch, removed!

theseanything:
> All the changes make sense, thanks for looking at this.
> 
> I think we can delete a few more things:
> 
> * app/lib/notify_template_formatter.rb (I think one method is still in use but the rest can go)
> * app/mailers/form_submission_mailer.rb

Thanks @stephencdaly - removed both of those too (moved the one used method in notify_template_formatter.rb).

stephencdaly (approved):
Nice work

PR #1410, opened 2025-05-16
Fix naming for methods to record CloudWatch metrics

theseanything:
Methods that were responsible for recording CloudWatch Metrics used the verb "log", which is confusing as "logs" are a different and distinct of observation signal to metrics.


chao-xian (approved):

stephencdaly (approved):
Makes sense to me, nice one

PR #1416, opened 2025-05-16
Move unlocalised content for the accessibility statement

jamie-o-wilkinson:

This PR moves unlocalised content in our accessibility statement into the the locale file, ready for addition of the Welsh locale.

Trello card: https://trello.com/c/EFZCq4bK


stephencdaly (approved):
Changes look good and I've compared the content and it's unchanged. 

I wonder slightly whether for the paragraphs under each heading, we should put all of the content below each heading as HTML in the translation file - or we should name the paragraphs as you have done the list items so that they appear in order in the translation file. I was just thinking it might be slightly more confusing/time consuming to update it if its out of order.

PR #1419, opened 2025-05-19
Move unlocalised content for the cookie policy

jamie-o-wilkinson:

This PR moves unlocalised content in our cookie policy into the the locale file, ready for addition of the Welsh locale.

Trello card: https://trello.com/c/EFZCq4bK

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -196,6 +196,57 @@ en:
     heading: Cookies on GOV.UK Forms
     policy_link: How we use cookies
     reject: Reject analytics cookies
+  cookies:
+    analytics_cookies:
+      heading: Analytics cookies (optional)
+      section_0: We use Google Analytics software to understand how people use GOV.UK Forms. This information helps us to improve our service.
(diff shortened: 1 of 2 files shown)


config/locales/en.yml
```diff
@@ -196,6 +196,57 @@ en:
...
+      section_0: We use Google Analytics software to understand how people use GOV.UK Forms. This information helps us to improve our service.
+      section_1: Google is not allowed to use or share our analytics data with anyone.
+      section_2:
+        caption: 'Google Analytics stores anonymised information about:'
+        list_item_0: the pages you visit
+        list_item_1: how long you spend on each page
+        list_item_2: how you arrived at the site
+        list_item_3: what you click on while you visit the site
+        list_item_4: the device and browser you use
+      section_3:
```
stephencdaly, line 211: Maybe `table` is better for this as it makes it instantly obvious what this content is for? But not a big deal either way

stephencdaly (dismissed):
Changes look good and the content is unchanged

PR #1428, opened 2025-05-21
Remove service name dimension from SubmissionDeliveryLatency

theseanything:
This will always be the same and is unused.

### DIFF
--- a/app/services/cloud_watch_service.rb
+++ b/app/services/cloud_watch_service.rb
@@ -72,7 +72,6 @@ class CloudWatchService
           metric_name: "SubmissionDeliveryLatency",
           dimensions: [
             environment_dimension,
-            service_name_dimension,
             {
               name: "SubmissionDeliveryMethod",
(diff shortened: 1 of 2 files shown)


app/services/cloud_watch_service.rb
jamie-o-wilkinson, line 1: I'm not particularly familiar with the metrics, but isn't the service name included to distinguish metrics that we could (in theory) add to forms-admin or forms-api?

theseanything: It's useful if the metric is emitted from multiple sources, however this metric will always be from forms-runner. 

Having unnecessary dimensions makes querying for these metrics more verbose than they need to be as you can't omit dimensions.

For example, if I want to query CloudWatch for SubmissionDeliveryLatency metric, that request also has to include the "{name: "ServiceName", value: "forms-runner" }" dimension

jamie-o-wilkinson (approved):

PR #1441, opened 2025-05-29
Stop using delivered mail status

jamie-o-wilkinson:

We've found that we can receive a delivery and bounce notification for a given SES message. This means there is potential for a delivery notification to be processed after a bounce notification for a given SES message, which will result in the corresponding `Submission` being deleted.

This PR stops us setting the `mail_status` for Submissions to delivered, and migrates any remaining Submissions with that `mail_status` to pending, in advance of code changes to remove the delivered `mail_status` entirely.

### DIFF
--- a/app/jobs/receive_submission_deliveries_job.rb
+++ b/app/jobs/receive_submission_deliveries_job.rb
@@ -76,7 +76,6 @@ private
   def process_delivery(submission)
     set_submission_logging_attributes(submission)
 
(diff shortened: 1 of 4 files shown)


lfdebrux:
I wonder if we have a possible race condition here... because the migration takes place before the new app is started, submissions could be marked as delivered after the old ones have been changed to pending.

PR #1444, opened 2025-05-29
Select locale based on the language set on the form

stephencdaly:

Trello card: https://trello.com/c/dh1b2Xqk

Set the locale per request from the language attribute for the form if the route is in the scope of a form.

For the cookies and accessibility statement pages, which are not in the scope of a form - include the locale in a `language` query parameter that is included on the links in the footer based on the current locale.

Add a placeholder `cy.yml` locale file. We don't yet have the Welsh translations, so this just includes all keys with the values suffixed by "in Welsh" for testing purposes. End users will not be able to see this as the form needs to be set to be Welsh, and we have not built the functionality to set this. The accessibility statement and cookies page could be viewed with this placeholder text if someone manually added `?language=cy` to the URL.

### DIFF
--- a/app/controllers/forms/base_controller.rb
+++ b/app/controllers/forms/base_controller.rb
@@ -9,7 +9,9 @@ module Forms
     end
 
     rescue_from ActiveResource::ResourceNotFound, Flow::StepFactory::PageNotFoundError, RepeatableStep::AnswerIndexError do
-      render template: "errors/not_found", status: :not_found
+      I18n.with_locale(locale) do
+        render template: "errors/not_found", status: :not_found
+      end
     end
 
     def error_repeat_submission
@@ -48,8 +50,13 @@ module Forms
     end
 
     def set_locale(&action)
-      locale = current_form.respond_to?(:language) ? current_form.language : I18n.default_locale
       I18n.with_locale(locale, &action)
+    end
+
+    def locale
+      return @current_form.language if @current_form.present? && @current_form.respond_to?(:language)
+
+      I18n.default_locale
     end
   end
 end
--- a/spec/requests/forms/base_controller_spec.rb
+++ b/spec/requests/forms/base_controller_spec.rb
@@ -350,46 +350,73 @@ RSpec.describe Forms::BaseController, type: :request do
   end
 
   describe "locale" do
-    before do
-      allow(LogEventService).to receive(:log_form_start)
-      travel_to timestamp_of_request do
+    context "when getting a form page that exists" do
+      before do
         get form_page_path(mode: "form", form_id: 2, form_slug: form_response_data.form_slug, page_slug: 1)
       end
-    end
 
-    context "when the form is English" do
-      it "renders content in English" do
-        expect(response.body).to include('<span class="govuk-details__summary-text">Get help with this form</span>')
+      context "when the form is English" do
+        it "renders content in English" do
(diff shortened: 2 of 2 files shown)


config/application.rb
```diff
@@ -83,5 +83,7 @@ class Application < Rails::Application
     # TODO: remove this when all sensitive data is encrypted
     # See https://guides.rubyonrails.org/active_record_encryption.html#support-for-unencrypted-data
     config.active_record.encryption.support_unencrypted_data = true
+
+    I18n.available_locales = %i[en cy]
```
jamie-o-wilkinson, line 87: I don't think we explicitly set `I18n. default_locale` anywhere, should we here?

The default must itself default to en, but I wonder if we should be explicit now.

app/controllers/application_controller.rb
jamie-o-wilkinson, line 1: This is only tangentially related to this file...

Should/how would we propagate the locale to the errors controller?

jamie-o-wilkinson (commented):
Left a couple of comments. I mentioned it in Slack that I slightly prefer use of locale instead of language, but that's not a strong feeling and I'm pretty persuadable.

config/application.rb
DavidBiddle: It's [documented as the default in the Rails guide](https://guides.rubyonrails.org/i18n.html#configure-the-i18n-module):

> The I18n library will use English as a default locale, i.e. if a different locale is not set, :en will be used for looking up translations.

But don't think there's any harm in being explicit

app/controllers/application_controller.rb
stephencdaly: Ah good point! Completely missed thinking about error pages. I guess we'll have to pass the locale with the redirect? I'll look into it.

config/application.rb
stephencdaly: Have set it explicitly

app/controllers/application_controller.rb
stephencdaly: For the errors managed by the errors controller:

- We can't translate the 404 error as we won't have a form related to the page to get the language from. If we're showing the 404 page for a resource related to a form not being found, we do this form the `rescue_from` in the BaseController, which I've made a change to to ensure it uses the correct locale.
- I can't find a way we can pass the locale automatically with the redirect to the 500 error. As 500 errors are unexpected, and so should be rare and not part of the normal functioning of the application I think that it's probably acceptable for these not to be translated.

config/locales/cy.yml
```diff
@@ -0,0 +1,582 @@
...
+    phase_banner_text_preview-live: You’re previewing the live version of this form in Welsh
+    title_text_preview-archived: Archived preview in Welsh
+    title_text_preview-draft: Draft preview in Welsh
+    title_text_preview-live: Live preview in Welsh
+  page:
+    archived_preview: archived preview in Welsh
+    draft_preview: draft preview in Welsh
+    live_preview: live preview in Welsh
+    none_of_the_above: None of the above in Welsh
+    optional: "(optional) in Welsh"
```
DavidBiddle, line 437: Mental note rather than something to address in this PR: we might not need to translate all of these bits, since they're only shown as part of the form creation journey

DavidBiddle (approved):
Code looks good and I can see the welsh transltions appearing for me as expected locally 🎉

PR #1445, opened 2025-05-30
Use code from GOV.UK Notify for validating email addresses in our email confirmation input object

lfdebrux:

Trello card: https://trello.com/c/tf6aUbYa/2266-make-email-address-validation-consistent 

We've been seeing errors in production from the GOV.UK Notify API from invalid email addresses [[1]]:

```
ValidationError: email_address Not a valid email address (Notifications::Client::BadRequestError)
```

This is happening because when we ask form fillers for an email address for their confirmation email, the validation we do to check whether the email address is valid is much simpler than the validation done by Notify themselves.

It's preferable that we pre-validate the emails on our servers rather than rely on Notify to validate for us, especially if we want to start sending confirmation emails asynchronously.

So we need to have the same validation logic for email addresses as Notify. Unfortunately the rules around what is a valid email address are quite complicated.

This PR copies a bunch of code from the alphagov/notifications-utils repo related to checking if an email address is valid. That code is written in Python, so the code in this commit has been lightly rewritten, but the emphasis was on keeping similarity with the original rather than making it fully idiomatic Ruby. We then wrap it in a very thin Rails-y validator class to make it work with our input objects.

This approach means that hopefully it is easier to stay in sync with Notify if they ever need to change their email address validation.

[1]: https://govuk-forms.sentry.io/issues/6360367242/

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -79,6 +79,9 @@ gem "csv"
 # the hash, also change in package.json
 gem "dfe-autocomplete", require: "dfe/autocomplete", github: "DFE-Digital/dfe-autocomplete", ref: "1d4cc65039e11cc3ba9e7217a719b8128d0e4d53"
 
+# IDNA conversion needed for validating email addresses
+gem "uri-idna"
+
 group :development, :test do
   # See https://guides.rubyonrails.org/debugging_rails_applications.html#debugging-with-the-debug-gem
   gem "debug", platforms: %i[mri mingw x64_mingw]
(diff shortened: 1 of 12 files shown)


chao-xian:
Looks good! Only suggestion I have is maybe rename `NotificationsUtils` to `GOVUKNotificationUtils` to be more specific and indicate the alignment. (We could have otherwise called it `EmailUtils` I guess, but I think my suggestion gives hint to context)

lfdebrux:
> Looks good! Only suggestion I have is maybe rename `NotificationsUtils` to `GOVUKNotificationUtils` to be more specific and indicate the alignment. (We could have otherwise called it `EmailUtils` I guess, but I think my suggestion gives hint to context)

I guess we could... I wanted to keep it as close as possible to what the original code is called.

PR #1454, opened 2025-06-02
Add Welsh confirmation email

DavidBiddle:

Trello card: https://trello.com/c/7RWlFifj/2299-make-confirmation-email-send-in-welsh-for-welsh-forms

Adds code to send a different email template (written in Welsh) when the locale of the request is set to Welsh.

Includes:

- configuring the new template
- Making the month part of the date appear in Welsh (adds the [rails-i18n](https://github.com/svenfuchs/rails-i18n/) gem to do this easily)
- A new support file which should make testing locale-related features easier

I've also added a new template in Notify for this.


stephencdaly (dismissed):
Code looks good and I tested locally and it uses the correct template for Welsh and English

PR #1485, opened 2025-06-16
Enable new type scale

stephencdaly:

Trello card: https://trello.com/c/N6wtyVbT/2339-switch-to-new-type-scale


DavidBiddle (approved):
Changes look good and the fixes are well explained in the commit messages 🎉

deberny:
I've checked the forms-runner using forms admin preview and the type scale on mobile has improved the readability and makes it more consistent when the user comes from the gov.uk page

| Before  | After |
| ------------- |:-------------:|
|   submit forms service gov uk_preview-live_7538_submit-your-sticker-design_27935(iPhone 14 Pro Max) |     
pr-1485 submit review forms service gov uk_preview-draft_4_my-form_17(iPhone 14 Pro Max) |

PR #1538, opened 2025-07-21
Add new renamed columns to generalise the Submission model

theseanything:

This PR adds new columns "delivery_status" and "last_delivery_attempt" to rename "mail_status" and "sent_at". The new names are to generalise the Submission model to allow it to be used to store submission data for all delivery methods (including S3). Therefore needed to remove direct references to the email delivery method.

This is first part of the migration to rename these columns. Here we are adding the columns with the new names, changing the code to use them over the old columns, but also support data in the old columns. 

Future PRs will be raised to migrate existing data in the old columns to the new columns, dropping support for using the columns and the dropping the columns all together.

This also adds an additional "delivered_at" column to represent the timestamp for when a successful deliver is made - useful for implementing our SLOs.

### DIFF
--- a/spec/jobs/delete_submissions_job_spec.rb
+++ b/spec/jobs/delete_submissions_job_spec.rb
@@ -29,7 +29,7 @@ RSpec.describe DeleteSubmissionsJob, type: :job do
            reference: "SENT8DAYS",
            form_id: form_with_file_upload.id,
            form_document: form_with_file_upload,
-           sent_at: 8.days.ago,
+           last_delivery_attempt: 8.days.ago,
            answers: form_with_file_upload_answers
   end
   let!(:sent_submission_sent_7_days_ago) do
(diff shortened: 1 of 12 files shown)


spec/jobs/delete_submissions_job_spec.rb
```diff
@@ -38,25 +38,50 @@
...
   end
   let!(:sent_submission_sent_6_days_ago) do
+    create :submission,
+           :sent,
+           reference: "SENT6DAYS",
+           form_id: form_without_file_upload.id,
+           form_document: form_without_file_upload,
+           last_delivery_attempt: 6.days.ago
+  end
+  let!(:sent_submission_sent_at_7_days_ago) do
```
chao-xian, line 51: I'm confused and don't understand why we're testing this specifically twice now?

PR #1542, opened 2025-07-24
Strip emails

thomasiles:
### Remove surrounding white space from entered email addresses

https://govuk-forms.sentry.io/issues/6656667020/?referrer=slack&notification_uuid=a6077a95-ff38-41c6-be36-f9fc2b225e20&alert_rule_id=14126387&alert_type=issue

We are seeing occasional uncaught exceptions from when trying to send confirmation emails.  

This can be [recreated in dev](https://govuk-forms.sentry.io/issues/6656667020/?project=6576261&query=is%3Aunresolved&referrer=issue-stream) by adding "/n" to the end of the confirmation input. I'm cheating using a proxy to modify the post request. Any odd whitespace before or after the email address works, like /n, /v, /r, tabs and spaces don't.

I don't know how actual users are sending values like that. The two live issues are from iphones, which is probably a clue. I would guess it's some kind of iphone keyboard, autocomplete, paste or autofill thing adding odd line endings. I don't think input boxes are ever supposed to span multiple lines though and googling isn't helping.

Why doesn't our validation code catch this? We [strip whitespace during validation](https://github.com/alphagov/forms-runner/blob/main/lib/notifications_utils/recipient_validation/email_address.rb) so our validation rules don't see it. The validation code doesn't actually modify the value though, only add errors. The [value from the input object](https://github.com/alphagov/forms-runner/blob/1fd074afa9863321db2e52d2d910a44c2ff8d4ed/app/input_objects/email_confirmation_input.rb#L10) will still have whatever random spaces before/after the email, even though it shows as valid.

The Mail gem seems happy to remove leading/trailing spaces and tabs but newlines change it from being an array into being a string. I don't understand the field logic either or why it converts it to a string.

This PR stops stripping whitespace in the email validation. Instead it removes the whitespace in the confirmation email input object and email question type.


lfdebrux (approved):

DavidBiddle (approved):
Good detective work 🙌 code looks good

PR #1547, opened 2025-07-25
Add rake task to migrate data from old columns in Submissions

theseanything:
"delivery_status" is to replace "mail_status", and "last_delivery_attempt" is to replace "sent_at". To drop the old column, this adds a rake task to ensure that data in the new columns is in sync with the old columns. After this task is run, we can remove usage and drop the old columns.

### DIFF
--- a/lib/tasks/submissions.rake
+++ b/lib/tasks/submissions.rake
@@ -5,6 +5,32 @@ namespace :submissions do
     Rails.logger.info "#{Submission.bounced.count} bounced submissions"
   end
 
+  # TODO: Remove this task after mail_status and sent_at are removed from the database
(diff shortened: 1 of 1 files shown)


DavidBiddle (dismissed):
Change looks good anyd works for me locally. The new task might benefit from some tests but I think that's optional given that it's a oneoff task we'll be deleting almost immediately.

theseanything:
> Change looks good anyd works for me locally. The new task might benefit from some tests but I think that's optional given that it's a oneoff task we'll be deleting almost immediately.

Yeah, I've manually run this and confident this runs as expected - don't think it's worth the effort for adding the tests.

PR #1568, opened 2025-08-05
Add form_id to Sentry msg for bounced submissions

thomasiles:
### Add form_id to Sentry message for bounced submissions

Trello card: https://trello.com/c/iG4lt8a0/2350-bug-investigate-bounced-email-alerts

Since we switched delivering submissions from Notify to AWS SES we have received around 10 bounce notifications.

We need to track failed deliveries to learn more about the cause of failures.

We log to sentry when a submission is bounced. This is useful for debugging submissions but produces a lot of alerts.

We've learned that submission email addresses which have an auto-reply will trigger a transient bounce and a successful delivery on every form submission.

To prevent this we have asked form creators to stop sending auto-replies from submission addresses. We also plan to make changes to our MOU.

We still want to keep an eye on bounced submissions because it's such an important part of the service.

We don't have a way to easily separate failed submissions from bounced submissions which were also delivered.

To help us separate forms which we know have an auto-reply this commit adds the form_id to the Sentry message for bounced submissions.

This should make it easier for us to manage the issues in sentry and highlight new bounces from known cases.

### DIFF
--- a/app/jobs/receive_submission_bounces_and_complaints_job.rb
+++ b/app/jobs/receive_submission_bounces_and_complaints_job.rb
@@ -101,7 +101,7 @@ private
     EventLogger.log_form_event("submission_bounced", ses_bounce: ses_bounce.merge(bounced_recipients:))
 
     unless submission.preview?
-      Sentry.capture_message("Submission email bounced - #{self.class.name}:", extra: {
+      Sentry.capture_message("Submission email bounced #{submission.form_id} - #{self.class.name}:", extra: {
         form_id: submission.form_id,
         submission_reference: submission.reference,
         job_id:,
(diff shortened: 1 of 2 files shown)


app/jobs/receive_submission_bounces_and_complaints_job.rb
```diff
@@ -101,7 +101,7 @@ def process_bounce(submission, ses_message)
     EventLogger.log_form_event("submission_bounced", ses_bounce: ses_bounce.merge(bounced_recipients:))
 
     unless submission.preview?
-      Sentry.capture_message("Submission email bounced - #{self.class.name}:", extra: {
+      Sentry.capture_message("Submission email bounced #{submission.form_id} - #{self.class.name}:", extra: {
```
stephencdaly, line 104: Would it be useful to indicate it's the form ID?

```suggestion
      Sentry.capture_message("Submission email bounced for form #{submission.form_id} - #{self.class.name}:", extra: {
```

PR #1587, opened 2025-08-15
Remove deprecated route and controller method

theseanything:
The deprecated route (/deprecated) and associated controller method
(ErrorsController#deprecated) are no longer needed since V1 API
usage has been removed.


theseanything:
> @theseanything I think there's some other code from #942 still lying around, including [config/initializers/forms_api_deprecations.rb](https://github.com/alphagov/forms-runner/blob/main/config/initializers/forms_api_deprecations.rb).

Removed that and the added active support deprecation warnings config.

lfdebrux (approved):

PR #1597, opened 2025-08-18
Access page_slug from page inside Step class

theseanything:
This removes the separate page_slug instance variable, which was passed into the initializer for Step via the StepFactory. As the original Page object is also passed in, we can reference it directly from that object. 

This helps simplify the code and makes it easier for future refactoring.


stephencdaly (approved):
Changes make sense, it's nice to get rid of this duplication. As discussed it would be good to rename all the parameters called `page_slug` next.

PR #1636, opened 2025-09-10
Ensure carriage returns are counted as 1 character for validation

stephencdaly:

Trello card: https://trello.com/c/rWMe8Qd4

We’re counting line breaks in long text inputs as 2 characters due to carriage returns (CR) as well as line feed characters (LF) being sent in the HTTP request when the form is submitted for line breaks.

Fix this by stripping our carriage returns from the input before doing the length validation.


app/models/question/text.rb
```diff
@@ -5,12 +5,20 @@ class Text < QuestionBase
...
       answer_settings.input_type == "single_line"
     end
 
     def has_long_answer?
       !is_single_line?
     end
+
+  private
+
+    def strip_carriage_returns
```
DavidBiddle, line 20: We call this `strip_carriage_returns!` in forms-admin (I guess because it mutates the input?) - should it be the same here / should we remove the `!` in admin?

stephencdaly: Good point, have added the `!`

PR #1665, opened 2025-10-01
Save SES message delivery time to submission record

lfdebrux:

Keep a record of the message delivery time in our database for reference, so it's clear when inspecting submissions with the Rake task whether or not a submission has been delivered.

### DIFF
--- a/app/jobs/receive_submission_deliveries_job.rb
+++ b/app/jobs/receive_submission_deliveries_job.rb
@@ -60,10 +60,10 @@ private
 
       submission = Submission.find_by!(mail_message_id: ses_message_id)
 
-      process_delivery(submission)
+      delivery_time = Time.zone.parse(ses_message["delivery"]["timestamp"])
+      process_delivery(submission, delivered_at: delivery_time)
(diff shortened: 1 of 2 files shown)


app/jobs/receive_submission_deliveries_job.rb
```diff
@@ -60,10 +60,10 @@ def process_messages(sqs_client, queue_url, messages)
 
       submission = Submission.find_by!(mail_message_id: ses_message_id)
 
-      process_delivery(submission)
+      delivery_time = Time.zone.parse(ses_message["delivery"]["timestamp"])
+      process_delivery(submission, delivered_at: delivery_time)
```
theseanything, line 64: Do we need to put a conditional to not set the delivery_time if the submission has bounced? (Due to race condition in received SES delivery notifications for asynchronous bounces)

lfdebrux: I think it's fine if the state is `bounced` but the delivery time is set, that's just recording what actually happened, and we only look at the state for handling bounced submissions anyway.

PR #1859, opened 2026-01-23
Add CI GitHub Actions workflow

theseanything:
Replace the `build_and_test` and `docker_build` workflows with a single `ci` workflow.  Each CI process now runs as an independent job, so checks execute in parallel and report failures individually in PR status checks.

This improves developer feedback by showing all failing CI areas at once (instead of masking later failures when an earlier step fails), and should reduce total CI runtime.


lfdebrux:
@theseanything is this PR something you're still working on? Can it be closed?

PR #1918, opened 2026-02-23
Move confirmation email reference generation to job runtime

theseanything:
Move confirmation email reference creation from EmailConfirmationInput request-time handling into SendConfirmationEmailJob, so the Notify reference is generated when the job executes only when we need it.

Removes a lot of unnecessary code and makes EmailConfirmationInput model simpler.

We don't need to log the notify_response_id at request time, as request will not fail due a notify error - so little use to correlate them.


theseanything:
work in https://github.com/govuk-forms/forms-runner/pull/2164 should do this.

PR #2028, opened 2026-04-08
Use FormDocumentResource model within Form model

stephencdaly:

Trello card: https://trello.com/c/02H0xEqG

Previously, the Form model was an ActiveResource model that was constructed using the JSON retrieved from the forms-admin API using the FormDocumentResource model.

Make it a plain Ruby object and instead delegate attributes to the FormDocumentResource model. This means we are not using the ActiveResource model directly in the rest of the codebase, and instead consuming the Form model which has a defined interface. Future changes to the API can be handled more easily within the Form model.

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -1,22 +1,33 @@
-class Form < ActiveResource::Base
-  self.site = Settings.forms_api.base_url
-  self.prefix = "/api/v2/"
-  self.include_format_in_path = false
+class Form
+  attr_reader :document_json
+  private attr_reader :form_document
 
-  class Step < ActiveResource::Base
-    self.site = Form.site
-    self.prefix = Form.prefix_source
-    self.include_format_in_path = false
+  def initialize(form_document, form_document_json = nil)
+    @form_document = form_document
+    @document_json = form_document_json
   end
 
-  has_many :steps, class_name: "Api::V2::StepResource"
-  attr_accessor :document_json
-
-  alias_method :form_document_steps, :steps
-
-  def form_id
-    @attributes["form_id"] || @attributes["id"]
-  end
+  delegate :steps, to: :form_document, prefix: true
+  delegate :declaration_text,
+           :form_id,
+           :form_slug,
+           :name,
+           :payment_url,
+           :privacy_policy_url,
+           :s3_bucket_aws_account_id,
+           :s3_bucket_name,
+           :s3_bucket_region,
+           :start_page,
+           :send_daily_submission_batch,
+           :send_weekly_submission_batch,
+           :submission_email,
+           :submission_type,
+           :support_email,
+           :support_phone,
+           :support_url,
+           :support_url_text,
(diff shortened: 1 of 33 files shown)


app/models/form.rb
```diff
@@ -1,22 +1,33 @@
...
+           :send_daily_submission_batch,
+           :send_weekly_submission_batch,
+           :submission_email,
+           :submission_type,
+           :support_email,
+           :support_phone,
+           :support_url,
+           :support_url_text,
+           :what_happens_next_markdown,
+           to: :form_document
```
lfdebrux, line 30: This could probably be simplified with [delegate_missing_to](https://api.rubyonrails.org/classes/Module.html#method-i-delegate_missing_to)

```suggestion
  delegate_missing_to :form_document
```

spec/models/form_spec.rb
```diff
@@ -173,26 +163,36 @@
     end
   end
 
-  describe "#multilingual?" do
+  describe "#multilingual" do
```
lfdebrux: I think we still want the question mark here, as it matches the name of the method, and the method is a predicate

```suggestion
  describe "#multilingual?" do
```

spec/factories/models/forms.rb
```diff
@@ -19,40 +17,26 @@
...
     s3_bucket_aws_account_id { nil }
 
-    trait :new_form do
-      submission_email { nil }
-      privacy_policy_url { nil }
-    end
-
-    trait :ready_for_live do
-      with_pages
+    trait :live? do
```
lfdebrux: You've renamed the `:live?` to `:live` in the `form_document_resource` factory, might be worth having this trait be called `:live` without the question mark as well?

stephencdaly: Thanks, I thought I'd done that but might have been on a different branch

spec/models/form_spec.rb
stephencdaly: Ah thanks, must have got lost when I went back and forth on some changes

app/models/form.rb
stephencdaly: I was thinking that by defining what can be delegated it's easier to know in the rest of the code what attributes the `Form` has. Otherwise we have to look in forms-runner to find out?

lfdebrux: Fair enough!

spec/factories/models/forms.rb
stephencdaly: Ah it was already renamed - it's just done in the following commit in the PR

spec/models/form_spec.rb
stephencdaly: Updated

lfdebrux (approved):

spec/factories/models/forms.rb
lfdebrux: Oh woops sorry 🤦

PR #2030, opened 2026-04-09
Only consume StepResources when building the Journey

stephencdaly:

Trello card: https://trello.com/c/02H0xEqG

The intention behind this PR is to only consume `Api::V2::StepResources` when we need them to construct `Step` models. In all other places in the code, we should use the `Step` or `QuestionBase` models to get information about steps, in order to decouple from the API and make it easier to make changes to the API in future.

To achieve this, construct the `Journey` using a `Api::V2::FormDocumentResource` so that we don't need to expose the `Api::V2::StepResources` via the `Form` model. Whenever we need to access steps in other areas of the code, get them from the `Journey`.

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -5,7 +5,6 @@ class Form
     @form_document = form_document
   end
 
-  delegate :steps, to: :form_document, prefix: true
   delegate :declaration_text,
            :form_id,
            :form_slug,
@@ -31,10 +30,6 @@ class Form
 
   def document_json
     form_document.as_json
-  end
-
-  def step_by_id(step_id)
-    form_document_steps.find { |s| s.id == step_id }
   end
 
   def payment_url_with_reference(reference)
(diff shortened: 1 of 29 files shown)


app/models/form.rb
```diff
@@ -31,6 +29,10 @@ def initialize(form_document, form_document_json = nil)
 
   alias_method :id, :form_id
 
+  def document_json
+    form_document.as_json
+  end
```
lfdebrux, line 33: It might be worth adding a test for `Form#document_json` that checks that the JSON that comes out is similar enough to the JSON that goes in to the form document...

stephencdaly: Yeh good point. I realised after making this change we will actually be setting some defaults on the StepResource after loading it in - but that should be fine I think. I'll add a test - although can always revert to storing it if we're not sure.

stephencdaly: Ok, I've added some basic tests to `form_document_resource_spec` and `form_spec`. Do you think these cover it?

lfdebrux: Okay yeah I think that covers it

lfdebrux (approved):

PR #2057, opened 2026-04-27
Add integration with One Login for getting a copy of answers

stephencdaly:

Trello card: https://trello.com/c/tKtA1VPQ

This PR adds in a basic integration with One Login which allows the user to log in and then stores their email from One Login on the session.

The basics are:
- We've added a new page after the copy of answers page with a button to continue to one login if the user selected "Yes". This is required because for OmniAuth the login request needs to be initiated by a POST request from the browser to the auth endpoint.
- We're using the https://github.com/OfficeForProductSafetyAndStandards/omniauth-govuk-one-login gem for the One Login integration. This gem isn't currently published so it's pinned to a commit SHA
- We store the form details on the session so we know where to send the user after they're returned to us from One Login.
- We handle the callback from One Login by storing the email for the user and then sending them to the check your answers page using the form details we stored on the session.

Not currently implemented:
- Currently for errors we're showing generic 404 or 500 error pages. We're currently designing more helpful error pages, so will update this in a future PR.
- Logging the user out of One Login.
- Welsh translations

### Local testing

You need a one login account with a service (https://admin.sign-in.service.gov.uk/)
In your service you need to:
- add a public key ([see docs for creating key pair](https://docs.sign-in.service.gov.uk/before-integrating/set-up-your-public-and-private-keys/#create-a-key-pair))
- add the Redirect URI http://localhost:3001/auth/govuk_one_login/callback

Run forms runner with `SETTINGS__GOVUK_ONE_LOGIN__CLIENT_ID=your-client-id SETTINGS__GOVUK_ONE_LOGIN__PRIVATE_KEY=base-64-encoded-private-key bundle exec rails s`.

You can base64 encode your private key using `base64 -i private_key.pem -o private_key_base64.pem`

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -90,6 +90,9 @@ gem "rails-i18n", "~> 8.1"
 # IDNA conversion needed for validating email addresses
 gem "uri-idna"
 
+gem "omniauth_govuk_one_login", github: "OfficeForProductSafetyAndStandards/omniauth-govuk-one-login", ref: "6c6b68e186bd7ae08d6c16b8983ddad1eeb6cfc7"
+gem "omniauth-rails_csrf_protection"
+
 group :development, :test do
   # See https://guides.rubyonrails.org/debugging_rails_applications.html#debugging-with-the-debug-gem
   gem "debug", platforms: %i[mri windows], require: "debug/prelude"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -11,6 +11,16 @@ GIT
       view_component
 
 GIT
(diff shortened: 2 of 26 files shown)


DavidBiddle (approved):
Code looks good and well organised, have tested with local One Login config and it works as expected - I see the One Login journey, and when I register or sign in I'm redirected back to the correct page 🎉

PR #2069, opened 2026-04-28
Allow routing to work with multiple conditions

lfdebrux:

Trello card: https://trello.com/c/Ndp9ZPrn/3062-update-forms-runner-to-work-with-more-than-one-routing-condition-per-question 

Update the logic in Step#next_step_slug_after_routing so that it will match a routing condition even if it isn't the first condition. This should allow forms-runner to support forms with multiple branches.

Note that this still won't match a "default" condition (answer_value of `nil`) that isn't the first in the list of routing conditions; our plans for routing don't currently have a use case for a default condition on a selection question, so there shouldn't be a situation where there are multiple conditions and one of them is a default condition.

### DIFF
--- a/app/models/step.rb
+++ b/app/models/step.rb
@@ -86,8 +86,8 @@ class Step
       return goto_condition_step_slug(routing_conditions.first)
     end
 
-    if first_condition_matches?
-      return goto_condition_step_slug(routing_conditions.first)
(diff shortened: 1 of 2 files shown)


thomasiles (approved):
looks really nice and clear 🏅

PR #2070, opened 2026-04-28
Log the user out of One Login when their form is submitted

stephencdaly:

Trello card: https://trello.com/c/tKtA1VPQ/2858-create-integration-with-one-login

As per the [One Login technical documentation](https://docs.sign-in.service.gov.uk/integrate-with-integration-environment/managing-your-users-sessions/#build-functionality-to-log-your-user-out), we should build functionality to log the user out to prevent risk of users accidentally sharing sessions. The docs say this can be done by coding the submit button to automatically log the user out.

Handle logging the user out by:

1. Storing the auth token we get from One Login when the user logs in, this is needed to log out
2. Redirecting the user to the One Login logout URL when they click the submit button, after we've handled submission
3. Handling the user being returned to us after logging out to show them the form submitted page

If the user's session cookie has been lost and we're unable to determine the form path parameters after they've been returned to us, let an exception bubble up and show an internal error page for now. We're currently discussing whether we can show something more friendly to the user.

### Local testing

Add `http://localhost:3001/auth/logged-out` to the `Post logout redirect URIs (optional)` in the [one login admin tool](https://admin.sign-in.service.gov.uk/)

Screenshot 2026-04-29 at 15 27 35

To see that it's logged you out, you can complete a form and log in with One Login. After submitting, try to complete another form and you should be asked to log in again if you want a copy of your answers.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -118,4 +118,8 @@ private
 
     false
   end
+
+  def auth_store
+    @auth_store ||= Store::AuthStore.new(session)
+  end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -119,11 +119,7 @@ private
     false
   end
 
-  def auth_store
-    @auth_store ||= Store::AuthStore.new(session)
-  end
-
-  def return_from_one_login_store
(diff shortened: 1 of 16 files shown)


DavidBiddle (approved):
Code looks good and the logout works for me locally 🎉

PR #2071, opened 2026-04-29
Add error pages for one login journey

stephencdaly:

Trello card: https://trello.com/c/tKtA1VPQ/2858-create-integration-with-one-login

Add error pages for the following scenarios:

**The auth details are missing or there is an unexpected error authorising with One Login**

Screenshot 2026-04-29 at 16 57 51

**The user's session has been lost when they're returned to us from One Login**

Screenshot 2026-04-29 at 17 01 45

**The user's session has been lost after they've been logged out when they submit their form**

In this case, rather than showing an error, show a generic success page without the form details as the user's form will still have been submitted and we should have sent them an email with their answers.

Screenshot 2026-04-29 at 17 03 46

### DIFF
--- a/app/views/errors/auth_error.html.erb
+++ b/app/views/errors/auth_error.html.erb
@@ -1,0 +1,9 @@
+<% set_page_title(t('errors.auth_error.title'), t("gov_uk_forms")) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l"><%= t('errors.auth_error.title') %></h1>
+
+    <%= t("errors.auth_error.body_html", link_url:) %>
+  </div>
+</div>
--- a/app/views/errors/return_from_one_login_session_missing.html.erb
+++ b/app/views/errors/return_from_one_login_session_missing.html.erb
@@ -1,0 +1,9 @@
+<% set_page_title(t('errors.return_from_one_login_session_missing.title'), t("gov_uk_forms")) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l"><%= t('errors.return_from_one_login_session_missing.title') %></h1>
+
+    <%= t("errors.return_from_one_login_session_missing.body_html") %>
+  </div>
+</div>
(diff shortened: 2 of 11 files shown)


stephencdaly:
@StephenGill could you double-check the content is correct in this one please?

StephenGill:
Copy looks good to me

app/views/errors/auth_error.html.erb
```diff
@@ -0,0 +1,10 @@
+<% set_page_title(t('errors.auth_error.title'), t("gov_uk_forms")) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l"><%= t('errors.auth_error.title') %></h1>
+    <p class="govuk-body">
+      <%= t("errors.auth_error.body_html", link_url:) %>.
+    </p>
```
DavidBiddle: The content inside this `<p>` tag is already wrapped in `<p>` tags, so this outer `<p>` tag can be omitted. 

There's also a `.` at the end of the line - because the content before this is in a `<p>` tag, it's rendered on a new line. I think this can be removed too:

```suggestion
    <%= t("errors.auth_error.body_html", link_url:) %>
```

app/views/errors/return_from_one_login_session_missing.html.erb
```diff
@@ -0,0 +1,10 @@
+<% set_page_title(t('errors.return_from_one_login_session_missing.title'), t("gov_uk_forms")) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l"><%= t('errors.return_from_one_login_session_missing.title') %></h1>
+    <p class="govuk-body">
+      <%= t("errors.return_from_one_login_session_missing.body_html") %>.
+    </p>
```
DavidBiddle: Same as above:

```suggestion
    <%= t("errors.return_from_one_login_session_missing.body_html") %>
```

DavidBiddle (commented):
All looks good aside from those HTML issues raised above - I've tested locally and everything else works as expected

app/views/errors/auth_error.html.erb
stephencdaly: Oops 🤦. I've updated now, thanks for spotting.

DavidBiddle (approved):

PR #2082, opened 2026-05-06
Store bounce details on delivery

stephencdaly:

Trello card: https://trello.com/c/CsEqMtDL

We want to send a daily notification of any bounced submission emails to the admins for a form. In the notification email, we want to include different content depending on whether we can identify the bounce as a hard or soft bounce, with the scope to be more specific in the future for different bounce types.

Add a `failure_details` jsonb column to the deliveries table to store additional failure details when a delivery fails.


lfdebrux (approved):
Changes look good to me

PR #2083, opened 2026-05-07
Make confirmation email for Welsh bi-lingual

chao-xian:

When a user answers Welsh questions, the confirmation email now uses a bilingual Notify template. The mailer passes both English and Welsh personalisation variables (_cy variants for submission_date, what_happens_next_text, and support_contact_details) so each language section of the template receives the correct content. Template selection is driven by the explicit submission_locale parameter rather than the ambient I18n locale.

Trello card:  https://trello.com/c/NLqWmwvf/2894-update-confirmation-emails-to-cater-for-new-welsh-form-submissions 

### DIFF
--- a/app/jobs/send_confirmation_email_job.rb
+++ b/app/jobs/send_confirmation_email_job.rb
@@ -5,19 +5,21 @@ class SendConfirmationEmailJob < ApplicationJob
   def perform(submission:, notify_response_id:, confirmation_email_address:)
     set_submission_logging_attributes(submission:)
 
-    I18n.with_locale(submission.submission_locale || I18n.default_locale) do
-      form = submission.form
-      mail = FormSubmissionConfirmationMailer.send_confirmation_email(
-        what_happens_next_markdown: form.what_happens_next_markdown,
-        support_contact_details: form.support_details,
-        notify_response_id:,
-        confirmation_email_address:,
-        mailer_options: mailer_options_for(submission:, form:),
-      )
+    form = submission.form
+    welsh_form = fetch_welsh_form(submission:, form:)
+    mail = FormSubmissionConfirmationMailer.send_confirmation_email(
+      what_happens_next_markdown: form.what_happens_next_markdown,
+      what_happens_next_markdown_cy: welsh_form&.what_happens_next_markdown,
+      support_contact_details: form.support_details,
+      support_contact_details_cy: welsh_form&.support_details,
+      notify_response_id:,
+      confirmation_email_address:,
+      mailer_options: mailer_options_for(submission:, form:),
+      submission_locale: submission.submission_locale,
+    )
 
-      mail.deliver_now
-      CurrentJobLoggingAttributes.confirmation_email_id = mail.govuk_notify_response.id
-    end
+    mail.deliver_now
+    CurrentJobLoggingAttributes.confirmation_email_id = mail.govuk_notify_response.id
   rescue StandardError
     CloudWatchService.record_job_failure_metric(self.class.name)
     raise
@@ -33,5 +35,16 @@ private
       submission_reference: submission.reference,
       payment_url: submission.payment_url,
     )
+  end
+
+  def fetch_welsh_form(submission:, form:)
+    return nil unless submission.submission_locale.to_sym == :cy
+
+    form_document = Api::V2::FormDocumentRepository.find_with_mode(
+      form_id: form.id,
+      mode: submission.mode_object,
+      language: :cy,
+    )
+    Form.new(form_document) if form_document
   end
 end
--- a/app/mailers/form_submission_confirmation_mailer.rb
+++ b/app/mailers/form_submission_confirmation_mailer.rb
@@ -1,13 +1,17 @@
 class FormSubmissionConfirmationMailer < GovukNotifyRails::Mailer
-  def send_confirmation_email(what_happens_next_markdown:, support_contact_details:, notify_response_id:, confirmation_email_address:, mailer_options:)
+  def send_confirmation_email(what_happens_next_markdown:, support_contact_details:, notify_response_id:, confirmation_email_address:, mailer_options:, submission_locale: :en, what_happens_next_markdown_cy: nil, support_contact_details_cy: nil)
+    @submission_locale = submission_locale.to_sym
     set_template(template_id)
 
     set_personalisation(
       title: mailer_options.title,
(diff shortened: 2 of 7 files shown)


spec/mailers/form_submission_confirmation_mailer_spec.rb
```diff
@@ -30,24 +32,18 @@
...
 
-    context "when the request locale is not set" do
-      it "uses the English language template" do
-        expect(mail.govuk_notify_template).to eq("123456")
-      end
-    end
-
-    context "when the request locale is set to :en" do
-      include_context "with locale set to :en"
+    context "when submission_locale is not set" do
```
stephencdaly: does this context name need to change as the submission_locale is set, to :en?

app/mailers/form_submission_confirmation_mailer.rb
```diff
@@ -1,13 +1,17 @@
...
+  def send_confirmation_email(what_happens_next_markdown:, support_contact_details:, notify_response_id:, confirmation_email_address:, mailer_options:, submission_locale: :en, what_happens_next_markdown_cy: nil)
+    @submission_locale = submission_locale.to_sym
     set_template(template_id)
 
     set_personalisation(
       title: mailer_options.title,
       what_happens_next_text: what_happens_next_markdown.presence || default_what_happens_next_text,
+      what_happens_next_text_cy: what_happens_next_markdown_cy.presence || what_happens_next_markdown.presence || default_what_happens_next_text,
       support_contact_details: format_support_details(support_contact_details).presence || default_support_contact_details_text,
+      support_contact_details_cy: format_support_details(support_contact_details, locale: :cy).presence || default_support_contact_details_text,
```
stephencdaly: We'll also need to pass in the support details from the Welsh form document, as different values can be provided for all of the support fields. Currently this is just using the English details and translating the call charges.

app/jobs/send_confirmation_email_job.rb
```diff
@@ -5,19 +5,20 @@ class SendConfirmationEmailJob < ApplicationJob
...
-
-      mail.deliver_now
-      CurrentJobLoggingAttributes.confirmation_email_id = mail.govuk_notify_response.id
-    end
+    form = submission.form
+    mail = FormSubmissionConfirmationMailer.send_confirmation_email(
+      what_happens_next_markdown: form.what_happens_next_markdown,
+      what_happens_next_markdown_cy: welsh_what_happens_next_for(submission:, form:),
+      support_contact_details: form.support_details,
+
```
stephencdaly: I don't think we want this blank line

stephencdaly (commented):
Changes look good, we just need to make sure we're passing in the Welsh support details like we're passing in the Welsh "what happens next"

spec/mailers/form_submission_confirmation_mailer_spec.rb
chao-xian: Good spot thank you! Fixed

app/mailers/form_submission_confirmation_mailer.rb
chao-xian: Thanks for spotting that! Have amended so we are fetching the Welsh Form and using that

spec/jobs/send_confirmation_email_job_spec.rb
```diff
@@ -80,6 +102,20 @@
...
+    it "passes the Welsh what happens next to the mailer" do
+      allow(FormSubmissionConfirmationMailer).to receive(:send_confirmation_email).and_call_original
+
+      described_class.perform_now(
+        submission:,
+        notify_response_id:,
+        confirmation_email_address:,
+      )
+
+      expect(FormSubmissionConfirmationMailer).to have_received(:send_confirmation_email).with(
```
stephencdaly, line 112: Should we update this test to check it also passes the Welsh support details?

app/jobs/send_confirmation_email_job.rb
```diff
@@ -5,19 +5,21 @@ class SendConfirmationEmailJob < ApplicationJob
...
-        confirmation_email_address:,
-        mailer_options: mailer_options_for(submission:, form:),
-      )
+    form = submission.form
+    welsh_form = fetch_welsh_form(submission:, form:)
+    mail = FormSubmissionConfirmationMailer.send_confirmation_email(
+      what_happens_next_markdown: form.what_happens_next_markdown,
+      what_happens_next_markdown_cy: welsh_form&.what_happens_next_markdown,
+      support_contact_details: form.support_details,
+      support_contact_details_cy: welsh_form&.support_details || form.support_details,
```
stephencdaly: should we be consistent with `what_happens_next_markdown_cy` and fall back to using the English support details as default in the mailer, and add a test for it?

spec/mailers/form_submission_confirmation_mailer_spec.rb
```diff
@@ -64,10 +62,37 @@
...
+        expect(mail.govuk_notify_personalisation[:what_happens_next_text_cy]).to eq("Please wait for a response")
+      end
+    end
+
     it "includes the forms support contact details" do
       expect(mail.govuk_notify_personalisation[:support_contact_details]).to eq("0203 222 2222\n\n[Find out about call charges](https://www.gov.uk/call-charges)")
     end
 
+    it "includes the Welsh support contact details with Welsh call charges link" do
+      expect(mail.govuk_notify_personalisation[:support_contact_details_cy]).to eq("0203 222 2222\n\n[Darganfyddwch am gostau galwadau](https://www.gov.uk/call-charges)")
```
stephencdaly: It might be good to change the support phone number for the Welsh in the setup for this test to not be the same as the English so that it's testing that it's using the Welsh support details

stephencdaly (commented):
Changes do what we need. I've suggested a few improvements to the tests for the additional logic added for the Welsh support details.

spec/jobs/send_confirmation_email_job_spec.rb
chao-xian: Good point have added

spec/mailers/form_submission_confirmation_mailer_spec.rb
chao-xian: Done

app/jobs/send_confirmation_email_job.rb
chao-xian: Done

stephencdaly (approved):

PR #2110, opened 2026-05-19
Remove MailerOptions Data object used by confirmation email mailer

stephencdaly:

Trello card: https://trello.com/c/NLqWmwvf

This Data object existed before the Submission model existed and was used to pass common data used by the submission and confirmation emails.

It is no longer needed, and we can instead pass the Submission and Form objects to the mailer to get the data we need.

This means we don't need to pass attributes for all the form attributes that have Welsh translations.


chao-xian (approved):
Nice! Ty. Works locally and code changes look good

PR #2113, opened 2026-05-22
Copy of answers

chao-xian:

Updates the flow for Copy Of Answers feature and switches from a global feature flag to checking against a value set on the form whether to show that feature to the form filler.

Trello card: https://trello.com/c/KZuqH0bA

### DIFF
--- a/app/controllers/forms/check_your_answers_controller.rb
+++ b/app/controllers/forms/check_your_answers_controller.rb
@@ -11,22 +11,24 @@ module Forms
       return redirect_to form_step_path(current_context.form.id, current_context.form.form_slug, current_context.next_step_slug, nil) unless current_context.can_visit?(CheckYourAnswersStep::CHECK_YOUR_ANSWERS_STEP_SLUG)
 
       setup_check_your_answers
-      email_confirmation_input = EmailConfirmationInput.new
+      skip_confirmation_email = skip_confirmation_email?
+      email_confirmation_input = skip_confirmation_email ? EmailConfirmationInput.new(send_confirmation: "skip_confirmation") : EmailConfirmationInput.new
 
       @support_details = current_context.support_details
 
-      render template: "forms/check_your_answers/show", locals: { email_confirmation_input: }
+      render template: "forms/check_your_answers/show", locals: { email_confirmation_input:, skip_confirmation_email: }
     end
 
     def submit_answers
       @support_details = current_context.support_details
-      email_confirmation_input = EmailConfirmationInput.new(email_confirmation_input_params)
+      skip_confirmation_email = skip_confirmation_email?
+      email_confirmation_input = skip_confirmation_email ? EmailConfirmationInput.new(send_confirmation: "skip_confirmation") : EmailConfirmationInput.new(email_confirmation_input_params)
       requested_email_confirmation = email_confirmation_input.send_confirmation == "send_email"
 
       unless email_confirmation_input.valid?
         setup_check_your_answers
 
-        return render template: "forms/check_your_answers/show", locals: { email_confirmation_input: }, status: :unprocessable_content
+        return render template: "forms/check_your_answers/show", locals: { email_confirmation_input:, skip_confirmation_email: }, status: :unprocessable_content
       end
 
       return redirect_to error_repeat_submission_path(@form.id) if current_context.form_submitted?
@@ -52,7 +54,7 @@ module Forms
       rescue FormSubmissionService::ConfirmationEmailToAddressError
         setup_check_your_answers
         email_confirmation_input.errors.add(:confirmation_email_address, :invalid_email)
-        render template: "forms/check_your_answers/show", locals: { email_confirmation_input: }, status: :unprocessable_content
+        render template: "forms/check_your_answers/show", locals: { email_confirmation_input:, skip_confirmation_email: }, status: :unprocessable_content
       end
     rescue StandardError => e
       log_rescued_exception(e)
@@ -76,8 +78,12 @@ module Forms
       end
     end
 
+    def skip_confirmation_email?
+      current_context.wants_copy_of_answers? && current_context.get_copy_of_answers_email_address.present?
+    end
+
     def back_link
-      if FeatureService.enabled?("filler_answer_email_enabled") && current_context.form.copy_of_answers_enabled?
+      if current_context.form.copy_of_answers_enabled?
         copy_of_answers_path(form_id: current_context.form.id, form_slug: current_context.form.form_slug)
       else
         previous_step = current_context.previous_step(CheckYourAnswersStep::CHECK_YOUR_ANSWERS_STEP_SLUG)
--- a/app/controllers/forms/check_your_answers_controller.rb
+++ b/app/controllers/forms/check_your_answers_controller.rb
@@ -77,7 +77,7 @@ module Forms
     end
 
     def back_link
-      if FeatureService.enabled?("filler_answer_email_enabled")
+      if FeatureService.enabled?("filler_answer_email_enabled") && current_context.form.copy_of_answers_enabled?
         copy_of_answers_path(form_id: current_context.form.id, form_slug: current_context.form.form_slug)
       else
         previous_step = current_context.previous_step(CheckYourAnswersStep::CHECK_YOUR_ANSWERS_STEP_SLUG)
--- a/app/controllers/forms/check_your_answers_controller.rb
+++ b/app/controllers/forms/check_your_answers_controller.rb
@@ -11,24 +11,24 @@ module Forms
       return redirect_to form_step_path(current_context.form.id, current_context.form.form_slug, current_context.next_step_slug, nil) unless current_context.can_visit?(CheckYourAnswersStep::CHECK_YOUR_ANSWERS_STEP_SLUG)
 
       setup_check_your_answers
-      skip_confirmation_email = skip_confirmation_email?
-      email_confirmation_input = skip_confirmation_email ? EmailConfirmationInput.new(send_confirmation: "skip_confirmation") : EmailConfirmationInput.new
+      will_send_copy_of_answers = will_send_copy_of_answers?
+      email_confirmation_input = will_send_copy_of_answers ? EmailConfirmationInput.new(send_confirmation: "skip_confirmation") : EmailConfirmationInput.new
 
       @support_details = current_context.support_details
 
-      render template: "forms/check_your_answers/show", locals: { email_confirmation_input:, skip_confirmation_email: }
+      render template: "forms/check_your_answers/show", locals: { email_confirmation_input:, will_send_copy_of_answers: }
     end
 
     def submit_answers
       @support_details = current_context.support_details
-      skip_confirmation_email = skip_confirmation_email?
-      email_confirmation_input = skip_confirmation_email ? EmailConfirmationInput.new(send_confirmation: "skip_confirmation") : EmailConfirmationInput.new(email_confirmation_input_params)
+      will_send_copy_of_answers = will_send_copy_of_answers?
+      email_confirmation_input = will_send_copy_of_answers ? EmailConfirmationInput.new(send_confirmation: "skip_confirmation") : EmailConfirmationInput.new(email_confirmation_input_params)
       requested_email_confirmation = email_confirmation_input.send_confirmation == "send_email"
 
       unless email_confirmation_input.valid?
         setup_check_your_answers
 
-        return render template: "forms/check_your_answers/show", locals: { email_confirmation_input:, skip_confirmation_email: }, status: :unprocessable_content
+        return render template: "forms/check_your_answers/show", locals: { email_confirmation_input:, will_send_copy_of_answers: }, status: :unprocessable_content
       end
 
       return redirect_to error_repeat_submission_path(@form.id) if current_context.form_submitted?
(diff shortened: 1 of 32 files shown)


spec/models/form_spec.rb
```diff
@@ -179,6 +179,24 @@
...
+      end
+    end
+
+    context "when send_copy_of_answers is \"disabled\"" do
+      let(:form_document) { build :v2_form_document, send_copy_of_answers: "disabled" }
+
+      it "returns false" do
+        expect(form.copy_of_answers_enabled?).to be false
+      end
+    end
```
stephencdaly, line 197: Can we add another test that this returns false if `send_copy_of_answers` is not present on the form document? We haven't backfilled this field on existing form documents so we need to be sure we're able to handle forms without it.

app/models/form.rb
```diff
@@ -64,6 +64,12 @@ def welsh?
     language == :cy
   end
 
+  def copy_of_answers_enabled?
+    return false unless form_document.respond_to?(:send_copy_of_answers)
+
+    form_document.send_copy_of_answers == "enabled"
+  end
```
stephencdaly, line 71: Not fully necessary, but it might be nice if the changes for `form.rb` were squashed into the same commit so we don't have 2 commits in the PR that make some changes and then change the same thing to work differently.

spec/requests/forms/step_controller_spec.rb
```diff
@@ -720,6 +721,25 @@
           post save_form_step_path(mode:, form_id: 2, form_slug: form_data.form_slug, step_slug: 2), params: { question: { text: "answer text" } }
           expect(response).to redirect_to(copy_of_answers_path(2, form_data.form_slug, mode:))
         end
+
+        context "when send_copy_of_answers is disabled on the form" do
```
stephencdaly, line 728: Can the `"Redirects to the copy of answers page"` above this be inside an example for "when send_copy_of_answers is enabled on the form" so it's clearer under which conditions we'd expect this scenario?

spec/requests/forms/step_controller_spec.rb
```diff
@@ -720,6 +721,25 @@
...
+          let(:form_data) do
+            build(:v2_form_document, :with_support,
+                  form_id: 2,
+                  start_page: first_step_id,
+                  privacy_policy_url: "http://www.example.gov.uk/privacy_policy",
+                  what_happens_next_markdown: "Good things come to those that wait",
+                  declaration_text: "agree to the declaration",
+                  available_languages:,
+                  send_copy_of_answers: "disabled",
```
stephencdaly: Can we just define a `let` for `send_copy_of_answers` at the top of the file so we don't need to redefine the whole form?

spec/requests/forms/check_your_answers_controller_spec.rb
```diff
@@ -194,6 +195,32 @@
...
+                start_page: 1,
+                privacy_policy_url: "http://www.example.gov.uk/privacy_policy",
+                what_happens_next_markdown: "Good things come to those that wait",
+                declaration_text: "agree to the declaration",
+                steps: steps_data,
+                support_phone: "0203 222 2222",
+                support_email: "help@example.gov.uk",
+                support_url: "https://example.gov.uk/help",
+                support_url_text: "Get help",
+                send_copy_of_answers: "disabled",
```
stephencdaly: Can we define a `let` for `send_copy_of_answers` so we don't need to redefine the whole form?

stephencdaly: As the default is for send_copy_of_answers to be disabled for a form, would it be better to invert this context so that the other tests check the back link goes back to the last step, and this context checks the scenario where `send_copy_of_answers` is enabled?

config/locales/cy.yml
```diff
@@ -362,12 +362,17 @@ cy:
       title: Mae’n ddrwg gennym, nid yw’r gwasanaeth hwn ar gael mwyach
     check_your_answers:
       agree_and_submit: Cytuno a chyflwyno
+      change: Newid
+      copy_of_answers: Cael copi o'ch atebion
       declaration: Datganiad
+      'no': Na
       not_completed: Heb ei gwblhau
+      send_copy: 'Byddwn yn anfon e-bost cadarnhad gyda chopi o''ch atebion i''ch cyfeiriad e-bost GOV.UK One Login: %{email_address}'
```
stephencdaly:
```suggestion
      send_copy: 'Byddwn yn anfon e-bost cadarnhad gyda chopi o’ch atebion i’ch cyfeiriad e-bost GOV.UK One Login: %{email_address}'
```

config/locales/cy.yml
```diff
@@ -362,12 +362,17 @@ cy:
       title: Mae’n ddrwg gennym, nid yw’r gwasanaeth hwn ar gael mwyach
     check_your_answers:
       agree_and_submit: Cytuno a chyflwyno
+      change: Newid
+      copy_of_answers: Cael copi o'ch atebion
```
stephencdaly:
```suggestion
      copy_of_answers: Cael copi o’ch atebion
```

config/locales/cy.yml
```diff
@@ -362,12 +362,17 @@ cy:
...
       declaration: Datganiad
+      'no': Na
       not_completed: Heb ei gwblhau
+      send_copy: 'Byddwn yn anfon e-bost cadarnhad gyda chopi o''ch atebion i''ch cyfeiriad e-bost GOV.UK One Login: %{email_address}'
       submit: Cyflwyno
       title: Edrychwch dros eich atebion cyn cyflwyno’ch ffurflen
     submitted:
       continue_to_pay: Parhau i dalu
+      copy_of_answers_sent: Rydym wedi anfon e-bost cadarnhau atoch gyda chopi o'ch atebion.
```
stephencdaly:
```suggestion
      copy_of_answers_sent: Rydym wedi anfon e-bost cadarnhau atoch gyda chopi o’ch atebion.
```

config/locales/cy.yml
```diff
@@ -395,14 +400,14 @@ cy:
     back: Yn ôl
     continue_to_one_login:
       show:
-        body: To help make sure the information we send you stays secure, you’ll need to sign in to or create a GOV.UK One Login.
-        continue_button: Continue to GOV.UK One Login
-        title: Use GOV.UK One Login to keep your information secure
+        body_html: "<p>Bydd yr e-bost gyda'ch atebion yn cael ei anfon i'r cyfeiriad e-bost sy'n gysylltiedig â'ch GOV.UK One Login.</p><p>Os ydych angen i'r atebion gael eu hanfon i gyfeiriad e-bost gwahanol - er enghraifft, eich e-bost gwaith - gallwch greu GOV.UK One Login newydd.</p>"
```
stephencdaly:
```suggestion
        body_html: "<p>Bydd yr e-bost gyda’ch atebion yn cael ei anfon i’r cyfeiriad e-bost sy’n gysylltiedig â’ch GOV.UK One Login.</p><p>Os ydych angen i’r atebion gael eu hanfon i gyfeiriad e-bost gwahanol - er enghraifft, eich e-bost gwaith - gallwch greu GOV.UK One Login newydd.</p>"
```

config/locales/cy.yml
```diff
@@ -395,14 +400,14 @@ cy:
...
+        continue_button: Parhau i GOV.UK One Login
+        title: Defnyddiwch GOV.UK One Login i gadw'ch gwybodaeth yn ddiogel
```
stephencdaly:
```suggestion
        title: Defnyddiwch GOV.UK One Login i gadw’ch gwybodaeth yn ddiogel
```

config/locales/cy.yml
```diff
@@ -395,14 +400,14 @@ cy:
...
     copy_of_answers:
       show:
-        heading: PLACEHOLDER! Ydych chi eisiau copi o'ch atebion?
-        hint: PLACEHOLDER! Byddwn ond yn defnyddio'r cyfeiriad e-bost rydych yn ei ddarparu yma i anfon cadarnhad bod eich ffurflen wedi'i chyflwyno'n llwyddiannus. Ni fydd yn cynnwys copi o'ch atebion.
-        title: PLACEHOLDER! Ydych chi eisiau copi o'ch atebion?
+        heading: Ydych chi eisiau cael e-bost gyda chopi o'ch atebion?
```
stephencdaly:
```suggestion
        heading: Ydych chi eisiau cael e-bost gyda chopi o’ch atebion?
```

config/locales/cy.yml
```diff
@@ -395,14 +400,14 @@ cy:
...
+        hint: Byddwch angen GOV.UK One Login ar gyfer hyn - byddwch yn gallu mewngofnodi neu greu un newydd.
+        title: Ydych chi eisiau cael e-bost gyda chopi o'ch atebion?
```
stephencdaly:
```suggestion
        title: Ydych chi eisiau cael e-bost gyda chopi o’ch atebion?
```

app/views/forms/check_your_answers/show.html.erb
```diff
@@ -19,29 +19,57 @@
 
   <%= render CheckYourAnswersComponent::View.new(form: @form, steps: @steps, mode: @mode) %>
 
-  <div class="govuk-grid-row">
-    <div class="govuk-grid-column-two-thirds-from-desktop">
-      <%= form.govuk_radio_buttons_fieldset(:send_confirmation, legend: { size: 'm', tag: 'h2' }) do %>
-        <%= form.govuk_radio_button :send_confirmation, 'send_email', link_errors: :true do %>
-          <%= form.govuk_email_field :confirmation_email_address, autocomplete: 'email', spellcheck: false  %>
+  <% unless skip_confirmation_email %>
```
stephencdaly: This section is currently showing if the form has send_copy_of_answers disabled.

spec/views/forms/check_your_answers/show.html.erb_spec.rb
```diff
@@ -19,7 +19,7 @@
     assign(:steps, steps)
     assign(:form, form)
     assign(:support_details, support_details)
-    render template: "forms/check_your_answers/show", locals: { email_confirmation_input: }
+    render template: "forms/check_your_answers/show", locals: { email_confirmation_input:, skip_confirmation_email: false }
```
stephencdaly: Can we add tests for the optional content when skip_confirmation_email is true/false in this view spec?

spec/requests/forms/check_your_answers_controller_spec.rb
```diff
@@ -196,6 +196,49 @@
...
+        before do
+          get check_your_answers_path(mode:, form_id:, form_slug: form_data.form_slug)
+        end
+
+        it "hides the confirmation email question" do
+          expect(response.body).not_to include("email_confirmation_input[send_confirmation]")
+        end
+      end
+
+      context "when the user has said yes to copy of answers but has no One Login email" do
```
stephencdaly, line 223: Do we have a test covering if they've said "no" to a copy of your answers?

app/views/forms/submitted/submitted.html.erb
```diff
@@ -3,6 +3,9 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <%= render partial: "forms/submitted/submitted_panel", locals: {current_context: @current_context, } %>
+    <% if @current_context.wants_copy_of_answers? && @current_context.get_copy_of_answers_email_address.present? %>
```
stephencdaly: We're checking `@current_context.wants_copy_of_answers? && @current_context.get_copy_of_answers_email_address.present? ` in a few places now.

We're also checking it in `form_submission_service.rb` if you rebase. Reckon it would be worth adding a method in `confirmation_details_store.rb` for this so we don't need tests checking both that they want a copy and the email exists in multiple places?

app/controllers/forms/check_your_answers_controller.rb
```diff
@@ -52,7 +54,7 @@ def submit_answers
       rescue FormSubmissionService::ConfirmationEmailToAddressError
         setup_check_your_answers
         email_confirmation_input.errors.add(:confirmation_email_address, :invalid_email)
-        render template: "forms/check_your_answers/show", locals: { email_confirmation_input: }, status: :unprocessable_content
+        render template: "forms/check_your_answers/show", locals: { email_confirmation_input:, skip_confirmation_email: }, status: :unprocessable_content
```
stephencdaly: The name of the `skip_confirmation_email` is slightly misleading as we will be sending a confirmation email if this is true. Could we rename it to something like `will_send_copy_of_answers` to make it clearer under what conditions this is true?

stephencdaly (changes requested):
I've left a bunch of style comments, it's up to you if you want to make the suggested changes for these.

The things that we should change are:

- We need a test for when the `send_copy_of_answers` isn't present on the form document to prevent regressions.
- We shouldn't show the extra section on the check your answers page if `send_copy_of_answers` is disabled for the form.
- The Welsh translations should have curly quotes

app/views/forms/check_your_answers/show.html.erb
```diff
@@ -19,29 +19,57 @@
...
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds-from-desktop">
+        <h2 class="govuk-heading-m">You've chosen to not get a copy of your answers</h2>
```
stephencdaly: This should use a translation

spec/models/form_spec.rb
```diff
@@ -179,6 +179,32 @@
...
+
+    context "when send_copy_of_answers is not present on the form document" do
+      let(:form_document) { build :v2_form_document, send_copy_of_answers: nil }
```
stephencdaly, line 200: We should also have a test for when the `send_copy_of_answers` field isn't present at all, as the behaviour is not the same as if it's set to nil. See the other tests in this file which use an OpenStruct to simulate this.

stephencdaly (commented):
Changes now work as expected. We just need to add a test to form_spec.rb for when the field is missing from the form document.

The Welsh translations haven't been updated to use curly quotes - you should just be able to apply my suggestions using the "Add suggestion batch" button on the comments to put them all in one commit.

stephencdaly (approved):
All good now, thanks for making those changes!

PR #2122, opened 2026-05-26
Add more logging to SendBounceNotificationsJob

lfdebrux:

I was looking at the runs of this job this morning and found myself a bit confused trying to work out how many emails were sent if any, this PR adds a bit of extra logging that I would have found useful.

See commit messages for details.


stephencdaly (approved):
Nice, this will be useful

PR #2123, opened 2026-05-26
Refactoring to prepare for sending confirmation emails using SES

stephencdaly:

Trello card: https://trello.com/c/hCWhyMOm

Refactoring to prepare for also sending confirmation emails containing a copy of the answers using SES. See individual commit messages for details.

### DIFF
--- a/app/helpers/email_format_helper.rb
+++ b/app/helpers/email_format_helper.rb
@@ -1,0 +1,14 @@
+module EmailFormatHelper
+  def normalize_whitespace(text)
+    text.strip.gsub(/\r\n?/, "\n").split(/\n\n+/).map(&:strip).join("\n\n")
+  end
+
+  def convert_newlines_to_html(text)
+    text.gsub("\n", "<br/>")
+  end
+
+  def normalize_whitespace_and_convert_to_html(text)
+    output = normalize_whitespace(text)
+    convert_newlines_to_html(output)
+  end
+end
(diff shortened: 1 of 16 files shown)


app/helpers/email_format_helper.rb
```diff
@@ -0,0 +1,22 @@
+module EmailFormatHelper
+  def normalize_whitespace(text)
+    text.strip.gsub(/\r\n?/, "\n").split(/\n\n+/).map(&:strip).join("\n\n")
+  end
+
+  def convert_newlines_to_html(text)
+    text.gsub("\n", "<br/>")
+  end
+
+  def normalize_whitespace_and_convert_to_html(text)
```
lfdebrux, line 10: Is this used anywhere?

lfdebrux (approved):

app/mailers/application_mailer.rb
```diff
@@ -1,3 +1,8 @@
 class ApplicationMailer < ActionMailer::Base
+  helper :email_format
```
lfdebrux, line 2: The documentation for `helper` recommends using the the module constant instead of a symbol [[1](https://api.rubyonrails.org/classes/AbstractController/Helpers/ClassMethods.html#method-i-helper)], is there any reason not to do that here?

```suggestion
  helper EmailFormatHelper
```

app/helpers/email_format_helper.rb
stephencdaly: Ah it will be - I just extracted out these commits so I didn't have a massive PR

app/mailers/application_mailer.rb
stephencdaly: I don't see why not, this is just what I found from Google. I'll update in my next PR

PR #2124, opened 2026-05-27
Send confirmation emails with a copy of answers using SES

stephencdaly:

Trello card: https://trello.com/c/hCWhyMOm

Send a confirmation email using Amazon SES containing a copy of the user's answers if they have signed in with One Login and asked for a copy of their answers.

If they have not asked for a copy of their answers, continue sending the confirmation email using Notify. However, the mailer for sending with SES accepts an option for including a copy of the answers or not, so switching over down the line to use SES for all confirmation emails should be easy.

### Email styling

We're not currently explicitly specifying the style for headings, as per the submission email. We probably should, especially as we're now introducing h4s. I'd welcome comments on the best way to do this - but might raise a follow up PR to look at this as this PR is already very big.

### Testing

You'll need to configure your GOV.UK One Login test service (ask me how if you haven't)

Then run locally, in an AWS shell for the dev-readonly role with the following environment variables:
```
gds aws forms-dev-readonly --shell

SETTINGS__GOVUK_ONE_LOGIN__CLIENT_ID=your-client-id SETTINGS__GOVUK_ONE_LOGIN__PRIVATE_KEY=your-base64-encoded-private-key ASSUME_DEV_IAM_ROLE=true bundle exec rails s
```

Screenshot 2026-05-27 at 12 08 13

Screenshot 2026-05-27 at 12 08 25

### DIFF
--- a/app/mailers/aws_ses_submission_confirmation_mailer.rb
+++ b/app/mailers/aws_ses_submission_confirmation_mailer.rb
@@ -1,0 +1,29 @@
+class AwsSesSubmissionConfirmationMailer < ApplicationMailer
+  def submission_confirmation_email(
+    submission:,
+    confirmation_email_address:,
+    include_copy_of_answers:
+  )
+    @submission_locale = submission.submission_locale.to_sym
+    @form = submission.form
+    @welsh_form = submission.welsh_form
+    @submission = submission
+    @include_copy_of_answers = include_copy_of_answers
+
+    mail(to: confirmation_email_address, subject: subject)
+  end
+
+  def subject
+    subject = if @submission.preview?
+                I18n.t("mailer.submission_confirmation.subject_preview", reference: @submission.reference)
+              else
+                I18n.t("mailer.submission_confirmation.subject", reference: @submission.reference)
+              end
+
+    if @submission_locale == :cy
+      subject = "#{subject} | #{I18n.t('mailer.submission_confirmation.subject', reference: @submission.reference, locale: :cy)}"
+    end
+
+    subject
+  end
+end
--- a/app/models/repeatable_step.rb
+++ b/app/models/repeatable_step.rb
@@ -77,9 +77,9 @@ class RepeatableStep < Step
     end
   end
 
-  def show_answer_in_email(submission_reference:)
+  def show_answer_in_email(submission_reference:, confirmation_email: false)
     questions.map.with_index(1) { |question, index|
-      "#{index}. #{question.show_answer_in_email(submission_reference:)}"
+      "#{index}. #{question.show_answer_in_email(submission_reference:, confirmation_email:)}"
     }.join("\n\n")
   end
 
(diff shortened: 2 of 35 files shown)


app/mailers/aws_ses_submission_confirmation_mailer.rb
```diff
@@ -0,0 +1,29 @@
+class AwsSesSubmissionConfirmationMailer < ApplicationMailer
```
theseanything, line 1: Nit: do we need to say "AwsSes" in the class name - there's nothing in this class implementation that's specific to AWS SES

theseanything: I know we do so with other mailer names - but really is irrelevant.

app/views/aws_ses_submission_confirmation_mailer/_content.html.erb
theseanything, line 1: Nit: don't think you need to call this out as "localised" content

app/models/submission.rb
```diff
@@ -79,8 +79,18 @@ def form_document_resource
...
-    SesEmailFormatter.new(submission_reference: reference, steps: journey.completed_steps)
+  def ses_email_formatter(locale)
+    SesEmailFormatter.new(submission_reference: reference, steps: journey_for_locale(locale).completed_steps)
+  end
+
+  def journey_for_locale(locale)
+    return welsh_journey if locale.to_sym == :cy
+
+    journey
+  end
```
theseanything: the original journey method should take in a argument locale - rather than a whole separate method

app/models/repeatable_step.rb
```diff
@@ -77,9 +77,9 @@ def show_answer
     end
   end
 
-  def show_answer_in_email(submission_reference:)
+  def show_answer_in_email(*)
```
theseanything: Why are we using a splat instead of explicit arguments?

app/mailers/aws_ses_submission_confirmation_mailer.rb
stephencdaly: I've created a card for using SES for all confirmation emails and added a task to remove the prefixes when we're no longer sending using Notify too https://trello.com/c/GfBTJbNL/3581-use-ses-to-send-all-confirmation-emails

app/views/aws_ses_submission_confirmation_mailer/_content.html.erb
stephencdaly: have renamed to just `_content`

app/models/submission.rb
stephencdaly: Good point, have updated

app/models/repeatable_step.rb
stephencdaly: Yeh we shouldn't, have updated.

theseanything (approved):

PR #2125, opened 2026-05-27
Send bounce notifications to org admins if there are no group admins

stephencdaly:

Trello card: https://trello.com/c/8tbtHbMj

It is possible for a group to have no group admins. To ensure that we notify someone when submissions are bounced, send the email to org admins for the initial notification the day after the bounce if there are no group admins.

In this case, we will still send the follow-up notification 7 days later if the bounce hasn't been resolved.

Update the content we include in the notifications to org admins so it makes sense either when we are contacting them about a bounce that just happened, or sending the second notification 7 days later and regardless of whether the group has admins or not.

Screenshot 2026-05-28 at 16 28 04

### DIFF
--- a/app/jobs/schedule_bounce_notifications_job.rb
+++ b/app/jobs/schedule_bounce_notifications_job.rb
@@ -6,7 +6,7 @@ class ScheduleBounceNotificationsJob < ApplicationJob
     CurrentJobLoggingAttributes.job_class = self.class.name
     CurrentJobLoggingAttributes.job_id = job_id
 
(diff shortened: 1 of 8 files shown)


stephencdaly:
@william-castoncook can you double check the content change please?

theseanything (approved):

DavidBiddle (approved):

william-castoncook:
> @william-castoncook can you double check the content change please?

Looks good to me!

PR #2128, opened 2026-05-28
Log delivered_at and delivery_latency for immediate email deliveries

theseanything:
Slow immediate email deliveries are hard to investigate when latency only lives in CloudWatch. This adds delivered_at and delivery_latency to job logs so we can spot patterns in log queries—e.g. whether delays cluster on a particular form—without cross-referencing metrics for every case.

Batch deliveries are unchanged. delivery_latency is milliseconds from submission creation to SES delivery confirmation.

### DIFF
--- a/app/jobs/receive_submission_deliveries_job.rb
+++ b/app/jobs/receive_submission_deliveries_job.rb
@@ -22,24 +22,25 @@ class ReceiveSubmissionDeliveriesJob < ApplicationJob
 
       delivered_at = Time.zone.parse(ses_message["delivery"]["timestamp"])
       process_delivery(delivery, submission, delivered_at:)
-
-      if delivery.immediate?
-        submission_duration_ms = ((delivered_at - submission.created_at) * 1000).round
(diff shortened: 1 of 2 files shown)


stephencdaly (dismissed):
This seems fine. I was wondering whether it makes sense to pass `delivery_duration_ms` in a hash as the second argument to `EventLogger.log_form_event`, as we only need to log this one place. 

The intention of CurrentJobLoggingAttributes is more for setting attributes that will be set on most/many log lines. I don't think it's a big deal either way though.

theseanything:
> This seems fine. I was wondering whether it makes sense to pass `delivery_duration_ms` in a hash as the second argument to `EventLogger.log_form_event`, as we only need to log this one place.
> 
> The intention of CurrentJobLoggingAttributes is more for setting attributes that will be set on most/many log lines. I don't think it's a big deal either way though.

Ahhhh yeah that makes more sense - just updated it do that instead.

stephencdaly (approved):

PR #2130, opened 2026-06-01
update welsh translations for copy of answers flow

alice-carr:

this commit updates the welsh translations for parts of the copy your answers flow. it also updates the auth_error page to show the error page in the language the user is viewing the form in.

Trello card: https://trello.com/c/6niCHGis/2910-update-runner-content-for-copy-of-answers

### DIFF
--- a/app/controllers/users/omniauth_controller.rb
+++ b/app/controllers/users/omniauth_controller.rb
@@ -7,7 +7,10 @@ module Users
       Sentry.capture_exception(exception)
 
       link_url = copy_of_answers_path(**auth_service.form_path_params)
-      render "errors/auth_error", locals: { link_url: }, status: :bad_request
+      locale = auth_service.form_path_params[:locale] || I18n.default_locale
+      I18n.with_locale(locale) do
+        render "errors/auth_error", locals: { link_url: }, status: :bad_request
+      end
     rescue Store::ReturnFromOneLoginStore::MissingReturnParamsError
       render "errors/return_from_one_login_session_missing", status: :bad_request
     end
--- a/config/locales/cy.yml
+++ b/config/locales/cy.yml
@@ -86,7 +86,7 @@ cy:
         copy_of_answers_input:
           attributes:
             copy_of_answers:
-              blank: PLACEHOLDER! Dewiswch ydw os hoffech dderbyn copi o'ch atebion
+              blank: Dewiswch ‘Ydw’ os ydych eisiau copi o’ch atebion
         email_confirmation_input:
           attributes:
             confirmation_email_address:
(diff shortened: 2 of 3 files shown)


app/controllers/users/omniauth_controller.rb
```diff
@@ -7,7 +7,9 @@ class FailureError < StandardError; end
       Sentry.capture_exception(exception)
 
       link_url = copy_of_answers_path(**auth_service.form_path_params)
-      render "errors/auth_error", locals: { link_url: }, status: :bad_request
+      I18n.with_locale(auth_service.form_path_params[:locale]) do
```
stephencdaly: `auth_service.form_path_params[:locale]` can return nil. While this seems to work, we should probably be a bit more deliberate about it and default to using `I18n.default_locale` if it's nil.

We should probably also add a test for when the locale is set to Welsh that it renders the error in Welsh

config/locales/cy.yml
```diff
@@ -86,7 +86,7 @@ cy:
         copy_of_answers_input:
           attributes:
             copy_of_answers:
-              blank: PLACEHOLDER! Dewiswch ydw os hoffech dderbyn copi o'ch atebion
+              blank: Dewiswch 'Ydw' os ydych eisiau copi o’ch atebion
```
stephencdaly:
```suggestion
              blank: Dewiswch ‘Ydw’ os ydych eisiau copi o’ch atebion
```

config/locales/cy.yml
```diff
@@ -337,9 +337,9 @@ cy:
       title: Er lles eich diogelwch, rydyn ni wedi dileu’ch atebion
     return_from_one_login_session_missing:
       body_html: |
-        <p>We could not confirm your email address using GOV.UK One Login.</p>
-        <p>You’ll need to go back to the form and submit your answers - try clicking the ‘Back’ button in your browser.</p>
-      title: Sorry, something has gone wrong
+        <p>Nid oeddem yn gallu cadarnhau eich cyfeiriad e-bost gan ddefnyddio GOV.UK One Login.</p>
+        <p>Bydd angen i chi fynd yn ôl i’r ffurflen a chyflwyno’ch atebion - ceisiwch glicio ar y botwm ’Yn ôl’ yn eich porwr.</p>
```
stephencdaly:
```suggestion
        <p>Bydd angen i chi fynd yn ôl i’r ffurflen a chyflwyno’ch atebion - ceisiwch glicio ar y botwm ‘Yn ôl’ yn eich porwr.</p>
```

app/controllers/users/omniauth_controller.rb
alice-carr: good shout, i've updated this now

spec/requests/users/omniauth_controller_spec.rb
```diff
@@ -65,6 +65,36 @@
       end
     end
 
+    context "when data is missing on the auth details on the request and the form is being used in welsh" do
```
stephencdaly: For organisation of the tests, it might be nice to nest these new contexts inside the "when data is missing on the auth details on the request" context so they're nicely grouped.

stephencdaly (dismissed):

stephencdaly (approved):

spec/requests/users/omniauth_controller_spec.rb
alice-carr: thanks Stephen!

PR #2133, opened 2026-06-03
Fix empty string filename_suffix triggering answered?

SamJamCul:
This change arose from a bug that occurred when a file upload question sits behind a route. If the routing question's answer value is changed form the check your answers page so that the file upload question is unlocked, the user would be brought straight back to the check your answers page, rather than to the file upload question.

This happened because when a file upload question was initially skipped, the default value for the `filename_suffix` is set as an empty string, "". This meant that when the `answered?` method is used to check whether the file upload question needed to be answered, it was always returning true, as the empty string in "filename_suffix" would return true when `nil?` was called on it.

This change avoids checking the filename_suffix entirely when considering whether a question is `answered?`, as it's not a user generated value, so shouldn't have any bearing on this calculation.

https://trello.com/c/jAkCdQEj/3133-file-upload-question-treated-as-skipped-when-mandatory

### DIFF
--- a/app/models/question/file.rb
+++ b/app/models/question/file.rb
@@ -117,6 +117,10 @@ module Question
       [caption, question_text_with_optional_suffix].join(" ")
     end
 
+    def answered?
+      @attributes.to_hash.reject { |key, value| key == "filename_suffix" && value == "" }.any? { |_key, value| !value.nil? }
+    end
+
   private
 
     def filename_for_submission(submission_reference:, is_s3_submission:)
--- a/spec/mailers/aws_ses_submission_confirmation_mailer_spec.rb
+++ b/spec/mailers/aws_ses_submission_confirmation_mailer_spec.rb
@@ -31,7 +31,7 @@ RSpec.describe AwsSesSubmissionConfirmationMailer, type: :mailer do
     {
       "q1" => { text: "blue" },
       "q2" => { first_name: "Jane", last_name: "Doe" },
-      "q3" => "",
+      "q3" => { original_filename: "" },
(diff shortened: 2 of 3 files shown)


app/models/question/question_base.rb
```diff
@@ -30,7 +30,7 @@ def attributes
     end
 
     def answered?
-      @attributes.to_hash.any? { |_key, value| !value.nil? }
+      @attributes.to_hash.reject { |key, value| key == "filename_suffix" && value == "" }.any? { |_key, value| !value.nil? }
     end
```
lfdebrux, line 34: To me this feels like a good place to rely on object oriented design; rather than changing the base class (and leaking the details of the child classes), we should be able to add an `#answered?` method to the `File` class (which inherits from `QuestionBase`) to override the behaviour. For me that feels like a cleaner and more maintainable approach. What do you think?

lfdebrux (commented):
Thanks for fixing this, and thanks for the detailed description/commit message describing the problem and the fix.

I think there might be a better place to change the code though, I've left a comment.

app/models/question/question_base.rb
SamJamCul: Ah I completely forgot about that, good shout!

SamJamCul: It now sits in the `file.rb`, er, file.

lfdebrux (dismissed):
Changes look good to me, and I've tested locally and confirmed that this does fix the issue.

lfdebrux (approved):

PR #2146, opened 2026-06-04
Fix 'None of the above' for selection questions in Welsh forms

stephencdaly:

The "None of the above" option for selection questions was using a localised translation for the value, which meant that if you switched language after selecting "None of the above", we wouldn't correctly identify that "None of the above" was selected, hiding the none of the above answer if one was present from the check your answers page and submission/confirmation emails.

It also meant we would include the localised translation of "None of the above" at time the question was answered in the submission email, when for other options we would always show the English value.

Fix this so the `value` is a constant regardless of the language, as for other selection options. Whenever we display the answer, use the `name` of the option, which will be localised as appropriate.

### DIFF
--- a/app/components/question/selection_component/view.rb
+++ b/app/components/question/selection_component/view.rb
@@ -37,7 +37,7 @@ module Question
         return nil unless question.is_optional?
 
         option = form_builder.govuk_radio_button :selection,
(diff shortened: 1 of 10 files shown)


lfdebrux (approved):
Changes look good to me and they work on my machine

PR #2147, opened 2026-06-04
Record submission counts

theseanything:
Records a SubmissionCount metric using an OpenTelemetry counter, incremented as each submission is created and exported through the OTel collector (metrics support enabled in https://github.com/govuk-forms/forms-deploy/pull/2182).

This allows as to track platform usage over time without manual reporting and is one of our key KPIs used to drive product development. We get stats exported every minute.

We are using Otel instead of standard CloudWatch Metrics as it's significantly cheaper. Allows us to have high cardinality and get granular submission counts per form.

Submissions to automated test forms (capybara, smoke test and s3 test forms) are labelled with mode test, so dashboards can exclude them without repeating name-based filters in every query.

Trello card: https://trello.com/c/b3cNvzu5/3136-record-kpis-in-cloudwatch

tested in dev.

### DIFF
--- a/app/services/metrics.rb
+++ b/app/services/metrics.rb
@@ -10,7 +10,6 @@ module Metrics
       submission_counter.add(
         1,
         attributes: {
-          "deployment.environment.name" => Settings.forms_env.downcase,
           "form.id" => form_id.to_s,
           "form.submission.mode" => mode_label(form_name:, mode:),
         },
--- a/app/services/metrics.rb
+++ b/app/services/metrics.rb
@@ -2,14 +2,17 @@ module Metrics
   METER_NAME = "forms-runner".freeze
   METER_VERSION = "1.0".freeze
 
+  # Forms with these names are used for automated testing of the platform
+  TEST_FORM_NAME_PATTERNS = [/\Acapybara/, /smoke/, /\As3/].freeze
+
   class << self
-    def record_submission(form_id:, mode:)
+    def record_submission(form_id:, form_name:, mode:)
       submission_counter.add(
         1,
         attributes: {
           "Environment" => Settings.forms_env.downcase,
           "FormId" => form_id.to_s,
-          "Mode" => mode.to_s,
+          "Mode" => mode_label(form_name:, mode:),
         },
       )
     rescue StandardError => e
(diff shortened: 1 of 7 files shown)


app/services/metrics/submission_count_service.rb
```diff
@@ -0,0 +1,63 @@
...
+    rescue Aws::CloudWatch::Errors::ServiceError,
+           Aws::Errors::MissingCredentialsError => e
+      Sentry.capture_exception(e)
+      raise
+    end
+
+  private
+
+    def submission_counts_by_form_id
+      Submission.where(mode: "form").group(:form_id).count
```
lfdebrux: Currently I think the submission counts collected by Anne exclude submissions to test forms by our end to end tests and smoke tests. Ought we do the same here?

theseanything: That's okay - submissions are grouped by form id, so we can filter them out after.

thomasiles: The end-to-end tests create a form, submit it and then delete it. I don't know if we'll easily have the form ids to exclude and including known form ids would be tricky too, i think?

lfdebrux: Mmm I agree with what Tom says, it feels like it will be easier to filter out before sending to CloudWatch, we can use the same criteria Anne uses currently [in Splunk](https://gds.splunkcloud.com/en-GB/app/gds-543-forms/report?s=%2FservicesNS%2Fnobody%2Fgds-543-forms%2Fsaved%2Fsearches%2FSubmissions%2520and%2520completions):

```
form_name!=capybara* form_name!=*smoke* form_name!=s3*
```

theseanything: I've updated this PR to now use OTel, and subsequently label these forms as mode "test" - so they can easily be filtered out.

app/services/metrics.rb
```diff
@@ -0,0 +1,45 @@
+module Metrics
+  METER_NAME = "forms-runner".freeze
+  METER_VERSION = "1.0".freeze
+
+  # Forms with these names are used for automated testing of the platform
+  TEST_FORM_NAME_PATTERNS = [/\Acapybara/, /smoke/, /\As3/].freeze
```
DavidBiddle, line 6: Not a problem for this PR, and it seems to be OK for our current forms in production, but I wonder if we should make these more specific at some point. 'smoke' at least seems like a word that could plausibly appear in a real form name eventually

DavidBiddle (approved):

app/services/metrics.rb
theseanything: Yeah agree! It'd be good to have an explicit tag on a for to use

PR #2148, opened 2026-06-05
Fix errors when a date question tries to load answer for repeatable step from store

lfdebrux:

Fixes an error we're seeing in production, see https://govuk-forms.sentry.io/issues/7446683762

If a form creator is half-way through filling in a preview of a draft form, and changes a date question in that form from "more than one answer" to "only one answer", this can cause a server error.

### DIFF
--- a/app/lib/flow/journey.rb
+++ b/app/lib/flow/journey.rb
@@ -120,7 +120,7 @@ module Flow
 
       begin
         step.load_from_store(@answer_store)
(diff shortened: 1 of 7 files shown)


stephencdaly (commented):
Change seems sensible and works for me locally. Looks like there's just a test that needs updating

lfdebrux:
> Change seems sensible and works for me locally. Looks like there's just a test that needs updating

Thanks. The tests should be passing now.

stephencdaly (approved):

PR #2164, opened 2026-06-16
Use SES to send all confirmation emails

alice-carr:

Trello card: https://trello.com/c/GfBTJbNL/3154-blocked-until-a-sprint-after-copy-of-answers-launch-use-ses-to-send-all-confirmation-emails

We currently use SES to send confirmation emails that include a copy of the user’s answers and Notify to send confirmation emails that do not. We should switch all confirmation emails to use SES.

This must be merged alongside the e2e pr: https://github.com/govuk-forms/forms-e2e-tests/pull/324


app/jobs/send_confirmation_email_job.rb
```diff
@@ -1,24 +1,18 @@
 class SendConfirmationEmailJob < ApplicationJob
   queue_as :confirmation_emails
 
-  def perform(submission:, notify_response_id:, confirmation_email_address:, include_copy_of_answers: false)
+  def perform(submission:, confirmation_email_address:, include_copy_of_answers: false)
```
theseanything: I don't think we can immediately drop notify_response_id: keyword as it breaks any job already serialised in solid_queue at deploy time. Old payloads raise ArgumentError: unknown keyword and can't be retried those confirmation emails are lost. 

Think you just need to do "notify_response_id: nil" for one release.

theseanything: Sorry had a quick peek, when it was up for review!

alice-carr: Thanks Sean! I've added it in as nil for now with a note to remove it once it's been deployed

theseanything (approved):

PR #2181, opened 2026-06-25
Add JWKS endpoint for one login

stephencdaly:

Trello card: https://trello.com/c/qwLQt9Ur

Add a publicly accessible endpoint that returns a JSON Web Key Set (JWKS) to share our public key with GOV.UK One Login.

I've tested that this works on the dev environment.


DavidBiddle (approved):
Code looks good and as far as I can tell it works locally with the OL integration environment 🎉

PR #2182, opened 2026-06-25
Render SES emails using markdown

DavidBiddle:

Trello card: https://trello.com/c/d5t1W2Hs/3151-render-ses-emails-using-markdown

Converts our existing SES email templates into markdown. This means that we no longer have to manually maintain separate HTML and plaintext templates for these emails - instead we can write a single markdown template and render it as HTML and plaintext respectively. It also reduces the number of inline styles we have to maintain.

#### Submission email
Screen Shot 2026-06-29 at 11 19 00

#### Daily batch email
Screen Shot 2026-06-29 at 11 19 35

#### Weekly batch email
Screen Shot 2026-06-29 at 11 19 53

#### Confirmation email with copy of answers
Screen Shot 2026-07-01 at 12 54 16


app/lib/ses_email_formatter.rb
```diff
@@ -57,6 +85,14 @@ def prep_html_none_of_the_above_answer_text(step)
     ]
   end
 
+  def prep_markdown_none_of_the_above_answer_text(step)
+    [
+      prep_answer_text(step),
+      "#### #{prep_none_of_the_above_question_text(step)}",
```
stephencdaly: Not directly related to this change but I think we missed this before when adding the answers to confirmation emails. Should this be one heading level deeper in confirmation emails? So h5 instead of h4?

This can be addressed separately to this PR if we think this needs to be changed.

app/helpers/email_format_helper.rb
```diff
@@ -9,11 +9,20 @@ def convert_newlines_to_html(text)
     text.gsub("\n", "<br/>")
   end
 
+  def convert_single_newlines_to_markdown_linebreaks(text)
+    text.gsub(/(?<!\n)\n(?!\n)/, "  \n")
+  end
+
   def normalize_whitespace_and_convert_to_html(text)
```
stephencdaly: This and a few other methods in this file can be removed now

app/helpers/email_format_helper.rb
```diff
@@ -9,11 +9,20 @@ def convert_newlines_to_html(text)
     text.gsub("\n", "<br/>")
   end
 
+  def convert_single_newlines_to_markdown_linebreaks(text)
+    text.gsub(/(?<!\n)\n(?!\n)/, "  \n")
```
stephencdaly, line 9: Something about this isn't working quite right in the plain text markdown renderer and it's rendering `  \n` as `<br>\n`. See the answer content in the mailer preview for an example of this http://localhost:3001/rails/mailers/aws_ses_form_submission_mailer/preview_submission_email.txt?locale=en

app/views/aws_ses_submission_batch_mailer/_daily_submission_batch_content.md.erb
```diff
@@ -0,0 +1,36 @@
...
+<%= I18n.t("mailer.submission_batch.daily.multiple_files_explainer", date: @date) %>
+
+<% else %>
+<%= I18n.t("mailer.submission_batch.daily.single_file_explainer", date: @date) %>
+
+<% end %>
+
+<% @filenames.each do |filename| %>
+
+- <%= filename %>
```
stephencdaly, line 17: Each list item is being rendered inside `<p>` tags which is different to how it was before. I think removing the whitespace around the item fixes this?

app/views/aws_ses_submission_batch_mailer/_weekly_submission_batch_content.md.erb
```diff
@@ -0,0 +1,36 @@
...
+<%= I18n.t("mailer.submission_batch.weekly.multiple_files_explainer", begin_date: @begin_date, end_date: @end_date) %>
+
+<% else %>
+<%= I18n.t("mailer.submission_batch.weekly.single_file_explainer", begin_date: @begin_date, end_date: @end_date) %>
+
+<% end %>
+
+<% @filenames.each do |filename| %>
+
+- <%= filename %>
```
stephencdaly, line 17: Same with this one regarding list items being in `<p>` tags

stephencdaly (commented):
Nice work, it all fits together really nicely and should be a lot easier to maintain. It's mainly working as expected, but just spotted a few small things.

app/views/aws_ses_submission_batch_mailer/_daily_submission_batch_content.md.erb
DavidBiddle: Yep, that worked - thanks!

app/helpers/email_format_helper.rb
DavidBiddle: Should be resolved by this change in the govuk-forms-markdown gem: https://github.com/govuk-forms/govuk-forms-markdown/pull/100

app/lib/ses_email_formatter.rb
DavidBiddle: Yeah, good spot - have updated to make this work!

stephencdaly (approved):
All looks good now. Nice work!

PR #2218, opened 2026-07-16
Support delivering submissions by multiple methods

stephencdaly:

Trello card: https://trello.com/c/WMnjLlJV

Use the `delivery_configurations` on the form to determine which delivery methods are enabled, and deliver submissions using each method.

I've tested on dev that this is still working as expected for email and S3 submissions.

I've tested locally that submissions are delivered by both S3 and email when both are enabled (I haven't added a rake task to enable this on dev yet)


thomasiles (approved):
I'm not super familiar with this part of the code so i might not be able to spot things missing.

This is comprehensive, well tested and covers the transition between the old and new 🏅

PR #2223, opened 2026-07-20
Add custom branding

SamJamCul:

Trello card: 


config/locales/cy.yml
theseanything, line 1: Obviously these need to be translated into welsh - but can happen in a separate PR. Worth making a TODO

config/branding.yml
```diff
@@ -0,0 +1,15 @@
...
+  logo: /brand_assets/cheshire-east/logo.png
+  favicon: /brand_assets/cheshire-east/favicon.ico
+  opengraph: /brand_assets/cheshire-east/opengraph.jpg
+south-gloucestershire:
+  background_colour: "white"
+  border_colour: "#396631"
+  organisation_name: "South Gloucestershire Council"
+  organisation_url: "https://www.southglos.gov.uk"
+  logo: /brand_assets/south-gloucestershire/logo.png
+  favicon: /brand_assets/south-gloucestershire/favicon.png
```
theseanything, line 15: Nit: No opengraph for south Gloucestershire. Something to fix later.

config/branding.yml
```diff
@@ -0,0 +1,15 @@
+cheshire-east:
+  background_colour: "white"
```
theseanything: Can we use hex values for all colours (as they will be easier to validate, when injecting dynamic css values)?

theseanything (dismissed):
I think this is fine to go in now (considering it won't affect our user's at the moment) - it will allow us to iterate on it.

config/branding.yml
DavidBiddle: I expect we'll find that orgs all have different combinations of icons and sharing images so when we iterate we can make it more resilient to missing icons / different ones existing, but agree we can look at this later

config/branding.yml
DavidBiddle: Done 👍

config/locales/cy.yml
DavidBiddle: I've made a placeholder card for this: https://trello.com/c/npWVLghO/3718-get-custom-branding-content-translated-into-welsh

DavidBiddle: I haven't added a TODO because the i18n normalizer will remove any comments from this file 😁

DavidBiddle (approved):

theseanything (approved):

PR #2238, opened 2026-07-27
Raise an error if a live form submission has no delivery configs

stephencdaly:

Trello card: https://trello.com/c/WMnjLlJV/245-allow-both-email-and-s3-submissions-simultaneously

Ensure that a user cannot complete a live form submission if it is not configured to be delivered to any destination. This shouldn't happen under normal circumstances, but we should protect against an unexpected state.

### DIFF
--- a/app/services/form_submission_service.rb
+++ b/app/services/form_submission_service.rb
@@ -77,9 +77,14 @@ private
   end
 
   def enqueue_deliveries(submission)
-    form.delivery_configurations
-        .filter { |c| c.delivery_schedule == "immediate" }
-        .each { |c| enqueue_delivery(c, submission) }
+    delivery_configurations = form.delivery_configurations.filter { |c| c.delivery_schedule == "immediate" }
+
+    if delivery_configurations.blank? && mode.live?
+      submission.destroy!
+      raise StandardError, "Form id(#{form.id}) has no immediate delivery configurations" if delivery_configurations.blank? && mode.live?
+    end
+
+    delivery_configurations.each { |c| enqueue_delivery(c, submission) }
   end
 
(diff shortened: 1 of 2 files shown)


app/services/form_submission_service.rb
```diff
@@ -77,9 +77,14 @@ def validate_submission
   end
 
   def enqueue_deliveries(submission)
-    form.delivery_configurations
-        .filter { |c| c.delivery_schedule == "immediate" }
-        .each { |c| enqueue_delivery(c, submission) }
+    delivery_configurations = form.delivery_configurations.filter { |c| c.delivery_schedule == "immediate" }
+
+    if delivery_configurations.blank? && mode.live?
+      submission.destroy!
```
theseanything, line 83: Is there a more user friendly way to determine whether a form is deliverable - before the user fills it out? Seems a bad user experience to let someone fill out an entire form, then error out just as they submit and delete all their answers.

stephencdaly: The form creator won't be able to make it live, and the rake tasks prevent against us disabling the delivery methods if there will be no delivery methods enabled. This is just an extra guard to make sure we don't inadvertently mess up some data that will cause a submission not to be sent and then never know about it.

stephencdaly: We could add a check whenever we load the form that it's in a valid state if you think that's worthwhile? Not sure whether that's overkill.

theseanything: >The form creator won't be able to make it live, and the rake tasks prevent against us disabling the delivery methods if there will be no delivery methods enabled. 

Ah okay that seems like it should make this v unlikely

theseanything (approved):

PR #2271, opened 2026-08-10
Add condition model

lfdebrux:

Add a `Condition` model for holding the definition of a condition in a form document.

This allows us to refactor the code to be a bit more Rails-y.

The refactors in this PR are deliberately light-touch, just enough to untangle existing implementation details from the `Step` model and make subsequent PRs related to exit pages easier. 


SamJamCul (approved):
Brilliant stuff, it's much more coherent using a Condition factory in the tests! I'm a big fan 🪭

PR #2272, opened 2026-08-10
Add exit page model

lfdebrux:

This PR adds an `ExitPage` model that works with new-style exit pages in a form document.

Currently we don't use the new exit page model for anything, so this shouldn't affect production, it's just laying preparatory ground work.

### DIFF
--- a/spec/models/exit_page_spec.rb
+++ b/spec/models/exit_page_spec.rb
@@ -1,0 +1,56 @@
+require "rails_helper"
+
+RSpec.describe ExitPage do
+  describe "#initialize" do
+    it "sets the attributes" do
+      exit_page = described_class.new(
+        id: 1,
+        heading: "You are not elegible for this service",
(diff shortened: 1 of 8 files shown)


spec/models/exit_page_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        form_document_exit_page = build(:v2_exit_page)
+        exit_page = described_class.from_form_document(form_document_exit_page)
+        other_exit_page = described_class.from_form_document(form_document_exit_page)
+
+        expect(exit_page == other_exit_page).to be true
+      end
+    end
+
+    context "when other exit page was created from a different form document exit page" do
+      it "returns true" do
```
lfdebrux: D'oh! Good catch thank you :D

lfdebrux: Should be fixed now

SamJamCul (approved):

PR #2273, opened 2026-08-10
remove aws-ses prefix from runner classes.

alice-carr:

We have recently switched to using ses for confirmation emails from notify. Since we are now only using one mail delivery method, we do no need to specify that they are using ses.

Trello card: https://trello.com/c/GfBTJbNL/3154-use-ses-to-send-all-confirmation-emails


app/mailers/delivery_method.rb
```diff
@@ -1,4 +1,4 @@
-class AwsSesDeliveryMethod
+class DeliveryMethod
```
theseanything: I think we probably want to keep this name AwsSesDeliveryMethod - as it's about specifically the AWS SES delivery method.

config/initializers/mailers.rb
```diff
@@ -1,7 +1,7 @@
-require_relative "../../app/mailers/aws_ses_delivery_method"
+require_relative "../../app/mailers/delivery_method"
 
 ActionMailer::Base.add_delivery_method :govuk_notify, GovukNotifyRails::Delivery, api_key: Settings.govuk_notify.api_key
-ActionMailer::Base.add_delivery_method :aws_ses, AwsSesDeliveryMethod
+ActionMailer::Base.add_delivery_method :aws_ses, DeliveryMethod
```
theseanything: Again I think we want to specify this delivery method is about AWS SES here.

theseanything (changes requested):
We probably want to keep the name AwsSesDeliveryMethod

app/mailers/delivery_method.rb
alice-carr: makes sense to me - I've updated this

theseanything (approved):

PR #2283, opened 2026-08-11
Update exit page controller to use new-style exit pages

lfdebrux:

This PR updates the `exit_page#show` action to work with new-style exit pages. The code still assumes that an question with an exit page has only one exit page and only one routing condition, but this PR does most of the changes needed to make the exit page controller work with multiple exit pages. We just need to make some more changes to the step model and the step controller before we can fully support more than one exit page per question in forms-runner.

### DIFF
--- a/app/controllers/forms/exit_pages_controller.rb
+++ b/app/controllers/forms/exit_pages_controller.rb
@@ -5,6 +5,7 @@ module Forms
 
       @back_link = form_step_path(@form.id, @form.form_slug, @step.id)
       @condition = @step.routing_conditions.first
+      @exit_page = @step.exit_pages.find { it.id == @condition.exit_page_id }
     end
   end
 end
--- a/app/controllers/forms/exit_pages_controller.rb
+++ b/app/controllers/forms/exit_pages_controller.rb
@@ -1,11 +1,14 @@
 module Forms
   class ExitPagesController < StepController
     def show
-      return redirect_to form_step_path(@form.id, @form.form_slug, current_context.next_step_slug) unless current_context.can_visit?(@step.id)
+      unless current_context.can_visit?(@step.id) && @step.exit_page_condition_matches?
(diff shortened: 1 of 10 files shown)


app/models/exit_page.rb
theseanything: You're these constructors, initialise `id ` as different types. It's an Integer here, and a String in self.from_form_document. Can we have consistency here? Helps stop bugs in other code that assumes types like equality checks `==` etc..

app/controllers/forms/exit_pages_controller.rb
```diff
@@ -1,10 +1,14 @@
...
     def show
-      return redirect_to form_step_path(@form.id, @form.form_slug, current_context.next_step_slug) unless current_context.can_visit?(@step.id)
+      unless current_context.can_visit?(@step.id) && @step.exit_page_condition_matches?
+        return redirect_to form_step_path(@form.id, @form.form_slug, current_context.next_step_slug)
+      end
 
       @back_link = form_step_path(@form.id, @form.form_slug, @step.id)
       @condition = @step.routing_conditions.first
+      @exit_page = @step.exit_pages.find { it.id == @condition.exit_page_id }
+      raise KeyError, "Couldn't find ExitPage with id=#{@condition.exit_page_id}" if @exit_page.nil?
```
theseanything, line 11: Is this okay to be returning a 500 error here?

theseanything (commented):
To minor comments - but generally looks fine!

app/controllers/forms/exit_pages_controller.rb
lfdebrux: Yes, I think so; I can't think what other HTTP status would be appropriate to return here. I don't think it's a 404 because the page should exist.

app/models/exit_page.rb
lfdebrux: Done

theseanything (approved):

PR #2285, opened 2026-08-12
Add custom branding to confirmation emails

theseanything:
Forms with a configured brand now send confirmation emails with the organisation's logo, colours and website link in the banner instead of GOV.UK branding, matching the branded web pages. This applies to both variants, including emails with a copy of the user's answers.

Only the confirmation emails, not submission emails to form processors. They keep the GOV.UK banner, and forms with no brand or an unrecognised brand ID fall back to it too.

Email clients need absolute image URLs, so there's a new `forms_runner.base_url` setting for this app's public URL. 

Added here:
https://github.com/govuk-forms/forms-deploy/pull/2239

image


stephencdaly (approved):
The changes look sensible. Have you tested this in all the different Outlook versions? https://github.com/govuk-forms/forms-team/wiki/Testing-submission-email-changes

PR #2301, opened 2026-08-18
Add exit page id to exit page path

lfdebrux:

Add an optional exit page id to the exit page path, and use it for new-style exit pages.

This is mainly useful for analytics/debugging purposes, it lets us see just from the request path which exit pages are visited most frequently for a form.

There are various edge cases we try to handle: as for other steps, if the path doesn't match the page the form filler should be shown based on their answers, rather than showing an error page they are redirected to the correct path. Also, to try and avoid redirect loops or 404s when deploying this change, there is a temporary fallback for paths without the exit page id which can be reverted in a future PR.

For old-style exit pages we continue to use the path without the exit page id.

### DIFF
--- a/spec/requests/forms/exit_pages_controller_spec.rb
+++ b/spec/requests/forms/exit_pages_controller_spec.rb
@@ -97,10 +97,12 @@ RSpec.describe Forms::ExitPagesController, type: :request do
       end
     end
 
-    context "when the request path does not include an exit page id" do
-      it "redirects to the correct exit page path" do
+    context "when the request path does not include an exit page id", :capture_logging do
+      it "falls back to the old behaviour and renders the exit page for the matching condition" do
         get exit_page_path(mode: "form", form_id: form.form_id, form_slug: form.form_slug, step_slug: step_with_exit_page.id)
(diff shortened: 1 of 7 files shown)


spec/requests/forms/exit_pages_controller_spec.rb
```diff
@@ -47,28 +47,72 @@
...
+
+      context "when the request path does not include an exit page id" do
+        it "redirects to the next unanswered question" do
+          get exit_page_path(mode: "form", form_id: form.form_id, form_slug: form.form_slug, step_slug: step_without_exit_page.id)
+          expect(response).to redirect_to form_step_path(mode: "form", form_id: form.form_id, form_slug: form.form_slug, step_slug: next_step_in_form.id)
+        end
+      end
+    end
+
+    context "when the request path does not include an exit page id" do
```
SamJamCul:
```suggestion
    context "when the request path does include an exit page id" do
```

SamJamCul (approved):
Looks good, LGTM

spec/requests/forms/exit_pages_controller_spec.rb
lfdebrux: Thanks for reviewing, I'm not sure why this change is needed though?

PR #2302, opened 2026-08-19
Handle steps with more than exit page

lfdebrux:

Update forms-runner to work with steps with more than one exit page.

This is mostly tests, with some changes to the step model, and some small changes to the step controller and exit page controller.

The code to get the exit page id in the step controller is very inefficient, but I've already spent too long refactoring forms-runner to make this change work.


thomasiles (approved):
This works well for me locally and I haven't been able to trip it up.

nice one 🏅

PR #2304, opened 2026-08-20
Fetch form branding from the brands API

theseanything:
Forms with a brand_id now get their branding from the forms-admin brands API. 

This PR keeps the `config/branding.yml` as a priority source of config. But if a brand isn't present in the config yaml, it will try the API.

We cache brands for 5 minutes, so we don't have to request the brand for every page load.

Removing the yml once brands and their assets are fully setup from forms-admin.


stephencdaly (approved):
All looks sensible to me

PR #2314, opened 2026-08-26
Update Cheshire East logo to higher res

theseanything:


DavidBiddle (approved):
I can't check how it looks in the email on my local machine but I've tested the runner view and it looks good!

PR #2325, opened 2026-09-04
Remove delivery configurations fallback code

DavidBiddle:

Trello card: Related to https://trello.com/c/P7bDQZTP/3248-get-submission-delivery-configuration-from-latest-live-version-of-the-form-in-forms-runner

When we implemented delivery configurations, we added fallback methods to handle any submissions that predated this work. We've now waited long enough that any submissions without delivery configurations have been processed and deleted, so we can remove this fallback code.

### DIFF
--- a/app/lib/batch_submissions_selector.rb
+++ b/app/lib/batch_submissions_selector.rb
@@ -1,15 +1,11 @@
 class BatchSubmissionsSelector
   Batch = Data.define(:form_id, :mode, :submissions)
 
-  # TODO: we can remove checking the send_daily_submission_batch and send_weekly_submission_batch in September 2026
-  # when all Submissions without delivery_configurations on the form document have been deleted.
   DAILY_BATCH_FORM_DOCUMENT_CONDITION = <<~SQL.squish
-    (form_document->>'send_daily_submission_batch')::boolean = true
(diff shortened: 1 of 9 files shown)


app/lib/batch_submissions_selector.rb
```diff
@@ -4,12 +4,10 @@ class BatchSubmissionsSelector
   # TODO: we can remove checking the send_daily_submission_batch and send_weekly_submission_batch in September 2026
```
stephencdaly: We can remove this comment now

app/lib/batch_submissions_selector.rb
```diff
@@ -59,14 +57,8 @@ def form_ids_and_modes_with_send_weekly_submission_batch(begin_at)
                 .pluck(:form_id, :mode)
     end
 
-    def batch_enabled_for_daily_submission?(form_document)
-      form_document["send_daily_submission_batch"] == true ||
-        daily_delivery_configuration?(form_document)
-    end
-
     def batch_enabled_for_weekly_submission?(form_document)
```
stephencdaly: Looks like this method is unused now

stephencdaly (commented):
Looks good, there's some extra small bits that can be removed.

stephencdaly (approved):

PR #2331, opened 2026-09-08
Name queue worker job traces with job class

whi-tw:

Trello card: https://trello.com/c/KnaPaYVl/3269-tracing-for-cron-jobs-queue-worker

This means that in the trace map, it's easier to understand
where the traces are being emitted from. As there's no user interaction
that triggers these jobs, there's no 'interaction wrapper' around
the traces.

This is related to https://github.com/govuk-forms/forms-deploy/pull/2266


cadmiumcat (approved):
Good call!

PR #2333, opened 2026-09-09
Support fetching form documents from v3 API

DavidBiddle:

Trello card: https://trello.com/c/P7bDQZTP/3248-get-submission-delivery-configuration-from-latest-live-version-of-the-form-in-forms-runner

Adds new resources for interacting with the v3 form document API.

### DIFF
--- a/app/resources/api/v3/form_document_resource.rb
+++ b/app/resources/api/v3/form_document_resource.rb
@@ -9,11 +9,20 @@ class Api::V3::FormDocumentResource < ActiveResource::Base
 
   class << self
     def find_by_tag(form_id, tag, **options)
-      get("#{form_id}/versions/#{tag}", **options)
+      return draft_form_document(form_id, **options) if tag == :draft
+
+      latest_version = get("#{form_id}/versions/#{tag}", **options)["version"]
+      find_by_version(form_id, latest_version, **options)
     end
 
     def find_by_version(form_id, version, **options)
       get("#{form_id}/versions/#{version}", **options)
+    end
+
+  private
+
+    def draft_form_document(form_id, **options)
+      get("#{form_id}/versions/draft", **options)
     end
   end
 end
--- a/app/resources/api/v3/form_document_resource.rb
+++ b/app/resources/api/v3/form_document_resource.rb
@@ -1,0 +1,15 @@
+class Api::V3::FormDocumentResource < ActiveResource::Base
+  self.element_name = "form"
+  self.site = Settings.forms_api.base_url
+  self.prefix = "/api/v3/"
+  self.include_format_in_path = false
+
+  has_many :steps, class_name: "Api::V2::StepResource"
+  has_many :delivery_configurations, class_name: "Api::V2::DeliveryConfigurationResource"
+
+  class << self
+    def find_by_tag(form_id, tag, **options)
+      get("#{form_id}/versions/#{tag}", **options)
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


app/resources/api/v3/form_document_resource.rb
```diff
@@ -0,0 +1,23 @@
...
+  self.site = Settings.forms_api.base_url
+  self.prefix = "/api/v3/"
+  self.include_format_in_path = false
+
+  has_many :steps, class_name: "Api::V3::StepResource"
+  has_many :delivery_configurations, class_name: "Api::V3::DeliveryConfigurationResource"
+
+  class << self
+    def find(form_id, tag, params: {})
+      super(:one, from: "#{prefix}#{collection_name}/#{form_id}/#{tag}", params:)
```
stephencdaly: The v3 endpoint is`api/v3/forms/:form_id/versions/live` etc. This is missing the `/versions`

app/resources/api/v3/form_document_resource.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+
+    def find_by_tag(form_id, tag, **options)
+      get("#{form_id}/#{tag}", **options)
```
stephencdaly: This doesn't work the same as the v2 API for `versions/live` and `versions/archived`. Rather than returning the form document data directly, it returns a redirect to the endpoint for the live or archived version (or 404 if the form is not in the correct state).

`/versions/:version` and `versions/draft` are the only endpoints that return the form document data.

We'll need to handle the redirect response and make a second request to get the actual form data.

spec/factories/v3_step.rb
```diff
@@ -0,0 +1,156 @@
+FactoryBot.define do
+  factory :v3_step, class: Api::V3::StepResource do
```
stephencdaly: I wonder whether we really need `v2_form_document` vs `v3_form_document` etc, as the schema hasn't changed between the API versions. I wonder whether we can just remove the version prefix for the factories and worry about this when we need to make a breaking change to the response schema in the future?

stephencdaly (changes requested):

spec/factories/v3_step.rb
DavidBiddle: I've removed the unnecessary extra resource classes. I think I'll do a separate PR after this one's merged to rename the factories

app/resources/api/v3/form_document_resource.rb
```diff
@@ -0,0 +1,28 @@
+class Api::V3::FormDocumentResource < ActiveResource::Base
+  self.element_name = "form"
+  self.site = Settings.forms_api.base_url
+  self.prefix = "/api/v3/"
+  self.include_format_in_path = false
+
+  has_many :steps, class_name: "Api::V2::StepResource"
+  has_many :delivery_configurations, class_name: "Api::V2::DeliveryConfigurationResource"
```
stephencdaly, line 8: I wonder whether we want to remove the `Api::V2::` namespace from these models if we're re-using them for the v3 API? We could always do that or remove them later though.

stephencdaly (approved):
All the changes seem to work as expected

app/resources/api/v3/form_document_resource.rb
DavidBiddle: I reckon it makes sense to do that in the same PR as renaming the factories - good suggestion 👍

